PR #9, opened 2014-05-01
Lint and documentation fixes

rtyler:


Rakefile
```diff
@@ -4,6 +4,7 @@ require 'puppetlabs_spec_helper/rake_tasks'
 
 PuppetLint.configuration.send('disable_80chars')
```
Spredzy, line 5: Would adding `PuppetLint.configuration.send('disable_documentation')` make sense ? I understand main class needs it. But for smaller, 'clear' define, maybe not ...

rtyler, line 5: Because of the nature of this repository and project, I am hoping to prevent undocumented code from getting merged :)

Spredzy, line 5: +1

PR #10, opened 2014-05-01
Add support for managing sudoers access across machines

rtyler:
This includes the saz/sudo module and defines a difference in sudoer needs
between OSUOSL managed machines and those outside of the OSUOSL datacenter,
which won't have an `osuadmin` account on them

[FIXED INFRA-17]

### DIFF
--- a/dist/profile/manifests/sudo.pp
+++ b/dist/profile/manifests/sudo.pp
@@ -1,0 +1,27 @@
+#
+# Main sudo management profile
+class profile::sudo {
+  include ::sudo
+
+  sudo::conf { 'env-defaults':
+    content => 'Defaults        env_reset',
+  }
+
+  sudo::conf { 'secure-path':
+    content => 'Defaults        secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"',
+  }
+
+  sudo::conf { 'root':
+    content  => 'root ALL=(ALL) ALL',
+  }
+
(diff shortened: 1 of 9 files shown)


dist/profile/manifests/sudo.pp
```diff
@@ -0,0 +1,27 @@
...
+  sudo::conf { 'admins':
+    priority => '10',
+    content  => '%admin ALL=(ALL) ALL',
+  }
+
+  sudo::conf { 'sudo':
+    priority => '10',
+    content  => '%sudo ALL=(ALL) ALL',
+  }
+}
```
Spredzy, line 27: Wouldn't it be better - more flexible -  to pass a hash, and the use create_resources().

```
create_resources('sudo::conf', $hash, $default)
```

rtyler, line 27: @Spredzy Yes, that might be better, for now I'm not sure I want to put "sudo config" into hiera.

I discussed this a bit IRL with @adamcrews, it's kind of a "6 in one hand, half dozen in the other" situation. I'd prefer to keep this here until we feel more confident in the separation of data and configuration in hiera.

Spredzy, line 27: You don't have to have those data in hiera _yet_.

```
  $data  = {
    'root' => {},
    'admins => {},
  }

create_resources('sudo::conf', $data)
```

Would work, and it will already be ready the day we decide to move to hiera.  But I have nothing against current situation.

PR #18, opened 2014-05-01
Add the NTP profile to jenkins-infra

Spredzy:
Add the NTP profile to the puppetmaster-role.
Configuration is taken from hiera.

### DIFF
--- a/dist/profile/manifests/base.pp
+++ b/dist/profile/manifests/base.pp
@@ -1,0 +1,7 @@
+#
+# Basic profile included in each node
+class profile::base {
+  include profile::accounts
+  include profile::ntp
+  include profile::sudo
+}
--- a/dist/role/manifests/edamame.pp
+++ b/dist/role/manifests/edamame.pp
@@ -1,4 +1,6 @@
 #
 # Edamame is a VM with 2x CPUs and 2GB of RAM at the OSUOSL
 class role::edamame {
+  include profile::base
+  include profile::sudo::osu
 }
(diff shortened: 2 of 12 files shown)


dist/role/manifests/puppetmaster.pp
```diff
@@ -1,8 +1,7 @@
 #
 # role::puppetmaster defines what a node role that should look like
 class role::puppetmaster {
-  include profile::accounts
+  include profile::base
   include profile::puppetmaster
   include profile::r10k
-  include profile::sudo::osu
```
rtyler, line 7: The `puppetmaster` role lives in the OSUOSL so it should still include `profile::sudo::osu`

dist/role/manifests/edamame.pp
```diff
@@ -1,4 +1,5 @@
 #
 # Edamame is a VM with 2x CPUs and 2GB of RAM at the OSUOSL
 class role::edamame {
+  include profile::base
```
rtyler, line 4: `edamame` is in the OSU so can you add `include profile::sudo::osu` as well as base?

dist/profile/manifests/base.pp
```diff
@@ -0,0 +1,7 @@
+#
+# Basic profile included in each node
+class profile::base {
+  include profile::accounts
+  include profile::ntp
+  include profile::sudo::osu
```
rtyler: Not all our machines are in the OSU, so this should incorporate `include profile::sudo`

rtyler:
Superseded by #22

PR #21, opened 2014-05-02
Adding apache module

kohsuke:
... and bringing the virtual host configuration back to maintain uniformity with the way we configure virtual hosts.

### DIFF
--- a/dist/jenkins_apache/manifests/log_rotation.pp
+++ b/dist/jenkins_apache/manifests/log_rotation.pp
@@ -12,6 +12,6 @@ class jenkins_apache::log_rotation {
     command => 'cd /var/log/apache2; ./compress-log.rb',
     user    => root,
     minute  => 7,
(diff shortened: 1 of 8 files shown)


dist/jenkins_apache/manifests/log_rotation.pp
```diff
@@ -0,0 +1,17 @@
...
+    require => Package['apache2'],
+  }
+
+  cron {
+  'compress logs':
+    ensure  => present,
+    command => 'cd /var/log/apache2; ./compress-log.rb',
+    user    => root,
+    minute  => 7,
+    require => Package['apache2'],
```
jeffweiss: Probably not a huge deal, but this cron job should likely depend on `File['/var/log/apache2/compress-log.rb']` instead.

rtyler:
This is now a redundant pull request, closing.

PR #27, opened 2014-05-10
Adding place holder manifest for celery

kohsuke:
The plan is to use this as a build slave for J-on-J

### DIFF
--- a/spec/server/celery/celery_spec.rb
+++ b/spec/server/celery/celery_spec.rb
@@ -1,0 +1,5 @@
+require_relative './../spec_helper'
+
+describe 'celery' do
+  it_behaves_like "a Rackspace hosted build machine"
+end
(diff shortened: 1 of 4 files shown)


spec/server/celery/celery_spec.rb
```diff
@@ -0,0 +1,5 @@
+require_relative './../spec_helper'
+
+describe 'celery' do
+  it_behaves_like "a Rackspace hosted build machine"
```
rtyler, line 4: This will actually fail because that shared context is not defined. The string of `it_behaves_like` is defined in [spec/server/support/basic_shared_contexts.rb](https://github.com/jenkins-infra/jenkins-infra/blob/staging/spec/server/support/basic_shared_contexts.rb)

Is there anything special about a Rackspace hosted machine that's distinctive from any other normal machine? Going off the `role::celery` class above, I'm guessing not.

PR #31, opened 2014-07-07
Define misc. apache settings

kohsuke:
There is a small amount of Apache setup that needs to happen whenever we run Apache. This change adds those.

### DIFF
--- a/dist/profile/manifests/apache_misc.pp
+++ b/dist/profile/manifests/apache_misc.pp
@@ -1,0 +1,22 @@
+#
+# Misc. apache settings
+#
+#
+#
+class profile::apache_misc {
+  include apache
+  # log rotation setting lives in another module
+  include apache-logcompressor
+
+  file { '/etc/apache2/conf.d/00-reverseproxy_combined':
+    ensure => present,
+    source => "puppet:///modules/${module_name}/apache/00-reverseproxy_combined.conf",
+    mode   => '0444',
+  }
+
+  file { '/etc/apache2/conf.d/other-vhosts-access-log':
+    ensure => present,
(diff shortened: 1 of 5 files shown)


dist/profile/manifests/apache_misc.pp
```diff
@@ -0,0 +1,22 @@
+#
+# Misc. apache settings
+#
+#
+#
+class profile::apache_misc {
```
rtyler, line 6: Any reason for using the underscore instead of a dash for the class name? Seems weird with "apache-logcompressor" right below

rtyler:
What a weird test failure, can you retrigger it?

```
1) profile::apache_misc 
     Failure/Error: it { should contain_file '/etc/apache2/conf.d/other-vhosts-access-log' }
     SignalException:
       SIGTERM
     # (eval):3:in `_racc_yyparse_c'
     # (eval):3:in `yyparse'
     # /usr/share/ruby/benchmark.rb:295:in `realtime'
     # ./spec/classes/profile/apache_misc_spec.rb:4:in `block (2 levels) in <top (required)>'

```

dist/profile/manifests/apache_misc.pp
kohsuke, line 6: I've originally used hyphen and puppet lint complained that I shouldn't do that. I like hyphen better myself.

kohsuke:
Closing & reopening to retrigger a pull request build.

dist/profile/manifests/apache_misc.pp
rtyler, line 6: Why not change it back, and I'll add a change to tell puppet-lint to STFU about hyphens. I like them better too

rtyler:
Going to merge and then rename the module to use dashes. puppet-lint be damned

PR #36, opened 2014-09-20
Manage and run jenkins-admin in Docker

rtyler:


rtyler:
I need @kohsuke to kill the existing `jenkins-admin` bot on Freenode so I can make sure that this works before deploying to production

PR #73, opened 2015-03-24
Puppet node name changes

kohsuke:
`confluence` is a service name. It should be using the real host name `edamame` instead.

For consistency, using this opportunity to set short vegetable names for everything.


kohsuke:
After confirming a successful validation, I'm self-merging this change.

PR #74, opened 2015-03-28
[INFRA-41] JIRA deployment step 1

kohsuke:
This is the first of the series of changes to deploy JIRA in more manageable fashion.

In this change, I'm deploying a fake webapp container instead of actual JIRA container. This allows us to test the way we manage Apache on edamame. The change has been successfully tested through serverspec.

In developing this PR, I had to fix a few problems in the way vagrant/serverspec runs. So those are a part of this change.

Once this change gets deployed to production, I can work toward replacing the fake webapp by the actual JIRA container, which involves some manual work and service down time.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,6 +11,9 @@ Vagrant.configure("2") do |config|
   config.vm.box = 'dummy'
   config.vm.box_url = 'https://github.com/mitchellh/vagrant-aws/raw/master/dummy.box'
 
+  # modules/account/.travis.yml has incorrect link target, and this blows up
+  # when vagrant tries to rsync files as it tries to resolves symlinks.
+  # see http://www.trilithium.com/johan/2011/09/delete-broken-symlinks/
   `find -L . -type l -delete`
 
   config.vm.provider(:aws) do |aws, override|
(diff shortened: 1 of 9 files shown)


Vagrantfile
```diff
@@ -11,6 +11,8 @@ Vagrant.configure("2") do |config|
   config.vm.box = 'dummy'
   config.vm.box_url = 'https://github.com/mitchellh/vagrant-aws/raw/master/dummy.box'
 
+  `find -L . -type l -delete`
```
rtyler, line 17: This should _definitely_ have some commentary around it, why are you deleting all my sym links!

dist/profile/manifests/jira.pp
```diff
@@ -0,0 +1,68 @@
+# Run containerized JIRA to serve issues.jenkins-ci.org
```
rtyler, line 1: Will you update the comment to include a link to the docker container repo, which I'm hoping has a good readme on how the container is all built out and whatnot?

kohsuke:
@rtyler addressed your reviews.

rtyler:
LGTM, I'm not thrilled with the sytematic deletion of symbolic links in the tree whenever you execute a vagrant command, but Idon't want to block you either

PR #75, opened 2015-03-28
[INFRA-41] manage SSL private key & certificate

kohsuke:
A triplet of (1) private key, (2) certificate, (3) certificate chain to root CA is grouped by naming convention and identified by `profile::ssl-cert::id` value.

Certificates are placed as files, but private key is defined in hiera as encrypted value.

The `wiki-jira` bundle consists of a real certificate used by eggplant, plus a bogus private key. If the soundness of this approach is validated, then I'll put the actual private key.

The `bogus` bundle is a generated self-signed certificate and private key. It is unprotected to allow vagrant+serverspec to use them.

The `profile::ssl-cert::id` value, which is defined in hiera, selects which key/certificate bundle the host is going to get. edamame has the `wiki-jira` bundle specified (in preparation of JIRA deployment), and Vagrant gets the `bogus` bundle.


kohsuke:
This one really needs @rtyler review

dist/profile/manifests/ssl-cert.pp
```diff
@@ -0,0 +1,29 @@
+#
+# SSL certificates
+#
+class profile::ssl-cert (
+  # all injected from hiera
+  $id,        # identify which private key / certificate pair we should use. This usually comes from hieradata/clients/*.yaml
```
rtyler: Are there restrictions on what "id" can be?

dist/profile/manifests/ssl-cert.pp
```diff
@@ -0,0 +1,29 @@
+#
+# SSL certificates
+#
+class profile::ssl-cert (
```
rtyler: Since this is a pretty Apache-specific profile, I'd rather it be named "apache-cert" or something like that to make sure the apache coupling is very very clear

rtyler:
Looks pretty good to me, I'm glad this is finally being addressed!

dist/profile/manifests/ssl-cert.pp
kohsuke: I'm not sure if I understand the question. They reference specific files and entry in hiera.
In dbd4d30 I've expanded this comment to hopefully clarify this.

kohsuke:
[test error](https://jenkins.ci.cloudbees.com/job/infra/job/jenkins-infra/295/console) appears to be unrelated, so I'm having Jenkins try this change again one more time.

rtyler:
:+1:

PR #76, opened 2015-03-30
Deploy datadog node to every node

kohsuke:
Thanks to [the support from Datadog the company](https://twitter.com/jhotta/status/581980645733244928), we can use Datadog in our infrastructure. So this change deploys datadog everywhere by adding it to `profile::base`.


rtyler:
YAY

PR #77, opened 2015-03-31
Tracking all the certificates here

kohsuke:
Following up on PR #75, this change replaces the encrypted bogus private key with the encrypted actual private key.

It also adds the other certificate we use on cucumber.


kohsuke:
Test is passing, so I'm merging this myself.

PR #78, opened 2015-03-31
Containerized Confluence deployment

kohsuke:
This is the Confluence equivalent for PR #74.

It deploys two containers (the static cache layer container and the Confluence container), and set up everything, except it uses the mock webapp instead of the actual Confluence container.

Once this change gets deployed to production, I can work toward replacing the fake webapp by the actual Confluence container, which involves some manual work and service down time.


kohsuke:
Test is passing, so I'm merging this myself.

PR #81, opened 2015-04-02
Changes to deploy mock-Confluenc on lettuce correctly

kohsuke:
I think I'm almost there.


kohsuke:
Reopening to rerun tests

PR #87, opened 2015-04-05
Adding another volume to okra

kohsuke:
archives area has grown past 150GB, and it's blocking UC.

150GB volume was added via Rackspace cloud console, and this step makes the OS use the new volume.


rtyler:
:+1:

PR #89, opened 2015-04-06
[INFRA-261] more Confluence related changes

kohsuke:
Using the mock wikidb to validate the new image better. Some more follow-up changes are needed.

### DIFF
--- a/dist/profile/manifests/confluence.pp
+++ b/dist/profile/manifests/confluence.pp
@@ -34,7 +34,7 @@ class profile::confluence (
     content => join([
         "LDAP_PASSWORD=${ldap_password}",
         "DATABASE_URL=${database_url}"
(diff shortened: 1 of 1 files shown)


dist/profile/manifests/confluence.pp
```diff
@@ -22,7 +22,9 @@
   }
   file { '/srv/wiki/home':
     ensure  => directory,
-    recurse => true,
+    # confluence container is baked with UID=1000 & GID=1001
+    owner   => 1000,
```
rtyler, line 26: How will you be sure that these aren't used already on any of the sytems that this Puppet might run on? 1000 is a pretty low and most likely is already used.

On my personal Linux machine, 1000 is the UID of the first user

PR #91, opened 2015-04-06
DO NOT MERGE YET [INFRA-261] switch wiki to lettuce

kohsuke:
This is one of the last steps. Switch the name server to drive production traffic to lettuce.

**This change needs to be coordinated with Confluence production site switch over. Do not merge yet.**


orrc:
Awesome. Maybe the existing wiki TTL can be reduce to five minutes or so already in anticipation of the switchover?

Is Confluence configured to send email via the "OSUOSL mail relays", or will lettuce send out email directly (i.e. it would need to be added to the SPF DNS entries)?

PR #96, opened 2015-04-11
[INFRA-261] rewrite confluence cache container & SPF change deployment

rtyler:


kohsuke:
Reverting the production switch over of Wiki while we still test out the confluence cache container rewiring.

PR #126, opened 2015-05-03
recurse tries to manage files in the directory.

kohsuke:
I thought it was for creating parent directories recursively. Ouch.


rtyler:
Will these parent directories still be created on a fresh provision now?

kohsuke:
account resource has `managed_home` set to true by default, so that guy creates `/srv/jira` and everyone above.

PR #128, opened 2015-05-04
Have Jenkins perform DNS zone file syntax check 

kohsuke:
Added rake task to perform syntax check on DNS zone file, and have
Jenkins test it all the time.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -35,3 +35,8 @@ task :resolve do
   Rake::Task['spec_clean'].invoke
   Rake::Task['spec_prep'].invoke
 end
+
+desc 'Check syntax of DNS zone file'
+task "test-zonefile" do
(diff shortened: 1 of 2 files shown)


kohsuke:
Ouch, DEV@cloud build slaves cannot run Docker.

rtyler:
"I made a docker container for this" doesn't make any sense to me. Can we not just install `named-checkzone` at runtime in Jenkins and execute that?

kohsuke:
If this runs on ci.jenkins-ci.org we could have added all the tools, but adding custom tools isn't very easy on jenkins.ci.cloudbees.com unless we bring in our own slave, which might be what we should do.

rtyler:
The problem that I have with this approach is that it's something I'll never do locally, and I can't reasonable make it a dependency on the `spec` task :/

rtyler:
@kohsuke I was thinking a serverspec test might be a good way to get this done too

kohsuke:
test via serverspec is easy to add, but now I tend not to run it locally, and Jenkins certainly doesn't do it!

PR #139, opened 2015-05-12
Punt arnaud into prod!

rtyler:


rtyler:
Running some local verifications right now, will merge once they're squared away

PR #140, opened 2015-05-18
Add atlassian-admins 

rtyler:
JIRA INFRA-307


kohsuke:
:+1: I've looked at the code and see no reasons to block this from getting merged. So I'm going to merge this.

That said, if you are going to let people login to the box and poke around, then I think this will likely need more work. For example, they'd need to be able to get inside containers via (`docker exec`), `jira` and `wiki` users have to have the same `atlassian-admins` as the primary group, and so on.

PR #141, opened 2015-05-18
Add SSH public key for orrc.

orrc:
As requested by @rtyler.


rtyler:
I plucked this into the staging branch directly

PR #145, opened 2015-08-01
Fixing puppet failures

kohsuke:
Accounts are created on every machine, so atlassian-admins group must
also exist everywhere, too.


rtyler:
This test failure looks legit

PR #146, opened 2015-08-02
Disable maintenance

daniel-beck:
Disable maintenance again.


kohsuke:
Reopening to induce another test run

PR #148, opened 2015-08-05
Adding a new server called cabbage

kohsuke:


rtyler:
:+1:

Now we can make sauerkraut!

PR #150, opened 2015-08-05
Make new servers automatically assume the default role,

kohsuke:
for example so that you can login as you.

Otherwise you cannot get in until you explicitly create a new class for the server.


rtyler:
:+1: I think this is reasonable, but I think we should keep stating it explicitly in other roles

PR #156, opened 2015-11-14
Dependency updates

rtyler:
This should all be fairly low impact, the tests are all looking good to me thus far. Mostly submitting this because Beaker will require some updated dependencies and want to merge this separate from that


rtyler:
I reproduced the test failure locally when I ran the CI scripts, back to the drawing board

rtyler:
The thing that's missing now is a Docker-capable build slave on `jenkins.ci.cloudbees.com`. I'm going to dig around for that, I really want the zonefile checking to be merged and done as part of CI

PR #161, opened 2015-12-07
Make both the pe_gem and pe_puppetserver_gem present.

rtyler:
There are still some old modules that need to the `pe_gem` provider to be
installed. All of this will become easier in the next PE server upgrade

References INFRA-502


rtyler:
These tests are bombing out because of something funny on the builder. Imagine my surprise =_=

PR #176, opened 2015-12-16
Deploy htop package everywhere - http://hisham.hm/htop/ How can you live without it ?

aheritier:


rtyler:
Hah, I was actually about to add this as well @aheritier :)

I'll merge this and then refactor it into my `profile::diagnostics` module which I'm still working on

aheritier:
yes I had a doubt to find the right place to add it. Thx

PR #191, opened 2015-12-23
Enable hard fails for SPF records

rtyler:


rtyler:
```
15:15 @Dave:� 1 characters makes a bit of difference
15:16 @Dave:� :+1: :ship:
```

From my pal @redterror

redterror:
I took it for granted that those IPs are the _only_ ones sending for jenkins-ci.org.  Assume that's the case, this is indeed the better policy.

PR #242, opened 2016-02-14
Expand the list of commands atlassian-admins can touch to /usr/bin/docker

rtyler:
This should ensure that atlassian-admins can inspect and interact with the
containers running atlassian services

### DIFF
--- a/dist/profile/manifests/atlassian.pp
+++ b/dist/profile/manifests/atlassian.pp
@@ -19,7 +19,7 @@ class profile::atlassian {
 
   sudo::conf { $group_name:
     priority => 10,
-    content  => "%${group_name} ALL=(ALL) NOPASSWD: /usr/sbin/service",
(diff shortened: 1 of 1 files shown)


orrc:
Syntax looks good.

I don't know if the Docker setup on lettuce also requires users being in the `docker` group, or something similar in order to access the unix socket etc..?

rtyler:
@orrc from my understanding adding a user to the `docker` group gives them unprivileged access to the Docker daemon. That would be one option, but then I would need to add each user that's in `atlassian-admins` to the `docker` group too.

I figured this would be the easiest to manage and least amount of effort, I suppose we'll see though :)

orrc:
Ignore me: I was forgetting that this is giving _sudo_ rights to execute the `docker` command; no need to change any user accounts or groups.

PR #244, opened 2016-02-18
Provision Kep as a new machine in our rackspace account

rtyler:
Cabbage is dead y'all

References INFRA-570


abayer:
:+1: - we'll have to modify the `infra_statistics` job (and related rsync from cucumber) to go there instead.

PR #272, opened 2016-03-08
Create a new profile for managing a Jenkins master

rtyler:
Right now this will obviously only be applied to Cucumber, but hopefully this
allows us to get to the point where we can migrate ci.j.o off that host
entirely

Fixes [INFRA-600](https://issues.jenkins-ci.org/browse/INFRA-600)

### DIFF
--- a/dist/profile/manifests/jenkins.pp
+++ b/dist/profile/manifests/jenkins.pp
@@ -1,0 +1,31 @@
+#
+# Profile for managing a Jenkins master installation
+class profile::jenkins {
+  include firewall
+
+  # This is a legacy role imported from infra-puppet, thus the goofy numbering
+  firewall { '108 Jenkins CLI port' :
+    proto  => 'tcp',
+    port   => 47278,
+    action => 'accept',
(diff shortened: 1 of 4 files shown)


kohsuke:
It copies some of the iptables rules but not all. Isn't this going to end up clobbering rules that's not explicitly defined here? Or is Puppet smart enough to keep them untouched?

rtyler:
@kohsuke the `firewall` module is, unfortunately, not as clever as you would hope that it is. It relies on the `name` of a resource, e.g. `'108 Jenkins CLI port'`. to determine whether it needs to execute an `iptables -A` command.

From production currently:

```
% sudo iptables -L | grep '108'
ACCEPT     tcp  --  anywhere             anywhere            multiport ports 47278 /* 108 Jenkins CLI port */ 
%
```

This means that Puppet will treat that single rule as idempotentn and not clobber it.

Like I said, the `firewall` module isn't actually very clever :-/

kohsuke:
OK, then.

PR #278, opened 2016-03-23
Puppetize the accounts-app and move it over to eggplant

rtyler:
## Work in progress

See [INFRA-11](https://issues.jenkins-ci.org/browse/INFRA-11) and [INFRA-613](https://issues.jenkins-ci.org/browse/INFRA-613)

This gets things started, what's missing right now is:
- [ ] Generate certificate for accounts.jenkins.io
- [x] Create Apache vhost for accounts.jenkins.io to terminate TLS

PR #280, opened 2016-03-23
New CNAMEs woot

rtyler:


rtyler:
This clears the way for #278

PR #295, opened 2016-03-25
Use a better worker module for Apache when serving the static site

rtyler:
I've already made this change in production due to an active site issue (HN hug
of death). This codifies that change for posterity


abayer:
+1

PR #327, opened 2016-04-15
Setup openldap server via puppet

nibalizer:
This uses the camptocamp openldap module. It defines a new host called
kale since cucumber is likely not the new host for holding an ldap
server.

New hiera keys:

ldap_rootpw (string)

New depenedencies:

libaugeas-ruby (for camptocamp/openldap)


dist/role/manifests/kale.pp
```diff
@@ -0,0 +1,6 @@
+# something something something
```
rtyler: epic documentation :smiling_imp:

Puppetfile
```diff
@@ -79,3 +79,8 @@ mod 'datadog/datadog_agent', '1.6.0'
 
 # Used for grabbing certificates for jenkins.io
 mod 'danzilio/letsencrypt', '1.0.0'
+
+# For managing ldap
+mod 'camptocamp/openldap', '1.14.0'
+mod  'herculesteam/augeasproviders_shellvar', '2.x'
```
rtyler: It's unclear to me what these additional augeas providers do here, are they actually being used by the openldap module?

Puppetfile
```diff
@@ -79,3 +79,8 @@ mod 'datadog/datadog_agent', '1.6.0'
 
 # Used for grabbing certificates for jenkins.io
 mod 'danzilio/letsencrypt', '1.0.0'
+
+# For managing ldap
+mod 'camptocamp/openldap', '1.14.0'
```
rtyler, line 84: This might need to go into `.fixtures.yml` too so rspec-puppet knows how to handle it with tests

manifests/site.pp
```diff
@@ -81,3 +81,8 @@
 node 'cucumber' {
   include role::cucumber
 }
+
+# kale
+node 'kale' {
```
rtyler, line 86: I'm assuming this is a totally made up thing that you created locally right?

dist/profile/manifests/openldap.pp
```diff
@@ -0,0 +1,84 @@
...
+    proto  => 'tcp',
+    port   => 9636,
+    action => 'accept',
+  }
+
+  class { 'openldap::server': }
+  openldap::server::database { 'dc=jenkins-ci,dc=org':
+    directory => '/var/lib/ldap',
+    rootdn    => 'cn=admin,dc=jenkins-ci,dc=org',
+    rootpw    => hiera('ldap_rootpw'),
```
rtyler, line 72: In `hieradata/common.yaml` will you define `ldap_rootpw` with some garbage value? we'll override it in env/production.yaml before this gets deployed

dist/profile/manifests/openldap.pp
```diff
@@ -0,0 +1,84 @@
+#
+# Manage an OpenLDAP authentication service
+#
+class profile::ldap {
+  include ::firewall
+  include ::datadog_agent
+
+  profile::datadog_check { 'ldap-process-check':
```
rtyler, line 8: :+1:

rtyler:
@nibalizer does the openldap module support any kind of defining of groups in LDAP, or is that something we would need to `Exec`?

We've got a couple groups defined right now (e.g. `puppet-admins`) which I would like to move into Puppet-based management if possible

Puppetfile
nibalizer: Yes these are used by the openldap module. Note in the commit message we also need to install an augeas package to make the provider functional.

nibalizer:
@rtyler I don't see any support for entries in thee main DB inside this module https://github.com/camptocamp/puppet-openldap

I also tend to look at that as a data problem anyways. Would you puppet rows in your postgres database?

rtyler:
> I also tend to look at that as a data problem anyways. Would you puppet rows in your postgres database?

this is where this question gets uncomfortable for me, we have groups that are defined for authorization of other services which depend on LDAP.  I might not puppet rows in a database, but I'm not sure I wouldn't puppetize critical database schemas in a database. If there's not something ready-baked then I'm just as happy to gloss over the whole question for now and when we migrate machines we'll just have to make sure we document the setup scripts we need for a functional LDAP installation.

rtyler:
Looks like jenkins disagrees with the versions its trying to fetch from the forge:

```
Notice: Preparing to install into /scratch/jenkins/workspace/infra/jenkins-infra/spec/fixtures/modules ...
Notice: Downloading from https://forgeapi.puppetlabs.com ...
Error: Could not install 'camptocamp-openldaap' (v1.14.0)
  No releases matching '1.14.0' are available from https://forgeapi.puppetlabs.com
rake aborted!
```

PR #328, opened 2016-04-15
SSH key pair to be used to push bits into OSUOSL master mirror

kohsuke:


hieradata/common.yaml
```diff
@@ -87,6 +87,13 @@ profile::jenkinsadmin::nick_password: ENC[PKCS7,MIIBeQYJKoZIhvcNAQcDoIIBajCCAWYC
 # <https://ci.jenkins-ci.org/view/Infrastructure/job/infra_ircbot>
 profile::jenkinsadmin::image_tag: 'build50'
 
+# Key that Jenkins uses to push bits into OSUOSL
+profile::jenkins::osuosl-sshkey-private: |
```
rtyler: If we're going to add these as parameterized class variables to `profile::jenkins` then I think the hyphens will cause trouble since `$foo-bar` is not a valid variable declaration in Puppet.

I suggest renaming these to `profile::jenkins::mirroring_privkey` and `profile::jenkins::mirrorring_pubkey` respectively

PR #329, opened 2016-04-16
Adding an ops script that resets LDAP password en mass

kohsuke:


scripts/passwordReset.groovy
```diff
@@ -0,0 +1,103 @@
...
+		
+		
+		if (l=="") {
+		  if (mail!=null && username!=null) {
+		    side.println(mail+"\t"+username);
+		  }
+		  mail = null;
+		  username = null;
+		  normalUser = false;
+		  return;
```
rtyler: I can't remember my Groovy syntax well enough, but don't these `return` statements exit out of `main()`? Or do they just bounce to the next iteration of `eachLine`?

kohsuke: The latter. It only exists a closure.

PR #334, opened 2016-04-16
Enable the appropriate interfaces for the openldap server

rtyler:
References [INFRA-1](https://issues.jenkins-ci.org/browse/INFRA-1)


dist/profile/manifests/ldap.pp
```diff
@@ -19,7 +19,10 @@
       'libaugeas-ruby',          # for augeas based puppet providers
   ])
 
-  include openldap::server
+  class { 'openldap::server':
+    ldap_ifs  => ['127.0.0.1'],
+    ldaps_ifs => ['/'],
```
kohsuke, line 24: I think you still want `ldapi` here for ease of backup, etc.

rtyler, line 24: `ldapi` is [part of the module defaults](https://github.com/camptocamp/puppet-openldap/blob/master/manifests/server.pp#L24)

I'll add it as an explicit declaration in a followup pull request.

PR #335, opened 2016-04-16
[INFRA-638] Adding a managed l10n server

kohsuke:


dist/profile/manifests/l10n_server.pp
```diff
@@ -0,0 +1,44 @@
...
+
+  docker::image { $image:
+    image_tag => $image_tag,
+  }
+
+  docker::run { 'l10n':
+    volumes  => ['/home/l10n/data:/var/l10n'
+    ],
+    username => 'l10n',
+    image    => "${image}}:${image_tag}",
```
bwalding: double close quote }} is correct?

dist/profile/manifests/l10n_server.pp
```diff
@@ -0,0 +1,44 @@
...
+  File <| title == '/etc/init/docker-l10n.conf' |> {
+    notify  => Service['docker-l10n'],
+  }
+
+  user { $user:
+    shell      => '/bin/false',
+    managehome => true,
+  }
+
+  file { '/home/l10n/data':
```
rtyler: Since this is not a "real" user (like `kohsuke` or `rtyler`) could we move the home directory to somewhere else like `/srv/l10n` or `/var/lib/l10n`, that would make it more consistent with our other services like mirrorbrain or jenkins

kohsuke:
Addressed all the outstanding comments so far

dist/profile/manifests/l10n_server.pp
```diff
@@ -0,0 +1,45 @@
...
+  validate_string($image_tag)
+  $user = 'l10n'
+  $image = 'jenkinsciinfra/l10n-server'
+
+  docker::image { $image:
+    image_tag => $image_tag,
+  }
+
+  docker::run { 'l10n':
+    volumes  => ['/home/l10n/data:/var/l10n'
```
rtyler: Since the homedir is duplicated a few places, I suggest making it a variable like `$user` above

rtyler:
:+1:  feel free to merge when Jenkins gives it the okay

PR #367, opened 2016-04-17
Added a command that runs tests locally more quickly

kohsuke:
I can't possibly remember those commands

### DIFF
--- a/check
+++ b/check
@@ -1,3 +1,6 @@
 #!/bin/bash -ex
 bundle exec rake lint test-zonefile
+
+# this assumes that you have run `bundle exec rake spec_prep` at least once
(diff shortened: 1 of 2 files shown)


check
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -ex
+bundle exec rake lint test-zonefile
+bundle exec parallel_rspec spec/classes
```
rtyler: The thing about parallel_rspec is that it's not the same as running the `spec` task in Rake.

In order for this to work properly, you would have to make sure that you had run `bundle exec rake spec_prep` at least once before hand to prepare the `spec/fixtures` directory

PR #375, opened 2016-04-17
[INFRA-1] deploy LDAPS

kohsuke:


dist/profile/manifests/ldap.pp
```diff
@@ -56,6 +68,31 @@
   ###############
 
 
+  # SSL Certificates
+  file { $ssl_dir:
+    ensure => directory,
+    mode   => '0700',
+    owner  => $openldap::params::server_owner,
```
rtyler, line 75: This _should_ require the class `openldap` as a dependency just to make sure that the `/etc/ldap` directory is created by the package installation. We're using manifest ordering so I don't think this should be a problem for us though.

PR #379, opened 2016-04-17
[INFRA-638] l10n.jenkins.io takes over l10n.jenkins-ci.org

kohsuke:
new VM takes over the service previously hosted on cucumber


dist/profile/files/bind/jenkins-ci.org.zone
```diff
@@ -49,7 +49,7 @@ wiki        3600    IN    CNAME    lettuce
 updates     3600    IN    CNAME    cucumber
 downloads   3600    IN    CNAME    cucumber
 fisheye     3600    IN    CNAME    cucumber
-l10n        3600    IN    CNAME    cucumber
+l10n        3600    IN    CNAME    l10n.jenkins.io
```
rtyler: Since this is outside the zone, it should have a trailing period to it otherwise bind thinks this is "l10n.l10n.jenkins.io.jenkins-ci.org" (from my understanding)

PR #433, opened 2016-04-20
Codify some manual updates made to mirrors.jenkins.io

rtyler:
While working through release processes with @kohsuke last night, we discovered a number of previously unmanaged changes. To get the release out the door, we made changes on the host and took some notes.


rtyler:
I've been running some `--noop` runs on `mirrors.jenkins.io` to make sure that this is not doing anything new but rather codifying what we've already done.

With this most recent adjustments, it looks like this is ready to land in staging.

PR #470, opened 2016-04-29
[INFRA-677] define the census server

kohsuke:
Tested with vagrant to satisfaction.


hieradata/clients/census.yaml
```diff
@@ -0,0 +1,9 @@
+---
+lvm::volume_groups:
+  data:
+    physical_volumes:
+      - /dev/xvdb
```
rtyler, line 5: This might not be the device name of the EBS volume as mentioned in [INFRA-676](https://issues.jenkins-ci.org/browse/INFRA-676)

So I'll make sure that the machine is bootstrapped prior to merging this.

dist/profile/manifests/census.pp
```diff
@@ -0,0 +1,75 @@
...
+    source => "puppet:///modules/${module_name}/census/monthly-passwords",
+  }
+
+
+  file { '/var/log/apache2/census.jenkins.io':
+    ensure => directory,
+  }
+
+  apache::vhost { 'census.jenkins.io':
+    servername      => 'census.jenkins.io',
```
rtyler, line 60: fwiw when you specify a FQDN in the `name` parameter of the resource (`apache::vhost {'foo.example.com' }`) `servername` is then redundant. We only need it when the resource name cannot be used as the FQDN.

rtyler:
I'll pull this in once I have a machine provisioned in production which has the right block storage layout

PR #476, opened 2016-05-02
R10k webhook for deploying bits more automatically again

rtyler:


rtyler:
There's not much of a way to test this without a local puppet enterprise installation but a no-op run in production looks promising.

PR #492, opened 2016-05-15
Demo uid

kohsuke:
Run demo with custom /etc/passwd that sets uid for Jenkins to 2002.


rtyler:
Without an associated ticket and more details in the pull request, I genuinely have no idea why this change was made?

PR #508, opened 2016-06-06
Fix mirrorbrain rsync to use pkg.jenkins.io

abayer:
Don't merge 'til after https://github.com/jenkinsci/packaging/pull/55 is merged, and probably make sure we've got a pre-staged /var/www/pkg.jenkins.io.staging ready too.


recena:
:+1:

PR #509, opened 2016-06-07
Devnull redirected access logs for usage.jenkins-ci.org

rtyler:
These are unnecessary as the destination (usage.jenkins.io) will properly log
the request when it generates a 200 status response.

Fixes [INFRA-739](https://issues.jenkins-ci.org/browse/INFRA-739)


rtyler:
I have validated this against an ephemeral machine with the `usage` role and no access logs are created for the requests.

PR #524, opened 2016-06-15
Default to using the aufs storage driver, which is not garbage

rtyler:
Unlike devicemapper, which is total garbage. Seeing failures in newer ci.j.io
and trusted.ci.j.io nodes where `docker push` commands are failing

This requires the aufs module, which is generally provided by `linux-image-extra-$(uname -r)`. Which I'm still looking into a good approach for installing

### DIFF
--- a/dist/profile/manifests/docker.pp
+++ b/dist/profile/manifests/docker.pp
@@ -7,6 +7,7 @@ class profile::docker {
     # kernel modules on Ubuntu 12.04 LTS and restart the host machine anyways
     manage_kernel    => false,
     extra_parameters => '--storage-driver=aufs',
+    require          => Package['linux-image-extra'],
   }
 
   include datadog_agent::integrations::docker
@@ -22,5 +23,10 @@ class profile::docker {
     # traffic within docker is OK
     iniface => 'docker0',
     action  => 'accept',
+  }
+
+  package { 'linux-image-extra':
+    ensure => present,
+    name   => "linux-image-extra-${::kernelrelease}",
   }
 }
--- a/dist/profile/manifests/docker.pp
+++ b/dist/profile/manifests/docker.pp
@@ -2,10 +2,11 @@
 # Profile for managing basics of docker installation/configuration
 class profile::docker {
   class { '::docker':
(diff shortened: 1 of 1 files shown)


rtyler:
This needs some more manual testing, which I am doing now, to very the behavior when upgrading a host that uses `devicemapper` to `aufs` that the containers get restarted properly.

rtyler:
So it looks like the Docker daemon is getting restarted but then nothing restarts properly.

rtyler:
Testing with `role::spinach` and after restart the upstart logs for `docker-bind.log` are spewed with:

```
Error response from daemon: Could not find container for entity id 66b333da4951232c27fa0b3cfb5e8037839921a81f4389136dd35c978ba95b24
Error: No such image or container: bind
[]
Error: No such image or container: bind
[]
Error response from daemon: Could not find container for entity id 66b333da4951232c27fa0b3cfb5e8037839921a81f4389136dd35c978ba95b24
Error: No such image or container: bind
[]
Error: No such image or container: bind
[]
Error response from daemon: Could not find container for entity id 66b333da4951232c27fa0b3cfb5e8037839921a81f4389136dd35c978ba95b24

```

```
root@ip-10-236-143-33:/var/log/upstart# docker images
REPOSITORY              TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
jenkinsciinfra/bind     builddcfe94         9298d5ac7992        3 months ago        226.3 MB
jenkinsciinfra/ircbot   build50             2e6198c5c5d9        3 months ago        366.6 MB
root@ip-10-236-143-33:/var/log/upstart# docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
root@ip-10-236-143-33:/var/log/upstart# docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
root@ip-10-236-143-33:/var/log/upstart#
```

I'm going to leave this where it is for now and come back to it later when I have some fresher ideas

rtyler:
This was merged as part of #592

PR #537, opened 2016-07-15
Upload process was modified so this is no longer needed

kohsuke:
The rsync directly logs in as 'usagestats' as of now.


rtyler:
This won't remove the file from the filesystem unfortunately, generally I will push an `ensure => absent` file resource first, then purge the code at a later date.

Since this is small, and in your homedir, I think you can handle purging the file yourself :smiley_cat:

PR #545, opened 2016-07-21
Add a catch-all virtualhost with redirects for jenkins-ci.org

rtyler:
This change also includes some serverspec to validate some of the more important
redirects

References [INFRA-639](https://issues.jenkins-ci.org/browse/INFRA-639)


kohsuke:
:bee:

PR #546, opened 2016-07-22
Deploy the migration of jenkins-ci.org to eggplant

rtyler:


rtyler:
I ran a `--noop` on `eggplant` specifically and the run looks good to me.

PR #551, opened 2016-07-27
Puppetize the serving of javadocs for various Jenkins releases

rtyler:
The javadoc archives are generated every monday morning at 1AM by this job:
    https://github.com/jenkins-infra/javadoc/

Since the archives are rather big, and we only need to update after a regular
weekly release, this commit includes a cron to grab the latest successful build
every monday morning at 4AM

Fixes [INFRA-49](https://issues.jenkins-ci.org/browse/INFRA-49)


rtyler:
Tested this with `vagrant up eggplant` and everything provisions nicely with serverspec passing

PR #558, opened 2016-08-17
INFRA-896 fix deployment

kohsuke:


kohsuke:
Looks like it hit the bad moment of forge

```
Notice: Downloading from https://forgeapi.puppetlabs.com ...
Error: Request to Puppet Forge failed.
  The server being queried was /v3/files/puppetlabs-puppetserver_gem-0.2.0.tar.gz
  The HTTP response we received was '502 Bad Gateway'
rake aborted!
```

PR #560, opened 2016-08-19
Deploy slightly modified access logs for usage

rtyler:


rtyler:
A no-op deploy on usage works fine and creates the right configurations, landing this

PR #570, opened 2016-09-13
Enable the Blueocean plugin in our clusters

rtyler:


abayer:
:+1: Bwahahaha, this brings in `pipeline-model-definition` at the same time. =)

rtyler:
@abayer I would request that nobody start introducing declarative Pipelines into our infrastructure until there's bloody documentation on jenkins.io for it, otherwise we lose a valuable demonstration use

abayer:
@rtyler I agree completely. @i386 needs to hurry up and make a PR. =)

rtyler:
This change is now blocked because the blueocean-plugin requires 2.7.3 or later of core, which means we must upgrade the full Jenkins instance just for this

i386:
I was just thinking about those docs today. Yes, PR is coming :)

PR #591, opened 2016-09-29
hiera_hash() will actually perform the deep merge on hiera hashes

rtyler:
Somewhere the docs lied to me


rtyler:
Let it be known for posterity, that the [deep_merge gem is already installed on the Puppet master](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/manifests/puppetmaster.pp#L88-L93)

PR #592, opened 2016-09-29
Upgrade Docker and switch to the aufs storage driver

rtyler:
See [INFRA-935](https://issues.jenkins-ci.org/browse/INFRA-935) for more details.

This pull request also upgrades our garethr/docker module which we have been running a forked version of, and no longer need to do so.

### DIFF
--- a/dist/profile/manifests/docker.pp
+++ b/dist/profile/manifests/docker.pp
@@ -7,6 +7,7 @@ class profile::docker {
     # kernel modules on Ubuntu 12.04 LTS and restart the host machine anyways
     manage_kernel    => false,
     extra_parameters => '--storage-driver=aufs',
+    require          => Package['linux-image-extra'],
   }
 
   include datadog_agent::integrations::docker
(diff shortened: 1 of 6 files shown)


dist/profile/manifests/docker.pp
```diff
@@ -2,10 +2,11 @@
 # Profile for managing basics of docker installation/configuration
 class profile::docker {
   class { '::docker':
-    version       => '1.9.1',
     # Disabling the management of the kernel, since we have to pre-install
     # kernel modules on Ubuntu 12.04 LTS and restart the host machine anyways
```
batmat: So from what you say on IRC, that comment above isn't actual anymore right?

batmat (approved):
\o/ :+1:  FWIU

dist/profile/manifests/docker.pp
rtyler: I'll replace 12.04 with 14.04, because it actually still applies :P

PR #627, opened 2016-11-10
Update plugin site to new UI message

daniel-beck:
Builds are only identified by number and `plugin-site-api` revision, not `plugin-site` revision; so changes to UI only don't show in commit ID.


daniel-beck:
@rtyler

rtyler:
@daniel-beck I'm pretty sure you can assign pull requests in this repository, that works too :)

daniel-beck:
Hah!

PR #636, opened 2016-11-15
[INFRA-977] fallback server went offline

kohsuke:
According to Tyler, we lost spinach and this server is dead.

This failure prevented the rest of the sync script to run till completion.


kohsuke:
I suspect the omission of the `--times` option was unintentional.

rtyler (approved):

PR #665, opened 2017-01-04
Increase Jenkins HTTP keepalive timeout

daniel-beck:
Untested, but https://github.com/jenkinsci/docker/blob/3cf0212a1a795902cabbd67b2920f4bedeb762fc/jenkins.sh#L22 looks like this should work.

People with slower connections hit the overly aggressive timeout in [JENKINS-39909](https://issues.jenkins-ci.org/browse/JENKINS-39909)

Reference:
https://botbot.me/freenode/jenkins/2017-01-04/?msg=78825222&page=5

CC @jimklimov

### DIFF
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -69,6 +69,7 @@ class profile::buildmaster(
       'HOME=/var/jenkins_home',
       'USER=jenkins',
       'JAVA_OPTS="-Duser.home=/var/jenkins_home -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2"',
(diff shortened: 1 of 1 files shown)


rtyler:
@daniel-beck This looks fine to me but I'm concerned by the idea of people frequently downloading artifacts from ci.jenkins.io, is that really a common thing?

jimklimov:
@rtyler : In my case at least it is not "frequently", but still I am unable to progress checking another PR because I can't download the `jenkins.war` to deploy. Tried dozens of times in a loop by now - and it is always broken (connection aborted roughly every minute and a half, despite progressing at a non-zero speed, and the resumption loses something so checksums mismatch and ZIP structure is invalid).

daniel-beck:
@rtyler 

> is that really a common thing?

Happens occasionally for testing PR builds. And having a canonical build for a change helps in testing and reproducing possible problems.

PR #671, opened 2017-01-11
Add dummy and production SSH keys for management of k8s with Jenkins integration code

rtyler:
It's a little difficult to tell how well this is going to work in production,
but as far as I am aware the snippet attached is properly configuring a
credential for use by the jenkins-infra/azure Jenkinsfile and other Jenkins
Pipelines which may orchestrate k8s workloads

References [INFRA-1002](https://issues.jenkins-ci.org/browse/INFRA-1002)


rtyler:
If you have some time to review @olblak, please do.

hieradata/common.yaml
```diff
@@ -248,6 +248,38 @@ profile::pluginsite::image_tag: '24-0d1adf'
 azure::releases::account_name: 'overridewithstorageaccountname'
```
olblak, line 250: What does 'releases' mean here?
release storage account? 
''${PREFIX}jenkinsrelease ?

olblak:
LGTM

hieradata/common.yaml
rtyler: Yes, i'll update that with a comment and merge

PR #674, opened 2017-01-17
Add missing trailing slash to non-TLS plugin site redirect.

orrc:
Otherwise, the redirect to port 443 goes awry:

    $ curl -sI http://plugins.jenkins.io/git | grep Location
    Location: https://plugins.jenkins.iogit

Added an rspec test, similar to other vhosts, but not sure whether adding an integration test is possible…


orrc:
Transient test failure:
```
Error: Request to Puppet Forge failed.
  The server being queried was https://forgeapi.puppetlabs.com/v3/releases?module=puppetlabs-stdlib
  The HTTP response we received was '502 Bad Gateway'
rake aborted!
```

PR #687, opened 2017-02-15
AllowOverride FileInfo for jenkins.io for custom error pages

daniel-beck:
i have no idea what i m doing

* Untested, but based on http://httpd.apache.org/docs/current/mod/core.html#allowoverride looks reasonable.
* This file is a mess as it still refers to this site as beta (AFAIU), so I'm only somewhat confident I edited the correct thing.

### DIFF
--- a/dist/profile/manifests/staticsite.pp
+++ b/dist/profile/manifests/staticsite.pp
@@ -118,6 +118,7 @@ class profile::staticsite(
     port          => '443',
     ssl           => true,
     docroot       => $beta_docroot,
+    override      => 'FileInfo',
     require       => File[$beta_docroot],
   }
 
--- a/spec/classes/profile/staticsite_spec.rb
+++ b/spec/classes/profile/staticsite_spec.rb
@@ -40,7 +40,7 @@ describe 'profile::staticsite' do
                 .with(:docroot => '/srv/jenkins.io/current') }
 
     it { should contain_apache__vhost('jenkins.io')
-                .with(:docroot => '/srv/jenkins.io/beta') }
+                .with( { :docroot => '/srv/jenkins.io/beta', :override  => ['FileInfo'] } ) }
 
     it 'should upgrade non-TLS to TLS' do
       expect(subject).to contain_apache__vhost('jenkins.io unsecured').with({


rtyler:
What's the usecase for this? I'm strongly against people dumping `.htaccess` files into jenkins.io.git, the fact that its served by Apache should be an implementation detail

daniel-beck:
@rtyler 

> What's the usecase for this?

Showing angry Jenkins, obviously:

screen shot

---

In all seriousness, we can't really keep all obsolete URLs around as redirects indefinitely. Allowing the definition of a (or possibly multiple, e.g. `/doc`, or even `/doc/developer` specific) 404 error page(s) will allow us to direct people to likely locations without the jarring experience of the default error page.

Alternatively, we could probably define a fixed 404 page directly here. That would not be as nice since we lose the flexibility I mention above, but it would be a step in the right direction -- and would not need support from `jenkins.io.git` besides the contract that e.g. `/404/index.html` needs to exist.

rtyler:
If you're suggesting the creation of a custom 404 page, that's fine.

Opening up `.htaccess` inside `jenkins.io.git` means that people who don't understand Apache can submit changes which dramatically affect the routing of requests and is not something I plan on accepting.

daniel-beck:
@rtyler 

> people who don't understand Apache can submit changes

Isn't this pretty tightly controlled with maybe a handful of people able to merge? Not clear to me where the problem is here besides theoretical.

> If you're suggesting the creation of a custom 404 page, that's fine.

Well, yes. See the linked PR. I opted to go with `AllowOverride` to allow configuring different `ErrorDocument`s per directory further down the road. E.g. a general one (that could look almost like a site map – e.g. Apple had 404 error pages that looked a lot like http://www.apple.com/sitemap/ for a while), and maybe one for `/doc` that only suggests locations inside that directory.

I can probably rework this if you want one fixed error page for the site.

daniel-beck:
@rtyler I just found [`AllowOverrideList`](http://httpd.apache.org/docs/current/mod/core.html#allowoverridelist).

Seems like this is what would work for you?

    AllowOverride None
    AllowOverrideList ErrorDocument

rtyler:
> Isn't this pretty tightly controlled with maybe a handful of people able to merge? Not clear to me where the problem is here besides theoretical.

You can merge jenkins.io, and you literally started this PR with "I have no idea what I'm doing." :smile: 


Therein lies the problem :smile_cat: 


> Seems like this is what would work for you?


Yes, AllowedOverrideList only for ErrorDocument would be acceptable. HOWEVER. if we find ourselves needing to switch away from Apache for serving jenkins.io, we're going to move ahead with it and let somebody else deal with updating those `.htaccess` files.

daniel-beck:
> we're going to move ahead with it

I think we can deal with not having fancy looking error pages for a bit if that happens.

PR #700, opened 2017-03-08
[INFRA-583] Add Datadog check for SSL certificate enddate monitoring

olblak:
This PR add ssh enddate monitoring. 
It inspired by this [blogpost](https://workshop.avatarnewyork.com/project/datadog-ssl-expires-check/)

I just adapt the script to monitor https website instead of 'local' certificate
Puppetmaster is configure to run this check but we can use another role.

Just update 'profile::datadog_ssl_check::sites:' within hieradata/common.yaml with the correct list of website.

### DIFF
--- a/dist/profile/files/datadog_ssl_check/ssl_check_expire_days.py
+++ b/dist/profile/files/datadog_ssl_check/ssl_check_expire_days.py
@@ -11,6 +11,7 @@ class SSLCheckExpireDays(AgentCheck):
     def check(self, instance):
         metric = "ssl.expire_in_days"
         site = instance['site']
+        tag = "site:"+site
         p = subprocess.Popen("echo | openssl s_client -showcerts -servername " + site + " -connect " + site + ":443 2>/dev/null | openssl x509 -noout -dates | grep notAfter | cut -f 2 -d\= | xargs -0 -I arg date -d arg \"+%s\"",stdout=subprocess.PIPE, shell=True)
         (output, err) = p.communicate()
         if output:
@@ -19,8 +20,6 @@ class SSLCheckExpireDays(AgentCheck):
             d1 = int(output)
             delta = d1 - d0
             days= delta/24/60/60 # convert the timestamp to days
-            tag="site:" + site # generate the tags
-#            print "metric: " + str(metric) + ", tag: " + str(tag) + ", days: " + str(days)
             self.gauge(metric, int(days), tags=[tag])
         else:
-            self.gauge(metric, -1, tags=[site])
+            self.gauge(metric, -1, tags=[tag])
--- a/dist/profile/files/datadog_ssl_check/ssl_check_expire_days.py
+++ b/dist/profile/files/datadog_ssl_check/ssl_check_expire_days.py
@@ -1,0 +1,26 @@
+# A rewrite of https://workshop.avatarnewyork.com/project/datadog-ssl-expires-check/
+# I prefer to test the site itself instead of the ssl cert file
+
+import time
+import datetime
+import subprocess
+import sys
+from checks import AgentCheck
+
+class SSLCheckExpireDays(AgentCheck):
+    def check(self, instance):
+        metric = "ssl.expire_in_days"
+        site = instance['site']
+        p = subprocess.Popen("echo | openssl s_client -showcerts -servername " + site + " -connect " + site + ":443 2>/dev/null | openssl x509 -noout -dates | grep notAfter | cut -f 2 -d\= | xargs -0 -I arg date -d arg \"+%s\"",stdout=subprocess.PIPE, shell=True)
+        (output, err) = p.communicate()
+        if output:
+            output = output.rstrip("\n")
+            d0 = int(time.time())
+            d1 = int(output)
+            delta = d1 - d0
+            days= delta/24/60/60 # convert the timestamp to days
+            tag="site:" + site # generate the tags
+#            print "metric: " + str(metric) + ", tag: " + str(tag) + ", days: " + str(days)
+            self.gauge(metric, int(days), tags=[tag])
+        else:
+            self.gauge(metric, -1, tags=[site])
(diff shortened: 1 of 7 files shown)


dist/profile/files/datadog_ssl_check/ssl_check_expire_days.py
```diff
@@ -0,0 +1,26 @@
...
+            output = output.rstrip("\n")
+            d0 = int(time.time())
+            d1 = int(output)
+            delta = d1 - d0
+            days= delta/24/60/60 # convert the timestamp to days
+            tag="site:" + site # generate the tags
+#            print "metric: " + str(metric) + ", tag: " + str(tag) + ", days: " + str(days)
+            self.gauge(metric, int(days), tags=[tag])
+        else:
+            self.gauge(metric, -1, tags=[site])
```
orrc: Is the `tags` value meant to be different between the two branches?

hieradata/common.yaml
```diff
@@ -290,6 +290,19 @@ azure::k8s::management_ssh_privkey: |
...
+    - wiki.jenkins-ci.org
+    - jenkins.io
+    - www.jenkins.io
+    - ci.jenkins.io
+    - mirrors.jenkins.io
+    - accounts.jenkins.io
+    - usage.jenkins.io
+    - pkg.jenkins.io
+    - updates.jenkins.io
+    - plugins.jenkins.io
```
orrc: rating.jenkins.io is another.

Minor: Alphabetical order would probably be nicer.

dist/profile/files/datadog_ssl_check/ssl_check_expire_days.py
```diff
@@ -0,0 +1,26 @@
...
+        p = subprocess.Popen("echo | openssl s_client -showcerts -servername " + site + " -connect " + site + ":443 2>/dev/null | openssl x509 -noout -dates | grep notAfter | cut -f 2 -d\= | xargs -0 -I arg date -d arg \"+%s\"",stdout=subprocess.PIPE, shell=True)
+        (output, err) = p.communicate()
+        if output:
+            output = output.rstrip("\n")
+            d0 = int(time.time())
+            d1 = int(output)
+            delta = d1 - d0
+            days= delta/24/60/60 # convert the timestamp to days
+            tag="site:" + site # generate the tags
+#            print "metric: " + str(metric) + ", tag: " + str(tag) + ", days: " + str(days)
```
rtyler: This commented line should be removed

dist/profile/manifests/datadog_ssl_check.pp
```diff
@@ -0,0 +1,29 @@
...
+class profile::datadog_ssl_check (
+  $sites = undef,
+){
+
+  include datadog_agent
+
+  file { 'ssl_check_expire_days.py':
+    ensure => present,
+    source => "puppet:///modules/${module_name}/datadog_ssl_check/ssl_check_expire_days.py",
+    path   => '/etc/dd-agent/checks.d/ssl_check_expire_days.py',
```
rtyler, line 16: I believe this should have a `require` that the `dd-agent` package has been installed to make sure this directory will be created before this File resource executes.

It might be enough to `require => Class[datadog_agent]`, but I'm not sure off the top of my head

dist/profile/manifests/datadog_ssl_check.pp
```diff
@@ -0,0 +1,29 @@
...
+    group  => $::datadog_agent::params::dd_group,
+    notify => Service['datadog-agent']
+  }
+
+  file { 'ssl_check_expire_days.yaml':
+    ensure  => present,
+    content => template("${module_name}/datadog_ssl_check/ssl_check_expire_days.yaml.erb"),
+    owner   => $::datadog_agent::params::dd_user,
+    group   => $::datadog_agent::params::dd_group,
+    path    => "${::datadog_agent::params::conf_dir}/ssl_check_expire_days.yaml",
```
rtyler, line 27: Same comment as the previous File resource in this manifest

dist/profile/manifests/datadog_ssl_check.pp
```diff
@@ -0,0 +1,29 @@
+#
+# Define datadog check for ssl expiration
+# Inspired by: https://workshop.avatarnewyork.com/project/datadog-ssl-expires-check/
+#
+
+class profile::datadog_ssl_check (
+  $sites = undef,
```
rtyler: Instead of `undef`, I recommend making the default an empty array here, that way there won't be a Puppet error then the ERB template below is evaluated (`sites.each`)

rtyler (changes requested):
Looks promising, some minor feedback :+1:

dist/profile/files/datadog_ssl_check/ssl_check_expire_days.py
olblak: nope, it should be the same, I'll fix that

dist/profile/manifests/datadog_ssl_check.pp
olblak: Good comment

dist/profile/manifests/datadog_ssl_check.pp
olblak: I am not sure to understand your point, I don't see any error if the value is define but I would llike to have a warning if the value is not defined.

olblak: I added a default value

olblak:
As @orrc mentionned, I now use the same tag format in all cases.
As @rtyler suggested, I added a requirement on class 'datadog_agent'

This PR is ready to be merged

dist/profile/manifests/datadog_ssl_check.pp
```diff
@@ -0,0 +1,30 @@
+#
+# Define datadog check for ssl expiration
+# Inspired by: https://workshop.avatarnewyork.com/project/datadog-ssl-expires-check/
+#
+
+class profile::datadog_ssl_check (
+  $sites = ["N/A"]
```
rtyler, line 7: This is failing puppet lint fwiw

PR #701, opened 2017-03-09
[INFRA-672] Create a CNAME and vhost for wiki.jenkins.io

olblak:
Add wiki.jenkins.io vhost and update bind configuration 
We still have to configure in confluence admin page, the vhost that we want to use (wiki.jenkins.io vs wiki.jenkins-ci.org)

### DIFF
--- a/dist/profile/files/bind/jenkins.io.zone
+++ b/dist/profile/files/bind/jenkins.io.zone
@@ -2,7 +2,7 @@
 
 ; SOA Record
 @  1D  IN  SOA    ns1.jenkins-ci.org. tyler.monkeypox.org. (
-                2017030801 ; serial, todays date + todays serial #
+                2015122201 ; serial, todays date + todays serial #
                 28800      ; refresh, seconds
                 7200       ; retry, seconds
                 604800     ; expire, seconds
--- a/dist/profile/files/bind/jenkins.io.zone
+++ b/dist/profile/files/bind/jenkins.io.zone
@@ -2,7 +2,7 @@
 
 ; SOA Record
 @  1D  IN  SOA    ns1.jenkins-ci.org. tyler.monkeypox.org. (
-                2015122201 ; serial, todays date + todays serial #
+                2017030801 ; serial, todays date + todays serial #
                 28800      ; refresh, seconds
                 7200       ; retry, seconds
                 604800     ; expire, seconds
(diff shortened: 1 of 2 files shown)


orrc:
I don't know whether redirecting https?://wiki.jenkins-ci.org/.* to https://wiki.jenkins.io was intended to be part of this ticket, but at the least, a Let's Encrypt certificate should be generated for the new hostname.

dist/profile/files/bind/jenkins.io.zone
```diff
@@ -2,7 +2,7 @@
 
 ; SOA Record
 @  1D  IN  SOA    ns1.jenkins-ci.org. tyler.monkeypox.org. (
-                2015122201 ; serial, todays date + todays serial #
+                2017030801 ; serial, todays date + todays serial #
```
rtyler: This doesn't need to change, the SOA record needn't be updated here.

rtyler (changes requested):
This should include letsencrypt changes

dist/profile/files/bind/jenkins.io.zone
olblak: ? What's the reason why we don't have to update SOA serial number?

olblak: Anyway I revert it

olblak:
@rtyler I reverted SOA serial number and I added letsencrypt certificate for wiki.jenkins.io
As @orrc suggested, I also redirect http://wiki.jenkins-ci.org to http://wiki.jenkins.io.
But If the goal is to use wiki.jenkins.io, we should also redirect https://wiki.jenkins-ci.org to https://wiki.jenkins.io

I think that we also have to configure the vhost inside confluence which mean that we need to update it in three steps:

1. Send wiki.jenkins.io requests to confluence
2. Update confluence config with wiki.jenkins.io instead of wiki.jenkins-ci.org
3 Stop sending wiki.jenkins-ci.org requests to confluence

dist/profile/files/bind/jenkins.io.zone
rtyler: @olblak The reason you needn't update the serial for the SOA record is because that record doesn't need to change for this. The SOA record is only (AFAIK) used to describe the authority of the zone, which has long since been established as `ns1.jenkins-ci.org`. and that's not changing here.

I don't think updating it is necessarily bad, but unnecessary when adding new unrelated A or CNAME records.

rtyler (approved):
This looks good to me, I want to do some manual testing myself before rolling it to production though.

olblak:
Is there something I can do to help?

PR #702, opened 2017-03-10
Kubernetes/datadog

olblak:
This PR is a proof of concept to use puppet in order to deploy datadog agent pods on kubernetes.

In order to test it, you need to define following variables based on .kube/config

`
profile::kubernetes::kubectl::certificate_authority_data:

profile::kubernetes::kubectl::server:

profile::kubernetes::kubectl::clustername:

profile::kubernetes::kubectl::username: 

profile::kubernetes::kubectl::client_certificate_data: 

profile::kubernetes::kubectl::client_key_data:
 `

Then you can use vagrant to apply configuration.
`vagrant up kubernetes && vagrant provision kubernetes`

Main advantages of this approach (instead of [k8s](https://github.com/jenkins-infra/k8s)) are:
1. We can reuse the same workflow to deploy in production,
2. We can use puppet to template kubernetes configuration files
3. We don't have to defined secrets in other places


dist/profile/manifests/kubernetes/resources/datadog.pp
```diff
@@ -0,0 +1,62 @@
...
+
+  $daemonset = loadyaml("${profile::kubernetes::params::resources}/datadog/daemonset.yaml")
+  # Search app label value
+  if ! empty($daemonset){
+    $label_app = $daemonset['spec']['template'][metadata][labels][app]
+  }
+
+  exec { 'apply secret':
+    command     => "kubectl apply -f ${profile::kubernetes::params::resources}/datadog/secret.yaml",
+    path        => ["${profile::kubernetes::params::bin}/"],
```
rtyler: There's no need for the string interpolation here

dist/profile/manifests/kubernetes/resources/datadog.pp
```diff
@@ -0,0 +1,62 @@
...
+    environment => ["KUBECONFIG=${profile::kubernetes::params::home}/.kube/config"] ,
+    logoutput   => true,
+  }
+
+  exec { 'apply daemonset':
+    command     => "kubectl apply -f ${profile::kubernetes::params::resources}/datadog/daemonset.yaml",
+    path        => ["${profile::kubernetes::params::bin}/"],
```
rtyler: There's no need for the string interpolation here

dist/profile/manifests/kubernetes/resources/datadog.pp
```diff
@@ -0,0 +1,62 @@
...
+    environment => ["KUBECONFIG=${profile::kubernetes::params::home}/.kube/config"] ,
+    logoutput   => true,
+  }
+
+  # Only delete pods if secrets were udpated and daemonset already exist
+  exec { 'reset daemonset pods':
+    path        => ["${profile::kubernetes::params::bin}/"],
```
rtyler: There's no need for the string interpolation here

dist/profile/manifests/kubernetes/resources/datadog.pp
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  exec { 'apply secret':
+    command     => "kubectl apply -f ${profile::kubernetes::params::resources}/datadog/secret.yaml",
+    path        => ["${profile::kubernetes::params::bin}/"],
+    environment => ["KUBECONFIG=${profile::kubernetes::params::home}/.kube/config"] ,
+    logoutput   => true,
+  }
+
+  exec { 'apply daemonset':
```
rtyler: This resource should have either an `unless` or `onlyif` statement to ensure that it doesn't execute `kubectl apply` on every Puppet agent run

dist/profile/manifests/kubernetes/resources/datadog.pp
```diff
@@ -0,0 +1,62 @@
...
+    owner   => $profile::kubernetes::params::user,
+  }
+
+  $daemonset = loadyaml("${profile::kubernetes::params::resources}/datadog/daemonset.yaml")
+  # Search app label value
+  if ! empty($daemonset){
+    $label_app = $daemonset['spec']['template'][metadata][labels][app]
+  }
+
+  exec { 'apply secret':
```
rtyler: This resource should have either an `unless` or `onlyif` statement to ensure that it doesn't execute `kubectl apply` on every Puppet agent run

Also, this Exec resource has the global name of "apply secret" which could conflict in the future with other resources. I suggest namnig the resource "datadog: apply secret". This also applies to other `exec` resources.

dist/profile/manifests/kubernetes/resources/datadog.pp
```diff
@@ -0,0 +1,62 @@
...
+  file { 'datadog_daemonset':
+    ensure  => 'present',
+    content => template("${module_name}/kubernetes/resources/datadog/daemonset.yaml.erb"),
+    path    => "${profile::kubernetes::params::resources}/datadog/daemonset.yaml",
+    owner   => $profile::kubernetes::params::user,
+  }
+
+  $daemonset = loadyaml("${profile::kubernetes::params::resources}/datadog/daemonset.yaml")
+  # Search app label value
+  if ! empty($daemonset){
```
rtyler: IMHO this `if` conditional should have an `else` statement that fails the Puppet run (`fail()` IIRC). If the `$label_app` doesn't get set here, then the "reset pods" will not work properly and fail won't it?

dist/profile/manifests/kubernetes/resources/datadog.pp
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  file { 'datadog_secret':
+    ensure  => 'present',
+    content => template("${module_name}/kubernetes/resources/datadog/secret.yaml.erb"),
+    path    => "${profile::kubernetes::params::resources}/datadog/secret.yaml",
+    owner   => $profile::kubernetes::params::user,
+  }
+
+  file { 'datadog_daemonset':
```
rtyler: The `name` ("datadog_daemonset") shoudl be changed to `"${profile::kubernetes::params::resources}/datadog/daemonset.yaml"` instead of using `path` below. The `path` parameter is unnecessary.

Same applies to the other `file` resources in this pull request

dist/profile/manifests/kubernetes/kubectl.pp
```diff
@@ -0,0 +1,56 @@
+# Ensure kubectl is installed and correctly configured
+
+class profile::kubernetes::kubectl (
+    $user = $profile::kubernetes::params::user,
```
rtyler: Please document these parameters and what they should be.

dist/profile/manifests/kubernetes/resources/datadog.pp
```diff
@@ -0,0 +1,62 @@
+# Deploy datadog resources on kubernetes cluster
+class profile::kubernetes::resources::datadog (
```
rtyler: This whole class might make better sense as a defined resource, see `datadog_check.pp` as an example. Then the usage could be parameterized to point to secrets, and resource names in the templates directory

olblak: I don't think that datadog class should move to a definition, but we can clearly 'extract' a part of it. 
which configure and apply kubernetes resource

dist/profile/manifests/kubernetes/resources/datadog.pp
olblak: ack

dist/profile/manifests/kubernetes/resources/datadog.pp
olblak: I am not using 'unless' or 'onlyif' as I only trigger exec based on events'

olblak: Good point with $title

rtyler:
I should be merging this no later than this weekend. I want to prepare the hiera configuration with the proper secrets so we can deploy this immediately, instead of stranding it in the `staging` branch for a few days.

olblak:
Good news :)

olblak:
I add a datadog agent to collect metrics from kubernetes api

olblak:
Is there something I can do to help?

PR #709, opened 2017-03-28
Kubernetes/registry [Require PR 702]

olblak:
This PR add docker registry configuration in order to pull docker images from private registry.
In order to use it we need to configure following variable
profile::kubernetes::resources::registry::dockerconfigjson:
with something like '{"auths":{"registry_endpoint":{"auth":"registry_crednential"}}}'
-> https://kubernetes.io/docs/concepts/containers/images/

! This PR require https://github.com/jenkins-infra/jenkins-infra/pull/702


olblak:
Checks failed because jenkins use the wrong version of puppetlabs/puppetlabs-stdlib

PR #719, opened 2017-04-02
Include the kubernetes role in puppetmaster

rtyler:
This also includes putting puppetlabs/stdlib 4.15.0 in .fixtures.yml to run the
tests with the proper puppetlabs/stdlib version


rtyler:
In theory, if all goes well, @olblak will finally be unblocked to make k8s things happen :fireworks:

PR #753, opened 2017-05-19
[INFRA-1178] Bump Jira container version

olblak:
This new version increase the java max heap size allocated to Jira


oleg-nenashev (approved):
This PR cannot be reviewed without a link to the upstream one, but 🚢 :it:

PR #755, opened 2017-06-01
Block all requests to api/json

rtyler:
I don't have anything in place here for checking to see if a request is
authenticated or not, which would be preferable. Failing that, I believe it's
important to prevent anonymous users from denying service to ci.jenkins.io in
the meantime.

Fixes INFRA-1204

### DIFF
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -306,6 +306,14 @@ RewriteEngine on
 RewriteCond %{HTTP_USER_AGENT} YisouSpider|Catlight*|CheckmanJenkins [NC]
 RewriteRule ^.* \"https://jenkins.io/infra/ci-redirects/\"  [L]
 
+# Block requests which attempt to access API endpoints and end up hanging
+# Jenkins. Really Jenkins should have a permission to block Anonymous/API
+# access but it doesn't, so Apache it is.
+# See also: https://issues.jenkins-ci.org/browse/INFRA-1204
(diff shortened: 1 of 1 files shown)


dist/profile/manifests/buildmaster.pp
```diff
@@ -306,6 +306,14 @@
 RewriteCond %{HTTP_USER_AGENT} YisouSpider|Catlight*|CheckmanJenkins [NC]
 RewriteRule ^.* \"https://jenkins.io/infra/ci-redirects/\"  [L]
 
+# Block requests which attempt to access API endpoints and end up hanging
+# Jenkins. Really Jenkins should have a permission to block Anonymous/API
+# access but it doesn't, so Apache it is.
+# See also: https://issues.jenkins-ci.org/browse/INFRA-1204
+RewriteCond %{REQUEST_FILENAME} ^(.*)api/json(.*)$ [NC]
```
orrc, line 313: Probably should also include `/api/xml`?

rtyler: @orrc For completeness' sake, perhaps. None of the slow requests that I found were hitting api/xml though.


@olblak Since i'm going to be offline for a bit, would you mind adjusting the PR accordingly and getting it into production?

rtyler:
Well, I'm disappointed this didn't get looked at while I was out. So I'm going to just merge it as is, and if we see `api/xml` requests crop up as problematic, I'll killdoze those too

rtyler:
Well, I'm disappointed this didn't get looked at while I was out. So I'm going to just merge it as is, and if we see `api/xml` requests crop up as problematic, I'll killdoze those too

PR #763, opened 2017-06-09
Turn off the pluginsite container and vhost on kelp

rtyler:
Once this is in production, this code can be deleted.

Fixes INFRA-1157


olblak:
This looks good to me

PR #764, opened 2017-06-09
Refactor the Azure CLI code into its own profile for inclusion in base

rtyler:
This is one of a few steps to get broader Azure-based backups published

Fixes INFRA-1148

### DIFF
--- a/dist/profile/manifests/mirrorbrain.pp
+++ b/dist/profile/manifests/mirrorbrain.pp
@@ -94,8 +94,6 @@ class profile::mirrorbrain (
 
   ## Files for Azure blob storage sync
   ##########################
-  ensure_packages(['python-pip'])
-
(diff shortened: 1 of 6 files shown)


dist/profile/manifests/mirrorbrain.pp
```diff
@@ -28,6 +28,7 @@
   }
 
   include profile::apachemisc
+  include profile::azure
```
olblak: Does it make sense to include profile::azure in mirrorbrain if it's already defined in base?

olblak (approved):
Excepted my comment about the duplicated include.
This PR looks good and it will very useful.

dist/profile/manifests/mirrorbrain.pp
rtyler: Nope

PR #780, opened 2017-06-20
[INFRA-1236] Add datadog check to monitor pluginsite api

olblak:
Once this check start to report data, we can monitor pluginsite api with datadog


spec/classes/profile/datadog_pluginsite_check.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
+
+describe 'profile::datadog_ssl_check' do
```
orrc: Should be `datadog_pluginsite_check`.

olblak: Thanks for catching this

PR #781, opened 2017-06-23
[INFRA-1241]  Redirect wiki.jenkins-ci.org to wiki.jenkins.io

olblak:
* Set permanent domain redirection at apache level.
* Add various tests for class profile::confluence

### DIFF
--- a/dist/profile/manifests/confluence.pp
+++ b/dist/profile/manifests/confluence.pp
@@ -94,31 +94,11 @@ class profile::confluence (
     restart_service => true,
   }
 
-  ### to put maintenance screen up, comment out the following and comment in the apache::vhost for https://jenkins-ci.org
-  ### #if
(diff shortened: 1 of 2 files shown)


dist/profile/manifests/confluence.pp
```diff
@@ -94,31 +94,11 @@
...
-    require         => File['/var/log/apache2/wiki.jenkins-ci.org'],
-  }
-  ### #endif
-
   # We can only acquire certs in production due to the way the letsencrypt
   # challenge process works
   if (($::environment == 'production') and ($::vagrant != '1')) {
     letsencrypt::certonly { 'wiki.jenkins.io':
-        domains     => ['wiki.jenkins.io'],
+        domains     => ['wiki.jenkins.io','wiki.jenkins-ci.org'],
```
rtyler, line 101: My only concern is that this won't trigger a certificate renewal automatically. I'm making a mental note to check that when we push this to production

PR #782, opened 2017-06-26
Upgrade plugin site to 65-36a0e4

michaelmccaskill:
Includes:

* [WEBSITE-363](https://issues.jenkins-ci.org/browse/WEBSITE-363)


olblak:
@michaelmccaskill Is this upgrade introduce the index health check status with a HEADER?

michaelmccaskill:
@olblak No. My understanding from reading [INFRA-1236](https://issues.jenkins-ci.org/browse/INFRA-1236) was you intended on performing the logic on your end so the header is no longer needed. Is that not correct?

rtyler:
Regardless of whether this includes the header check, it includes a good bugfix so I'm going to merge.

PR #788, opened 2017-06-26
More stabs in the dark :/

rtyler:


rtyler:
I'm really not pleased with this stab in the dark, but there is not much other difference between the wiki.jenkins.io vhost and the wiki.jenkins-ci.org vhost, and the latter isn't being configured correctly for some reason :frowning:

PR #791, opened 2017-06-28
Add checkstyle and findbugs plugins to ci

rsandell:
Add checkstyle and findbugs

It seems like dependencies are not needed in this list, so I skipped adding those. But this would then also add matrix and maven unfortunately.


rtyler (approved):

rsandell:
I'm writing some additions to buildPlugin that will enable easy access to these plugins.

PR #793, opened 2017-06-28
[INFRA-1252] Add various datadog http checks

olblak:
* Bump datadog module to 1.10.0
    * Rename datadog_agent::integrations::docker to datadog_agent::integrations::docker_daemon
      as it's deprecated since datadog module 1.7.0
    * Create class profile::datadog_http_check which define all http_checks
    * Remove datadog http_check from accountapp/confluence/jira
    * Remove profile::datadog_ssl_check as it's now included in http_check from datadog module

### DIFF
--- a/dist/profile/manifests/datadog_http_check.pp
+++ b/dist/profile/manifests/datadog_http_check.pp
@@ -1,0 +1,171 @@
+# Class: profile::datadog_http_check
+#
+# This class will install the necessary config to monitor jenkins infrastructure https endpoint
+# Parameters:
+#     contact
+#       Specify a list of users to notify regarding http_check alert
+#
+#     days_critical
+#     days_warning
+#        The check_certificate_expiration will instruct the check
+#        to create a service check that checks the expiration of the
+#        ssl certificate. Allow for a warning to occur when x days are
+#        left in the certificate, and alternatively raise a critical
+#        warning if the certificate is y days from the expiration date.
+#        The SSL certificate will always be validated for this additional
+#         service check regardless of the value of disable_ssl_validation
+#
+#     timeout
+#        The (optional) timeout in seconds.
+#
+#     threshold
+#     window
+#        The (optional) window and threshold parameters allow you to trigger
+#        alerts only if the check fails x times within the last y attempts
+#        where x is the threshold and y is the window.
+#
+
+class profile::datadog_http_check(
+  $contact = ['pagerduty'],
+  $days_warning = 30,
+  $days_critical = 10,
+  $timeout = 1,
+  $treshold = 3,
+  $window = 5
+){
+
+  include datadog_agent
+
+  class { 'datadog_agent::integrations::http_check':
+    instances => [{
+      'sitename'                     => 'plugins.jenkins.io',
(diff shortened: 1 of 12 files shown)


dist/profile/manifests/datadog_http_check.pp
```diff
@@ -0,0 +1,171 @@
...
+  $days_warning = 30,
+  $days_critical = 10,
+  $timeout = 1,
+  $treshold = 3,
+  $window = 5
+){
+
+  include datadog_agent
+
+  class { 'datadog_agent::integrations::http_check':
```
rtyler, line 39: Collecting the checks onto the Puppet master I like!

However, I'm not fond of this approach because it puts the code for monitoring these sites very separate from the code that defines the sites themselves.


I think a better approach might be to have the Puppet code which defines these sites export the monitoring resource, then in `profile::datadog_http_check` "collect" those exported resources.

[See the Puppet docs for more on exported resources](https://docs.puppet.com/puppet/4.10/lang_exported.html)

olblak: I was already concerned by that but my puppetfu is not good enough to solve it.
And I don't think that 'exported resources' will help here.
We can only call the class 'datadog_agent::integrations::http_check' one time per node
  
'profile::datadog_http_check' use the variable 'profile::datadog_http_check::instances'
to know what need to be define in /etc/dd-agent/conf.d/http_check.yaml.

Which means that each application needs to append his instance configuration to profile::datadog_http_check::instances.

A solution would be to move this variable to hiera

rtyler: @olblak I am not sure I understand how an exported resource doesn't solve this problem. `profile::datadog_http_check` doesn't necessarily _need_ to be a `class`, it could also be implemented as a "defined type".


The collection of exported resources would work very well with the `concat` model that the Datadog code uses, since the puppetmaster would collect a resource from all the nodes and then concat that file on the master.

olblak: To clarify my point, it's possible to use exported resources if we continue to generate '/etc/dd-agent/conf.d/http_check.yaml' like we do now with something like profile::datadog_check.

But now the datadog_agent module provides an easy way to configure multiple sitename within one puppet resource, like I do in profile::datadog_http_check.
In that case, I don't see how to use exported resources even if I declare profile::datadog_http_check to be definition instead of a class.

The reason why I went with the profile::datadog_http_check was because I found it clearier even if it meant splitting the code of one 'profile' in several places but I can definitely implement the first solution.

Remark: Exported resource required puppetdb which is fine for production environment  (as it's already there) but it adds complexity to vagrant environment.

olblak: In short there are three possibilities
1. I define one http_check per 'profile' (confluence,accountapp,jira,...)
2. I define everything in one place like profile::datadog_http_check
3. I use exported resources, which will require more work for the sandbox environment as each puppet node need an access to puppetdb (https://puppetdb:8001/).

Personally I prefer option 1 or 2.

PR #794, opened 2017-06-28
Add the Azure VM agents plugin to trusted.ci

rtyler:
Fixes INFRA-1066


rtyler:
The manual configuration (bleh) in production has been finished and agents are provisioning and running pipelines

PR #814, opened 2017-08-01
Return an empty JSON file for all api/json requests

rtyler:
This should effectively break a lot of the abusive clients which are not
following redirects. Notably Checkman which behaves in a spectacularly dumb
fashion with regards to redirects.

### DIFF
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -22,6 +22,7 @@ class profile::buildmaster(
 ) {
   include ::stdlib
   include ::apache
+  include apache::mod::alias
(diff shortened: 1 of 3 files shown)


dist/profile/manifests/buildmaster.pp
```diff
@@ -275,6 +276,13 @@
     require => Package['httpd'],
   }
 
+  file { "${docroot}/empty.json" :
+    ensure  => file,
+    content => '{}',
+    mode    => '0755',
```
olblak: Is there any reasons why empty.json is an executable file?
I think 0644 is more appropriated

olblak (approved):
👍

PR #819, opened 2017-08-11
[INFRA-1287] Update puppet code to handle multiple Kubernetes cluster

olblak:
This PR add multiple Kubernetes cluster support and three new puppet definitions:

1. 'profile::kubernetes::backup':
 which exports on a daily basis a Kubernetes resource from clusters to yaml file.
It's currently used to backup secret (that contain letsencrypt certificate) generated by Kube-lego in order to restore them on another cluster.
2. 'profile::kubernetes::reload':
Which delete pods based on the label 'app'.
It's used to delete&restart pods managed by daemonset && deployment.
3. 'profile::kubernetes::delete':
Which ensures that a specific resource is correctly removed from the cluster.
I only use it for testing purpose at the moment.

Regarding multiple cluster support, I generate one Kubernetes configuration files per cluster and use it to run kubectl operations on each clusters which means no differences between clusters as I run same operations everywhere.

In order to add additional clusters, we just have to update the variable 'profile::kubernetes::params::clusters' as explain in  'dist/profile/manifests/kubernetes/params.pp'


olblak:
I am still missing several tests files but I would be interested by your comment on the puppet code @rtyler  :)

dist/profile/manifests/kubernetes/backup.pp
```diff
@@ -0,0 +1,47 @@
...
+  include ::stdlib
+  include profile::kubernetes::params
+
+
+  $clusters.each | $cluster | {
+    cron { "Backup ${type}/${resource} from ${cluster[clustername]}":
+      ensure  => $ensure,
+      user    => $user,
+      name    => "Backup ${type}/${resource} from ${cluster[clustername]}",
+      command => "backup.sh ${cluster[clustername]} ${resource} ${type}",
```
rtyler: I think this should interpolate `${profile::kubernetes::params::bin}` into the `command` attribute to make sure that there are no mistakes when calling the script

rtyler (approved):
My only minor comment can be handled in a follow up as far as I am concerned

PR #823, opened 2017-08-15
Correct the username/password for secrets used in Kubernetes

rtyler:
These are no longer base64 encoded in hiera, since the `accountapp.pp` file already does that task for the password.


olblak:
I may be wrong but the SMTP user is not correct. 
According this documentation [Doc](https://sendgrid.com/docs/Classroom/Basics/Email_Infrastructure/recommended_smtp_settings.html) 
the SMTP user must be set to 'apikey' and the password must use a valid API key

rtyler:
@olblak the credentials provided work, and are already configured in our CI infrastructure. That document is referring to setting up additional SMTP API users.

olblak:
Ack

olblak (approved):

PR #827, opened 2017-08-16
Accountapp: Enable Cookie Session affinity

olblak:


olblak:
@rtyler I am not sure to understand why the latest commit is needed ( e92cece )

rtyler (dismissed):
@olblak it's not, and never should be. merging *only* ever is in the direction of `staging` :arrow_right: `production`.

olblak:
It's well what I was thinking 
@rtyler What do you suggest for this PR? 
It's already in staging (I updated staging from this PR by mistake ) so should I undo the branch merge or do we stay like this?

rtyler:
Fine, but please stop merging and smashing code around into repositories now that you have the access to.

PR #837, opened 2017-08-25
More jenkins automation!

rtyler:
Special thanks to @oleg-nenashev and @markewaite who have helped make some of these scripts.

### DIFF
--- a/dist/profile/files/buildmaster/agent-security.groovy
+++ b/dist/profile/files/buildmaster/agent-security.groovy
@@ -1,0 +1,16 @@
+#!/usr/bin/env groovy
+/*
+ * Restrict the JNLP protocols to only those which should be enabled (modern
+ * and secure ones)
+ */
+
+import jenkins.model.*
+import org.kohsuke.stapler.StaplerProxy
+import jenkins.security.s2m.AdminWhitelistRule
+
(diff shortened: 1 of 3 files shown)


dist/profile/files/buildmaster/agent-security.groovy
```diff
@@ -0,0 +1,16 @@
...
+ * Restrict the JNLP protocols to only those which should be enabled (modern
+ * and secure ones)
+ */
+
+import jenkins.model.*
+import org.kohsuke.stapler.StaplerProxy
+import jenkins.security.s2m.AdminWhitelistRule
+
+/* Restrict agent protocols to only modern and secure ones */
+Jenkins.instance.agentProtocols = ['JNLP4-connect', 'Ping']
```
oleg-nenashev, line 12: "Ping" is not required, it will be enabled by Jenkins anyway

oleg-nenashev (approved):
👍 though I highly recommend setting up Job Restrictions on the master as well. https://github.com/oleg-nenashev/demo-jenkins-config-as-code/blob/master/init_scripts/src/main/groovy/MasterComputer.groovy

Zero executors does not protect from all cases (e.g. not-so-Flyweight tasks)

PR #839, opened 2017-08-25
Add support for setting up consistent global Pipeline configurations

rtyler:
Thanks to @oleg-nenashev for figuring out how to do this. :smile:


oleg-nenashev:
I would rather do it on the Folder level. Otherwise it may be complicated to test Pipeline Libraries on the same instance.

rtyler:
@oleg-nenashev that's not what I'm deploying in ci.jenkins.io. This represents the current (manually configured) state of things.

oleg-nenashev (approved):
OK, we can always setup a new Jenkins instance for it. Or Dockerize it somehow to run Jenkins-in-Jenkins-on-Jenkins

PR #847, opened 2017-09-02
Enable deployment to non-Azure Kubernetes environments for development

orrc:
As I don't have an Azure account, I tried out running under a local [minikube-powered](https://github.com/kubernetes/minikube) Kubernetes cluster (with help from @olblak).

However, many containers won't start up if they're using an Azure-specific resource type, such as an `azureFile` volume.  This change lets us swap out parts of the config for alternative non-Azure implementations, based on a parameter.

In my `hieradata/roles/kubernetes.yaml`, I now just have to define:
```
profile::kubernetes::params::deploy_context: local
```

We tried a couple of approaches to swap out resource types during provisioning, and this ended up being the simplest.

I also took notes while getting started with this Vagrant- and minikube-powered setup, so we could add some documentation for people who might want to develop/test locally in this manner.

### DIFF
--- a/dist/profile/manifests/kubernetes/params.pp
+++ b/dist/profile/manifests/kubernetes/params.pp
@@ -20,6 +20,11 @@
 #     $trash:
 #       Directory that contains Kubernetes resources that need to be removed from Kubernetes
 #       clusters.
+#     $deploy_context:
+#       Describes the context in which this code is running, e.g. locally, or on Azure.
+#       This can be used, for example, to swap out Azure-specific storage volumes
+#       for ephemeral, local volumes during development with minikube or similar.
+#       Currently used values are: 'local', 'azure'.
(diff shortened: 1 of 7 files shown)


rtyler:
Very interesting PR @orrc, I'll try to take a look at it tomorrow or later today after I'm properly caffeinated and of the appropriate mind :smiling_imp:

orrc:
Yeah… with this, I was able to successfully launch the entire Kubernetes setup on my laptop with minikube, but actually testing the services was difficult given that minikube exposes its services on a weird port — I'm not sure yet how configurable the network/ingress stuff is with minikube.

If not, this PR might not be hugely useful, since one can presumably always test the individual services (e.g. accountapp) in isolation.  Plus those who do the most work on the infra will have Azure accounts.

PR #851, opened 2017-09-11
[INFRA-1328] Give */api/python requests the same thing as */api/json

abayer:
[INFRA-1328](https://issues.jenkins-ci.org/browse/INFRA-1328)

I *think* this is right?

### DIFF
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -364,8 +364,8 @@ ProxyPassReverse / http://localhost:8080/
         path       => "${docroot}/empty.json",
       },
       {
-        # Send all api/xml to `empty.xml` to prevent abusive clients
-        # (checkman) from receiving an invalid XML response and repeatedly attempting
+        # Send all api/xml to `empty.xml` to prevent abusive clients (like checkman over
+        # in JSON land) from receiving an invalid XML response and repeatedly attempting
         # to hammer us to get a better response.
         aliasmatch => '(.*)/api/xml(/|$)(.*)',
         path       => "${docroot}/empty.xml",
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -344,16 +344,16 @@ RewriteRule ^/cli.* https://github.com/jenkinsci-cert/SECURITY-218
 # proxy_pass_match configurations
 ProxyRequests Off
 ProxyPreserveHost On
-ProxyPassMatch (.*)/api/json(/|$)(.*)  !
+ProxyPassMatch (.*)/api/(json|python)(/|$)(.*)  !
 ProxyPass / http://localhost:8080/ nocanon
 ProxyPassReverse / http://localhost:8080/
 ",
     aliases               => [
       {
-        # Send all api/json requests to `empty.json` to prevent abusive clients
+        # Send all api/json or api/python requests to `empty.json` to prevent abusive clients
         # (checkman) from receiving an invalid JSON response and repeatedly attempting
-        # to hammer us to get a better response
-        aliasmatch => '(.*)/api/json(/|$)(.*)',
+        # to hammer us to get a better response. Works for Python API as well.
+        aliasmatch => '(.*)/api/(json|python)(/|$)(.*)',
         path       => "${docroot}/empty.json",
       },
     ],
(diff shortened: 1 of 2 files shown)


dist/profile/manifests/buildmaster.pp
```diff
@@ -344,16 +344,16 @@
 # proxy_pass_match configurations
 ProxyRequests Off
 ProxyPreserveHost On
-ProxyPassMatch (.*)/api/json(/|$)(.*)  !
+ProxyPassMatch (.*)/api/(json|python)(/|$)(.*)  !
```
jglick: `|xml`

dist/profile/manifests/buildmaster.pp
```diff
@@ -344,16 +344,16 @@
...
     aliases               => [
       {
-        # Send all api/json requests to `empty.json` to prevent abusive clients
+        # Send all api/json or api/python requests to `empty.json` to prevent abusive clients
         # (checkman) from receiving an invalid JSON response and repeatedly attempting
-        # to hammer us to get a better response
-        aliasmatch => '(.*)/api/json(/|$)(.*)',
+        # to hammer us to get a better response. Works for Python API as well.
+        aliasmatch => '(.*)/api/(json|python)(/|$)(.*)',
         path       => "${docroot}/empty.json",
```
jglick, line 364: Well that will be invalid for Python.

Anyway we should not be redirecting to an empty response! Use an HTTP error code.

spec/server/jenkins_master/jenkins_master_spec.rb
```diff
@@ -47,6 +47,11 @@
           its(:stdout) { should match '{}' }
         end
       end
+      # Same trick but this time with api/python
+      describe command("curl --verbose --insecure -A \"#{agent}\" -H 'Location: https://ci.jenkins.io/' https://127.0.0.1/api/python") do
+        its(:exit_status) { should eql 0 }
```
jglick, line 52: Crazy. Should be a 403 I think.

jglick (changes requested):

dist/profile/manifests/buildmaster.pp
abayer: It isn't! `{}` is valid JSON and, so far as I can tell, valid Python. =)

spec/server/jenkins_master/jenkins_master_spec.rb
abayer: @rtyler can correct me if I'm wrong, but I believe this approach is because doing a 403 doesn't necessarily shut up some of these shitty clients - they just keep trying.

dist/profile/manifests/buildmaster.pp
rtyler: The responses for api/json are specifically set because of abusive clients which don't respect any headers or other data. Do not attempt to apply logic here.

spec/server/jenkins_master/jenkins_master_spec.rb
rtyler: correctomundo! :frowning:

jglick: So can we not just continue sending 403s?

dist/profile/manifests/buildmaster.pp
```diff
@@ -356,6 +363,13 @@
         aliasmatch => '(.*)/api/(json|python)(/|$)(.*)',
         path       => "${docroot}/empty.json",
       },
+      {
+        # Send all api/xml to `empty.xml` to prevent abusive clients
+        # (checkman) from receiving an invalid XML response and repeatedly attempting
```
jglick: Presumably it is not checkman (whatever that is) that would be asking for XML.

dist/profile/manifests/buildmaster.pp
jglick: And where are sources for this client? Have you filed a PR to fix it?

jglick (approved):

spec/server/jenkins_master/jenkins_master_spec.rb
daniel-beck: Every response that's not 200 with fake data will just result in reattempted requests from terrible clients.

rtyler:
Since ci.jenkins.io is a bit underwater, I've run the tests locally and I think we're good to yolo this

dist/profile/manifests/buildmaster.pp
daniel-beck: @jglick https://github.com/cppforlife/checkman/issues/74 (but it requests JSON)

abayer:
@rtyler Unless I truly borked things, worst that can happen is...we make some API client sad. Which I'm fine with. =)

PR #853, opened 2017-09-13
Remove unneeded datadog http check for jira/confluence

olblak:
Datadog configuration tests fail when we generate a configuration with no concat fragment.
If a puppet manifest only contains datadog_check sets to absent, puppet generate a file which is not a valid yaml and a wrong datadog configuration.

Each time datadog package is updated, it runs configuration tests during upgrade which lead to a failure.

.Example Puppet code
---
    profile::datadog_check { 'accountapp-http-check':
      ensure  => 'absent',
      checker => 'http_check',
      source  => 'puppet:///modules/profile/accountapp/http_check.yaml',
    }   
---

.conf.d/http_check.yaml
---
  init_config:

  instances:
---

Once puppet stop generating those files, we can delete them manually

### DIFF
--- a/dist/profile/manifests/confluence.pp
+++ b/dist/profile/manifests/confluence.pp
@@ -141,12 +141,6 @@ class profile::confluence (
   profile::apachemaintenance { 'wiki.jenkins.io':
   }
 
-  profile::datadog_check { 'confluence-http-check':
-    ensure  => 'absent',
-    checker => 'http_check',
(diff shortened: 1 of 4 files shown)


rtyler (approved):

rtyler:
FWIW these tests are correctly failing

olblak:
I removed accountapp modification from this PR as I do some clean up in this PR -> PR #854

rtyler (approved):

PR #876, opened 2017-11-07
Add SessionAffinity for ingress, accountapp, repo_proxy

olblak:
Some application require that all http request are send to the same container, especially the accountapp.
-> https://kubernetes.io/docs/concepts/services-networking/service/#proxy-mode-iptables


rtyler (approved):
Huh, I thought this was already there.

PR #887, opened 2017-11-10
Revert accountapp

olblak:
As the accountapp doesn't behave correctly on the kube cluster.
Let's put it back on his previous machine while we migrate the ldap server on this cluster and do more testing.


spec/classes/profile/accountapp_spec.rb
```diff
@@ -1,24 +1,24 @@
...
-describe 'profile::accountapp' do
-  files.each do | file|
-    it { should contain_file(file).with_ensure(:absent) }
-  end
-  vhosts.each do | vhost|
-    it { should contain_apache__vhost(vhost).with_ensure(:absent)}
-  end
-  it { should contain_docker__run('account-app').with_ensure(:absent)}
-end
+#files = [
```
rtyler: I don't understand why this file wasn't just removed if it's no longer valid.

rtyler (approved):

PR #896, opened 2017-11-13
Properly grant READ permissions to anonymous and authenticated users

rtyler:
These may have been manually configured, a manual configuration which recently
was nuked. :fire: :smile:

### DIFF
--- a/dist/profile/templates/buildmaster/lockbox.groovy.erb
+++ b/dist/profile/templates/buildmaster/lockbox.groovy.erb
@@ -45,12 +45,18 @@ if (instance.securityRealm == SecurityRealm.NO_AUTHENTICATION) {
   instance.securityRealm = realm
 }
 
-println "Checking authorization strategy"
 /* Every time we run, it's worth re-defining our authorization strategy to make
  * sure that it is correct
  */
 AuthorizationStrategy strategy = new GlobalMatrixAuthorizationStrategy()
-strategy.add(Jenkins.READ, 'anonymous')
+
+[
+    Jenkins.READ,
+    Item.READ,
+].each { permission ->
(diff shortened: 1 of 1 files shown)


dist/profile/templates/buildmaster/lockbox.groovy.erb
```diff
@@ -45,12 +45,20 @@ if (instance.securityRealm == SecurityRealm.NO_AUTHENTICATION) {
...
-println "Checking authorization strategy"
 /* Every time we run, it's worth re-defining our authorization strategy to make
  * sure that it is correct
  */
 AuthorizationStrategy strategy = new GlobalMatrixAuthorizationStrategy()
-strategy.add(Jenkins.READ, 'anonymous')
+
+[
+    Jenkins.READ,
+    Item.DISCOVER,
```
daniel-beck: This doesn't do anything if you grant `Item.READ`.

dist/profile/templates/buildmaster/lockbox.groovy.erb
```diff
@@ -45,12 +45,20 @@ if (instance.securityRealm == SecurityRealm.NO_AUTHENTICATION) {
...
+    Item.READ,
+].each { permission ->
+    strategy.add(permission, 'anonymous')
+    strategy.add(permission, 'authenticated')
```
daniel-beck: This shouldn't actually be necessary – if it is, I would consider it a bug.

daniel-beck: Yep, should really not be necessary: https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/security/SidACL.java#L88...L93

PR #900, opened 2017-11-15
[Infra-1398] Use volume hostPath for repo-proxy

olblak:
The PR change the mounted volume used by repo-proxy container.
Instead of using azure files which rely on the network, we use hostPath that mount local node directory inside a container.
This means that instead of having one directory shared across containers (Azure file), we now have one caching directory per node per container.
Remark: If for some reason a container has to move on another nodes, it lose/changes his cache directory.

### DIFF
--- a/dist/profile/templates/kubernetes/resources/repo_proxy/deployment.yaml.erb
+++ b/dist/profile/templates/kubernetes/resources/repo_proxy/deployment.yaml.erb
@@ -16,30 +16,36 @@ spec:
                 type: repo-proxy
                 logtype: archive
         spec:
+            # Nginx user need write permission on /var/cache/nginx
+            initContainers:
+                - name: volume-mount-hack
+                  image: busybox
+                  command: ["sh", "-c", "chown -R 101:101 /var/cache/nginx"]
+                  volumeMounts:
+                      - name: repo-proxy
+                        mountPath: /var/cache/nginx
             containers:
                 - name: repo-proxy
                   image: jenkinsciinfra/repo-proxy:<%= @parameters['image_tag'] %>
                   imagePullPolicy: Always
                   livenessProbe:
                       httpGet:
-                          path: /api/system/ping
+                          path: /repo1-cache/org/springframework/spring-tx/maven-metadata.xml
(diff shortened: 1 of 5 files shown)


dist/profile/templates/kubernetes/resources/repo_proxy/deployment.yaml.erb
```diff
@@ -16,30 +16,36 @@ spec:
                 type: repo-proxy
                 logtype: archive
         spec:
+            # Nginx user need write permission on /var/cache/nginx
+            initContainers:
+                - name: volume-mount-hack
+                  image: busybox
+                  command: ["sh", "-c", "chown -R 101:101 /var/cache/nginx"]
```
rtyler: Is the nginx user guaranteed to always be 101 in the container? Assuming not, this could be something like `$(id -u nginx):$(id -g nginx)`

dist/profile/templates/kubernetes/resources/repo_proxy/deployment.yaml.erb
```diff
@@ -16,30 +16,36 @@ spec:
...
+                      - name: repo-proxy
+                        mountPath: /var/cache/nginx
             containers:
                 - name: repo-proxy
                   image: jenkinsciinfra/repo-proxy:<%= @parameters['image_tag'] %>
                   imagePullPolicy: Always
                   livenessProbe:
                       httpGet:
-                          path: /api/system/ping
+                          path: /releases/com/perfectomobile/jenkins/2.20.0.19/jenkins-2.20.0.19.pom
```
rtyler: Why is this checking for an artifact which doesn't matter? I think the liveness check should be for a Jenkins core version pom which is frequently used by our CI infrastructure for `@JenkinsRule` tests.

olblak: I took this one because it was a small file, but I updated the check to fetch '/repo1/org/springframework/spring-tx/maven-metadata.xml' which the most download file according artifactory

rtyler (approved):

PR #901, opened 2017-11-15
Purge the staticsite and its unnecessary wasting of space from eggplant

rtyler:


dist/profile/manifests/staticsite.pp
```diff
@@ -91,19 +89,14 @@
...
     target  => "${site_root}/archives",
     require => File["${site_root}/archives"],
   }
 
-  exec { 'chown staticsite':
-    command     => "/bin/chown -R ${deployer_user}:${deployer_group} ${site_root}",
-    refreshonly => true,
-  }
-
   apache::vhost { 'beta.jenkins-ci.org':
```
olblak, line 101: I think this vhost is not used anymore and can be remove  with ensure => 'absent'

olblak: This vhost can also be removed

olblak (dismissed):
Vhost configuration should be removed (especially if docroots are deleted) with ensure => 'absent', remove letsencrypt configuration and delete old letsencrypt certificates.
Remark: jenkins-ci.org zone should be updated to send traffic to the kube instead of this eggplant like it's done for jenkins.io but his should be done in another PR

rtyler:
@olblak the letsencrypt stuff doesn't support explicitly being marked as `ensure => absent` 🙁

olblak:
Required [#PR907](https://github.com/jenkins-infra/jenkins-infra/pull/907) for jenkins-ci.org

PR #903, opened 2017-11-16
Introduce an intentional failure

rtyler:
This is just to test some Azure function silliness


rtyler:
I'm done testing, let's see how this works for a few days 😄

PR #918, opened 2017-11-21
Domain name verification for certificate renewal

kohsuke:
Follow up to 768b7bdeb8b409ffa99a9b2408487d92c3a03ac7

I think TXT record needs to target the domain name itself, not after
resolving CNAME.


kohsuke:
Ah, and now I know why Olivier did what he did originally...

PR #922, opened 2017-11-22
Ensure that Class[Letsencrypt] is in the catalogue for the account app

rtyler:
Turns out this was previously being pulled in by profile::staticsite and I
could not get the catalogue compilation error below to reproduce in the testing
environment unless I explicitly tested the expectation in rspec-puppet

    Error: Could not retrieve catalog from remote server: Error 500 on SERVER: {"message":"Server Error: Invalid relationship: Exec[letsencrypt certonly accounts.jenkins.io] { require => Class[Letsencrypt] }, because Class[Letsencrypt] doesn't seem to be in the catalog","issue_kind":"RUNTIME_ERROR","stacktrace":["Warning: The 'stacktrace' property is deprecated and will be removed in a future version of Puppet. For security reasons, stacktraces are not returned with Puppet HTTP Error responses."]}
    Warning: Not using cache on failed catalog
    Error: Could not retrieve catalog; skipping run

### DIFF
--- a/dist/profile/manifests/accountapp.pp
+++ b/dist/profile/manifests/accountapp.pp
@@ -30,6 +30,7 @@ class profile::accountapp(
 ) {
   include profile::docker
   include profile::apachemisc
+  include profile::letsencrypt
 
   $docroot = '/var/www/html'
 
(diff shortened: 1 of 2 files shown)


rtyler:
@olblak I would rather wait for your review, but I'm going to fast-track this so eggplant can pass the puppet runs before Thanksgiving 🦃

PR #929, opened 2017-11-29
[INFRA-1380] Ingress pod to daemonset

olblak:
In order to preserve the source IP, loadbalancer service need to have the annotation  'service.beta.kubernetes.io/external-traffic: OnlyLocal'.
which forces nodes without Service endpoints to remove themselves from the list of nodes eligible for loadbalanced traffic by deliberately failing health checks. 
In order to keep all nodes eligible for loadbalanced traffic, each node must run one nginx endpoint.

annotation: 'onlyLocal' is already used since months but at the moment we define the number of container needed to 3 instead of using the number of nodes.
This has already leaded to have all nginx endpoint on the same node or nodes without nginx endpoints

[Doc](https://kubernetes.io/docs/tutorials/services/source-ip/)

### DIFF
--- a/dist/profile/manifests/kubernetes/resources/nginx.pp
+++ b/dist/profile/manifests/kubernetes/resources/nginx.pp
@@ -49,9 +49,9 @@ class profile::kubernetes::resources::nginx (
       context  => $context,
       resource => 'nginx/default-service.yaml'
     }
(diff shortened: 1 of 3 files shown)


rtyler (changes requested):
The commit doesn't explain at all why this is important

rtyler (approved):

PR #946, opened 2018-01-03
Pin the azure-cli repository only to amd64

rtyler:
Azure's i386 file seems to be junk:
    W: Failed to fetch https://apt-mo.trafficmanager.net/repos/azure-cli/dists/wheezy/InRelease  Unable to find expected entry 'main/binary-i386/Packages' in Release file (Wrong sources.list entry or malformed file)

Pinning the architecture seems to work well enough


rtyler:
fwiw I validated this with a real instance, it does make the error go away and the azure-cli installable.

PR #952, opened 2018-01-08
[INFRA-1409] - Assign a fixed public IP to kubernetes cluster

olblak:
### Cluster
This PR allow us to set a different fixed public IP per cluster instead of using random ephemeral IP. 
It's needed for Ingress and LDAP (coming).

Some re-factoring was needed as explain below.
Up to now, I made the assumption that all clusters were configured
the same way. So puppet was responsible to generate the same yaml files
for all clusters.

But even if we try to keep the same configuration for all clusters identically,
they are still some benefits to have customized clusters.

~~~~
# Current Situation
resources
├─nginx
├─kubelego
└─...
~~~~

~~~~
# New situation
resources
├─cluster1
│ ├─nginx
│ ├─kubelego
│ └─ ...
├─cluster2
│ ├─nginx
│ ├─kubelego
│ └─ ...
└── ...
~~~~

### Config

 Merge all kubectl configuration into one file, the default one '~/.kube/config'.
 This introduces two changes:

 1. We use --context 'clustername' instead of --kubeconfig .kube/clustername.conf to interacte with a specific cluster.
 2. By default, kubectl use the first cluster defined in the puppet variable 'profile::kubernetes::params::clusters:''

### Set nginx public IP

If defined, use a fixed public IP


dist/profile/manifests/kubernetes/apply.pp
```diff
@@ -24,46 +24,56 @@
 #
 define profile::kubernetes::apply (
   String $resource = $title,
+  String $context = '',
```
rtyler: There is no documentation for what `context` is supposed to be for using this defined type. Please update the Puppet  Doc above.

dist/profile/manifests/kubernetes/backup.pp
```diff
@@ -26,6 +26,7 @@
 #     }
 #
 define profile::kubernetes::backup(
+  String $context = '',
```
rtyler: Please add some docs for `context` here as well

dist/profile/manifests/kubernetes/backup.pp
```diff
@@ -36,15 +37,16 @@
   include ::stdlib
```
rtyler, line 39: Shouldn't the `$clusters` parameter be deleted?

dist/profile/manifests/kubernetes/apply.pp
```diff
@@ -24,46 +24,56 @@
...
+  String $user = $profile::kubernetes::params::user,
+  String $kubeconfig = $profile::kubernetes::params::kubeconfig,
   Hash $parameters = {},
   $clusters = $profile::kubernetes::params::clusters
```
rtyler: Shouldn't the `$clusters` parameter be deleted?

dist/profile/manifests/kubernetes/delete.pp
```diff
@@ -19,8 +19,10 @@
...
 #
 define profile::kubernetes::delete (
+  String $context = '',
   String $resource = $title,
-  $clusters = $profile::kubernetes::params::clusters
+  String $user = $profile::kubernetes::params::user,
+  String $kubeconfig = $profile::kubernetes::params::kubeconfig,
 ){
   include ::stdlib
   include profile::kubernetes::params
```
rtyler: Shouldn't there be a similar `fail()` statement in this resource when `$context` == ''

dist/profile/manifests/kubernetes/delete.pp
```diff
@@ -29,24 +31,34 @@
...
 
-  $clusters.each | $cluster | {
-    # Only run kubectl delete if the resources is deployed.
-    exec { "Remove ${resource} on ${cluster[clustername]}":
-      command     => "kubectl delete --grace-period=60 --ignore-not-found=true -f ${profile::kubernetes::params::trash}/${dirname}.${basename}",
-      path        => [$profile::kubernetes::params::bin,$::path],
-      environment => ["KUBECONFIG=${profile::kubernetes::params::home}/.kube/${cluster[clustername]}.conf"] ,
-      onlyif      => "test \"$(kubectl apply --dry-run=true -f ${profile::kubernetes::params::trash}/${dirname}.${basename} | grep configured)\""
-    }
+$delete_args = "\
```
rtyler: I'm surprised this variable declaration and `$apply_args` are passing Puppet Lint. Please indent them to the appropriate level within the resource.

spec/classes/profile/kubernetes/accountapp_spec.rb
```diff
@@ -1,73 +1,86 @@
 require 'spec_helper'
```
rtyler, line 1: As far as I can tell, noone of the RSpec changes appropriately test the behavior of multiple `$clusters` defined. If I missed it, please point it out to me, otherwise I think some test coverage for that scenario should be added.

rtyler (changes requested):
I'm supportive of this change, I think there are some loose ends to tidy up before it's merged however.

spec/classes/profile/kubernetes/accountapp_spec.rb
olblak: You are right. I removed some tests because it was not working as expected. I still have some test to write for that

rtyler (approved):

PR #953, opened 2018-01-11
Update prodbean credentials

olblak:


rtyler:
@olblak I don't have the time to decrypt these values, so I trust that you're submitting them correctly 😄

olblak:
@rtyler If not we'll know it very quickly :D

PR #990, opened 2018-03-21
[INFRA-1530] Only run kubectl apply when resource file changed

olblak:


rtyler (changes requested):
Why bother commenting these items out? Why not just remove them entirely? They'll always be there in the Git history :smile:

PR #994, opened 2018-04-05
Add my new machine's SSH public key

rtyler:
Never hurts to re-key with a new machine :smile:


olblak:
Is the old key still used?

rtyler:
@olblak yes

PR #996, opened 2018-04-05
Add DCV record for the new ldap.jenkins.io certificate

rtyler:
See  INFRA-1545


rtyler:
@olblak Would you mind sanity checking my syntax here before I merge and deploy this?

rtyler:
ETIMEOUT, let's give it a try!

PR #1006, opened 2018-04-25
[INFRA-1571] Permit API requests when there is an Authorization header

jglick:
Amending #814 to behave normally when there is an `Authorization` header, used by requests with API tokens. (If there is an _invalid_ header, Jenkins will just spit back a 403.)

Needed for https://github.com/jenkins-infra/community-functions/pull/10.

I am not able to test this directly, but I tested [this model](https://github.com/jglick/jenkins-demo-reverse-proxy/compare/api) (using the same Apache 2.4.7 as ci.jenkins.io runs IIUC) and it worked.


rtyler (approved):
This looks good to me, I'm going to do some more checks on this tomorrow (Wednesday) and then merge/deploy it.

olblak:
:+1:

PR #1061, opened 2018-07-24
Provision the Evergreen service layer

rtyler:
This is roughly following other examples for provisioning the evergreen
container

Fixes JENKINS-50985

@olblak I'm not entirely sure how to test this, but tossing this up here for now


olblak:
I am happy to see this service running :) 
It would be nice to namespace this application like [lego](https://github.com/jenkins-infra/jenkins-infra/tree/staging/dist/profile/templates/kubernetes/resources/lego), I know that I almost never do that (and I should), but in this case the main advantage will be to provide a read only access to a small group of users to that namespace with RBAC.
RBAC is not supported with the current version but I plan to migrate the cluster to AKS which support version upgrade

dist/profile/manifests/kubernetes/resources/evergreen.pp
```diff
@@ -0,0 +1,66 @@
...
+      ensure => 'directory',
+      owner  => $profile::kubernetes::params::user,
+    }
+
+    profile::kubernetes::apply { "evergreen/service.yaml on ${context}":
+      context  => $context,
+      resource => 'evergreen/service.yaml'
+    }
+
+    profile::kubernetes::apply { "evergreen/secret.yaml on ${context}":
```
olblak, line 39: I don't see any reference to a secret in the deployment so It sounds more like a wrong copy/paste

olblak: Sorry so you need a secret but you don't provide the good parameter

olblak: parameters should be:
```
parameters => {
  'postgres_url' => $postgres_url
}
```

dist/profile/templates/kubernetes/resources/evergreen/service.yaml.erb
```diff
@@ -0,0 +1,17 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: evergreen
+  labels:
+    app: evergreen
+    type: evergreen
+spec:
+  sessionAffinity: ClientIP
```
olblak: do you need a session affinity?
The accountapp is the only application that need basically it ensure that a client wlll always be routed to the same pod

dist/profile/templates/kubernetes/resources/evergreen/secret.yaml.erb
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret 
+metadata: 
+    name: accountapp
```
olblak: I think you forgot this resource :P 
all current data aren't needed and you should have the postgres_key

dist/profile/templates/kubernetes/resources/evergreen/deployment.yaml.erb
```diff
@@ -0,0 +1,37 @@
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: evergreen
+spec:
+  replicas: 1
```
olblak: Is it a stateless or statefull application? 
If it's a stateless app then I would increase the replicas to at least 3 in order to bring redundancy and the possibility to do  AB deployment.

dist/profile/manifests/kubernetes/resources/evergreen.pp
```diff
@@ -0,0 +1,66 @@
...
+#
+#   This class deploys the Jenkins Essentials backend service layer, a
+#   container which communicates with a provisioned Azure PostgreSQL database
+#
+class profile::kubernetes::resources::evergreen (
+    Array $clusters = $profile::kubernetes::params::clusters,
+    Array $domain_alias = [],
+    String $image_tag = 'latest',
+    String $domain_name = 'evergreen.jenkins.io',
+    String $postgres_url = '',
```
olblak: Where is this postgres database? On azure?

dist/profile/manifests/kubernetes/resources/evergreen.pp
```diff
@@ -0,0 +1,66 @@
...
+      parameters => {
+        'url'     => $domain_name,
+        'aliases' => $domain_alias
+      },
+      resource   =>  'evergreen/ingress-tls.yaml'
+    }
+
+    profile::kubernetes::apply { "evergreen/deployment.yaml on ${context}":
+      context    => $context,
+      parameters => {
```
olblak, line 58: you are missing ```'image_tag'             => $image_tag``` in your parameter otherwise your deployment template will not not which tag to use

olblak:
I should document the way I test this kind of PRs :\

dist/profile/manifests/kubernetes/resources/evergreen.pp
```diff
@@ -0,0 +1,66 @@
+#   Class: profile::kubernetes::resources::evergreen
+#
+#   This class deploys the Jenkins Essentials backend service layer, a
+#   container which communicates with a provisioned Azure PostgreSQL database
+#
+class profile::kubernetes::resources::evergreen (
+    Array $clusters = $profile::kubernetes::params::clusters,
+    Array $domain_alias = [],
+    String $image_tag = 'latest',
```
olblak, line 9: If you plan to always us 'latest' tag, you'll need a way to tell kubernetes to restart running pods.
A trick would be  to add an information to the deployment resource that change every weeks/days/months
Something like 
metadata:
  date:  $(date "+%V")

PR #1073, opened 2018-08-06
[INFRA-1730] Define a different fallblack mirror for /plugins

olblak:
This PR define a different fallback mirror than archives.jenkins-ci.org for mirrors.jenkins.io/plugins
I can't test this PR but I based my configuration on this [documentation](http://svn.mirrorbrain.org/viewvc/mirrorbrain/trunk/mod_mirrorbrain/mod_mirrorbrain.conf?view=markup). 
Search for 'MirrorBrainFallback'

### DIFF
--- a/dist/profile/manifests/mirrorbrain.pp
+++ b/dist/profile/manifests/mirrorbrain.pp
@@ -401,7 +401,7 @@ date \"+%s\" > /srv/releases/jenkins/TIME
 
             # we serve most files from mirrors, but as a fallback,
             # this slow server has everything.
-            MirrorBrainFallback na us https://updates.azure.jenkins.io/
+            MirrorBrainFallback na us http://updates.azure.jenkins.io/
 
             # Do not redirect for files smaller than 4096 bytes
             MirrorBrainMinSize 4096
--- a/dist/profile/manifests/mirrorbrain.pp
+++ b/dist/profile/manifests/mirrorbrain.pp
@@ -390,6 +390,30 @@ date \"+%s\" > /srv/releases/jenkins/TIME
     ],
     directories       => [
       {
+        path            => "${docroot}/plugins",
+        options         => 'FollowSymLinks Indexes',
+        allow_override  => ['All'],
+        custom_fragment => '
+            MirrorBrainEngine On
+            MirrorBrainDebug Off
+            FormGET On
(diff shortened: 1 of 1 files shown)


dist/profile/manifests/mirrorbrain.pp
```diff
@@ -389,6 +389,30 @@
...
+        allow_override  => ['All'],
+        custom_fragment => '
+            MirrorBrainEngine On
+            MirrorBrainDebug Off
+            FormGET On
+            MirrorBrainHandleHEADRequestLocally Off
+
+            # we serve most files from mirrors, but as a fallback,
+            # this slow server has everything.
+            MirrorBrainFallback na us https://updates.azure.jenkins.io/
```
olblak: I am not sure if I should redirect to http or https

daniel-beck: No. Jenkins cannot handle protocol upgrades. Pointless anyway, since the htaccess file only has http URLs.

rtyler (approved):

dist/profile/manifests/mirrorbrain.pp
rtyler: Switching to HTTP is probably best here

dist/profile/manifests/mirrorbrain.pp
```diff
@@ -389,6 +389,30 @@
...
+        options         => 'FollowSymLinks Indexes',
+        allow_override  => ['All'],
+        custom_fragment => '
+            MirrorBrainEngine On
+            MirrorBrainDebug Off
+            FormGET On
+            MirrorBrainHandleHEADRequestLocally Off
+
+            # we serve most files from mirrors, but as a fallback,
+            # this slow server has everything.
```
daniel-beck, line 403: Does not apply here.

dist/profile/manifests/mirrorbrain.pp
```diff
@@ -389,6 +389,30 @@
...
+            MirrorBrainFallback na us http://updates.azure.jenkins.io/
+
+            # Do not redirect for files smaller than 4096 bytes
+            MirrorBrainMinSize 4096
```
daniel-beck, line 407: Why? I mean, unlikely any plugin is less than 4k, but…

PR #1078, opened 2018-08-08
[INFRA-1744] Remove all apache services from eggplant (Part 2)

olblak:
WARNING: This PR must be deployed after https://github.com/jenkins-infra/jenkins-infra/pull/1077 otherwise accounts.jenkins-ci.org will not work until the dns is totally switched on k8s.

This PR remove following unused apache services from eggplant:
* javadoc
* accounts.
* maven
* jenkins-ci.org
* stats

### DIFF
--- a/dist/profile/manifests/accountapp.pp
+++ b/dist/profile/manifests/accountapp.pp
@@ -39,85 +39,24 @@ class profile::accountapp(
   # I put a very bad permission on /var/log/accountapp in order to avoid permission issue.
   #
   file { '/var/log/accountapp':
-    ensure => 'directory',
-    mode   => '0777',
-    group  => 'docker'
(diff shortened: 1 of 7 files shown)


rtyler:
Admirable work, but why bother? Why don't we just remove the references, and ask OSUOSL to just delete the VM?

olblak:
Indeed the first question would have been "do we want to keep that machine?" :D
Anyway it was interesting to verify that nothing important was left on this machine and as soon as accounts.jenkins-ci.org is correctly redirected to accounts.jenkins.io. we can decommissioned  eggplant

rtyler:
Closing this. It can be re-opened in the future if necessary. _OR_ we could just de-provision eggplant entirely :smile_cat:

PR #1088, opened 2018-08-17
Upgrade evergreen to the latest at the end of the week

rtyler:
Now with 50% more home page


rtyler:
If Jenkins isn't going to chime in with an opinion, I'm just gonna merge this! :angry:

PR #1123, opened 2018-09-20
Add the uplink resources for running uplink in kubernetes

rtyler:
@olblak needs to provide the DB_CONNECTION_STRING variable and then we should be
good to go here


dist/profile/manifests/kubernetes/resources/uplink.pp
```diff
@@ -0,0 +1,79 @@
+#   Class: profile::kubernetes::resources::uplink
+#
+#   This class deploys the Uplink app which is a simple  container
+#   container which communicates with a provisioned Azure PostgreSQL database
+#
+class profile::kubernetes::resources::uplink (
+    Array $clusters = $profile::kubernetes::params::clusters,
+    Array $domain_alias = [],
+    String $image_tag = 'latest',
+    String $domain_name = 'uplink.jenkins.io',
```
daniel-beck, line 10: I was expecting telemetry.jenkins.io. Is this deliberate?

dist/profile/manifests/kubernetes/resources/uplink.pp
```diff
@@ -0,0 +1,79 @@
+#   Class: profile::kubernetes::resources::uplink
+#
+#   This class deploys the Uplink app which is a simple  container
+#   container which communicates with a provisioned Azure PostgreSQL database
```
daniel-beck: container container

dist/profile/manifests/kubernetes/resources/uplink.pp
rtyler: I would rather the domain name reflect the application that's running there. Since this isn't a general purpose telemetry service, I don't see any reason to call it as such.
:man_shrugging:

daniel-beck: Hence `jira.jenkins.io` and `confluence.jenkins.io` 😉

olblak (approved):
Looks fine to me, once the secret from this PR is merged https://github.com/rtyler/jenkins-infra/pull/2/files

PR #1157, opened 2018-10-10
Bump to latest evergreen backend '201810102027'

batmat:
Particularly to pick-up, at least:
* https://github.com/jenkins-infra/evergreen/pull/314
* https://github.com/jenkins-infra/evergreen/pull/313

@rtyler @olblak cc @mandie722

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -348,7 +348,7 @@ profile::datadog_pluginsite_check::sites:
 
 profile::kubernetes::resources::fluentd::image_tag: '0.7.2'
 
-profile::kubernetes::resources::evergreen::image_tag: '201809261928'
+profile::kubernetes::resources::evergreen::image_tag: '201810102027'
 
 limits:
   '*':


rtyler:
We do not yet have migrations running automatically, does this need to have migrations run before or after it?

batmat:
I do not think it matters, as it's not like there's data migration or something and code using the new schema. It will just allow now, for instance, that a just rollbacked instance can call `reportVersions` without receiving an error.

Does this make sense, or am I misunderstanding your question?

olblak:
It looks good to me but more globally as evergreen stores states in a database, it's important to know if we can easily rollback to a previous version and if some manual operations are needed before or after the upgrade.

olblak:
@batmat I can push the green button when you are ready

batmat:
```
evergreen_prod=> \d versions;
                                         Table "public.versions"
        Column         |           Type           |                       Modifiers
-----------------------+--------------------------+-------------------------------------------------------
 id                    | integer                  | not null default nextval('versions_id_seq'::regclass)
 uuid                  | character varying(255)   | not null
 manifest              | json                     |
 manifestSchemaVersion | integer                  |
 createdAt             | timestamp with time zone | not null
 updatedAt             | timestamp with time zone | not null
 checksum              | character varying(255)   | not null
Indexes:
    "versions_pkey" PRIMARY KEY, btree (id)
    "uuid_checksum_uniq" UNIQUE CONSTRAINT, btree (uuid, checksum)
```

@rtyler AFAICT, the constraint is still there, which makes sense given the discussion above. 
How can wipe it out? Should we run the migrations container? Or delete it manually :scream: ?

In any case, I've seen errors, fortunately not blocking yet, here and there when rolling back for instance, so it would be very good to have this fixed.

Thanks!

batmat:
To clarify `"uuid_checksum_uniq" UNIQUE CONSTRAINT, btree (uuid, checksum)` should not be there anymore.

PR #1161, opened 2018-10-15
[INFRA-1810] Deploy jenkinsadmin ircbot on kubernetes cluster

olblak:
This PR deploy the jenkins admin ircbot on the kubernetes cluster.
It also stop:
* jenkinsadmin docker container from OKRA
* Stop bind service

### DIFF
--- a/dist/profile/manifests/jenkinsadmin.pp
+++ b/dist/profile/manifests/jenkinsadmin.pp
@@ -1,6 +1,14 @@
 #
 # IRC bot that runs most project related tasks
 # containerized in https://github.com/jenkins-infra/ircbot
+
+#############################################################
+##                                                         ##
+##  THIS CLASS IS DEPRECATED IN FAVOR OF                   ##
+##  profile::kubernetes::resources::chatbot_jenkinsadminl  ##
+##                                                         ##
+#############################################################
+
 class profile::jenkinsadmin (
   # Parameters supplied by Hiera
   $github_login,
@@ -24,6 +32,7 @@ class profile::jenkinsadmin (
     # to the invocation of the Java command, since the IRC bot .jar file
     # requires:
     #    java -jar /home/ircbot/*.jar $NICKPASSWORD
+    ensure   => 'absent',
     command  => $nick_password,
     volumes  => ['/home/ircbot/.github:/home/ircbot/.github',
                 '/home/ircbot/.jenkins-ci.org:/home/ircbot/.jenkins-ci.org',
--- a/dist/profile/manifests/jenkinsadmin.pp
+++ b/dist/profile/manifests/jenkinsadmin.pp
@@ -5,7 +5,7 @@
 #############################################################
 ##                                                         ##
 ##  THIS CLASS IS DEPRECATED IN FAVOR OF                   ##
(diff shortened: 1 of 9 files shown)


dist/profile/manifests/jenkinsadmin.pp
```diff
@@ -1,6 +1,14 @@
 #
 # IRC bot that runs most project related tasks
 # containerized in https://github.com/jenkins-infra/ircbot
+
+#############################################################
+##                                                         ##
+##  THIS CLASS IS DEPRECATED IN FAVOR OF                   ##
+##  profile::kubernetes::resources::chatbot_jenkinsadminl  ##
```
rtyler: not important, but a type in the resourec name "jenkinsadminl"

dist/profile/manifests/kubernetes/resources/chatbot_jenkinsadmin.pp
```diff
@@ -0,0 +1,63 @@
...
+        'jira'   => base64('encode', $jira_credentials , 'strict'),
+      },
+      resource   => 'chatbot_jenkinsadmin/secret.yaml'
+    }
+
+    profile::kubernetes::apply { "chatbot_jenkinsadmin/deployment.yaml on ${context}":
+      context    => $context,
+      parameters => {
+        'image_tag'     =>  $image_tag,
+        'nick_password' => $nick_password
```
rtyler: Is there a reason this isn't being set as a secret too?

dist/profile/templates/kubernetes/resources/chatbot_jenkinsadmin/secret.yaml.erb
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: jenkinsadmin
+  namespace: chatbot
+data:
+  .github: |
```
rtyler: There must be some YAML feature being used here that I'm not aware of, why does the key have a leading `.`?

dist/profile/manifests/jenkinsadmin.pp
olblak: :trumpet:

dist/profile/templates/kubernetes/resources/chatbot_jenkinsadmin/secret.yaml.erb
olblak: It's not a YAML feature but a kubernetes specificity, a secret can either be mounted as a environment variable or mounted as a volume into a pod. If it's mounted as volume, by default it mounts all secret information into a mounted directory where each secret key represent a filename and the file contents is defined by the value. In this case .github is mounted in /home/chatbot/.github. But we can still decided to only mount several keys or change the destination filename, which is what I will do as I am moving the nick_password to that secret and I don't need to mount it as file

dist/profile/manifests/kubernetes/resources/chatbot_jenkinsadmin.pp
olblak: I'll update it

rtyler (approved):

PR #1177, opened 2018-10-23
[INFRA-1769] Add ssh gateway for evergreen contributors

olblak:
This service deploys a ssh gateway with posgresql-client installed and pgpass set to the evergreen  readonly database.
User access must be configured on this [docker image ](https://github.com/jenkins-infra/sshd)  

- [x] Configure READ only database access
- [x] Set docker image tag

Following commands must be run on the evergreen database (username and password will be changed)
```
# This will prevent default users from creating tables
REVOKE CREATE ON SCHEMA public FROM public;
# Create user
CREATE ROLE readonly_user WITH LOGIN PASSWORD 'password';
GRANT CONNECT ON DATABASE db_name TO readonly_user;
GRANT SELECT ON ALL TABLES IN SCHEMA public TO readonly_user;
ALTER DEFAULT PRIVILEGES IN SCHEMA public GRANT SELECT ON TABLES TO readonly_user;
```


dist/profile/templates/kubernetes/resources/evergreen_gateway/deployment.yaml.erb
```diff
@@ -0,0 +1,37 @@
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: gateway
+  namespace: evergreen
+spec:
+  replicas: 3
```
rtyler, line 9: It's not clear to me why this Deployment requires _3_ replicas? Please add some commentary to the YAML explaining a bit more please.

dist/profile/templates/kubernetes/resources/evergreen_gateway/deployment.yaml.erb
```diff
@@ -0,0 +1,37 @@
...
+  template:
+    metadata:
+      labels:
+        app: gateway
+    spec:
+      containers:
+        - name: gateway
+          image: jenkinsciinfra/sshd:evergreen-<%= @parameters['image_tag'] %>
```
rtyler: I believe this should be `jenkinsciinfra/sshd:<%= @parameters['image_tag'] %>`

Based on how `$image_tag` is passed in currently, this will result in the image tag being: `jenkinsciinfra/sshd:evergreen-evergreen-d69c65`

(redundant `evergreen` components)

dist/profile/manifests/kubernetes/resources/evergreen_gateway.pp
```diff
@@ -0,0 +1,84 @@
...
+#     $db_name:
+#       Define evergreen database name
+#
+#     $db_pass:
+#       Define evergreen database password
+#
+
+class profile::kubernetes::resources::evergreen_gateway (
+  Array $clusters = $profile::kubernetes::params::clusters,
+  String $image_tag = 'evergreen-d69c65',
```
rtyler: This image doesn't exist https://hub.docker.com/r/jenkinsciinfra/sshd/tags/

dist/profile/templates/kubernetes/resources/evergreen_gateway/configmap.yaml.erb
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
rtyler: As far as I understand, these credentials are being inserted and stored unencrypted in Kubernetes. Shouldn't values like `@parameters['db_pass']` be stored in a `Secret`?

dist/profile/templates/kubernetes/resources/evergreen_gateway/deployment.yaml.erb
olblak: Thanks nice catch

dist/profile/templates/kubernetes/resources/evergreen_gateway/configmap.yaml.erb
olblak: True, It was a shortcut that I should have cleaned earlier

dist/profile/templates/kubernetes/resources/evergreen_gateway/deployment.yaml.erb
olblak: It's only for redundancy purpose.

olblak: It doesn't really matter on which gateway we arrive since they don't store any state.

PR #1179, opened 2018-10-29
Load $NICK_PASSWORD env variable before starting the bot

olblak:


dist/profile/templates/kubernetes/resources/chatbot_jenkinsadmin/deployment.yaml.erb
```diff
@@ -17,7 +17,9 @@ spec:
                 - name: jenkinsadmin
                   image: jenkinsciinfra/ircbot:<%= @parameters['image_tag'] %>
                   imagePullPolicy: IfNotPresent
-                  args: ["$NICK_PASSWORD"]
+                  # A shell is required to load $NICK_PASSWORD env variable
+                  command: ["/bin/sh", "-c"]
```
rtyler, line 21: It surprises me that just calling `java` as the `command` doesn't process the environment variable properly. That is contrary to my understanding of how this all works in Kubernetes

olblak: It's not a kubernetes feature but how docker handle env variables in entrypoints. They are two different entrypoint syntax, exec or shell and environment are only loaded with the shell form 

```
FROM alpine:latest
# Exec form
ENTRYPOINT ["ls","$HOME"] # no shell so we can't resolve $HOME
ENTRYPOINT ["/bin/sh","-c","ls","$HOME"] 
# Shell form
ENTRYPOINT echo $HOME # equivalent to /bin/sh -c "echo $HOME"
```

rtyler: Ew :smiling_imp:

rtyler (approved):

PR #1196, opened 2018-12-05
Use jenkins/jenkins Docker image

daniel-beck:
Manually applied to `ci.jenkins.io`, and the agent is currently, deliberately, disabled.

### DIFF
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -67,7 +67,7 @@ class profile::buildmaster(
   }
 
   docker::run { 'jenkins':
-    image            => 'jenkinsci/jenkins:lts-alpine',
+    image            => 'jenkins/jenkins:lts-alpine',
     # This is a "clever" hack to force the init script to pass the numeric UID
     # through on `docker run`. Since passing the string 'jenkins' doesn't
     # actually map the UIDs properly. Using the extra_parameters option because


batmat (approved):

dist/profile/manifests/buildmaster.pp
```diff
@@ -67,7 +67,7 @@
   }
 
   docker::run { 'jenkins':
-    image            => 'jenkinsci/jenkins:lts-alpine',
+    image            => 'jenkins/jenkins:lts-alpine',
```
batmat, line 70: Out of scope for this, but this should be using a specific tag IMO, not a moving target.

olblak (approved):

dist/profile/manifests/buildmaster.pp
daniel-beck: Disagree, I'm not going to push changes through this repo whenever we're releasing a new version.

batmat: I totally understand your point and it looks like I'm trying to add more work on your back. That's obviously not my goal here.

That's possibly an automation issue or so, but still what I'm saying we shouldn't be relying on a moving tag for configuration-as-code infrastructure, this means that upgrades could happen automatically when we didn't plan it. 

I'm not sure what the exact behaviour is with Puppet right now, I _guess_ that we currently were doing something like `docker rmi jenkinsci/jenkins` and then apply changes again to force pulling/updating to the new image?

rtyler: The `lts-alpine` tag was explicitly chosen. Every time the service restarts it ensures that it is at the latest version of the tag which is 100% the correct behavior here. We must have the latest LTS as soon as it is available.

PR #1201, opened 2018-12-10
Bump libssl version

olblak:


rtyler (commented):
There's a test that's going to fail now too. I (almost) always forget it

olblak:
I just realize that I didn't push the last commit to update the spec test -_- thanks @rtyler for doing that

PR #1231, opened 2019-02-03
Enable HTTP/2 across all Apache hosts

rtyler:
Verified manually with Vagrant :clap:

Fixes INFRA-789


olblak (commented):
I didn't test it but it seems very good

PR #1236, opened 2019-02-04
[INFRA-1985]Replace azure upload script for releases

olblak:
This PR replace [release-blob-sync](https://github.com/jenkins-infra/azure/blob/master/scripts/release-blob-sync) by another one that rely on blobxfer and azure-cli  which is more robust 

The output looks like this (in dry-run) and allow to skip upload based on different algorithms (md5 or  last modified )
 
`2019-02-04 19:14:53,877 INFO - [DRY RUN] skipping: /srv/releases/jenkins/windows/jenkins-2.75.zip -> windows/jenkins-2.75.zip
`

Remark this PR as not been tested yet

### DIFF
--- a/dist/profile/files/mirrorbrain/sync.sh
+++ b/dist/profile/files/mirrorbrain/sync.sh
@@ -46,7 +46,7 @@ popd
 
 echo ">> Delivering bits to fallback"
 /srv/releases/populate-archives.sh
-/srv/releases/azure-sync.sh
+/srv/releases/batch-upload.bash
 
 echo ">> Updating the latest symlink for weekly"
 /srv/releases/update-latest-symlink.sh
--- a/dist/profile/files/mirrorbrain/batch-upload.bash
+++ b/dist/profile/files/mirrorbrain/batch-upload.bash
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+source /srv/releases/.azure-storage-env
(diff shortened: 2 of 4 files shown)


rtyler (approved):

dist/profile/files/mirrorbrain/sync.sh
```diff
@@ -46,7 +46,7 @@ popd
 
 echo ">> Delivering bits to fallback"
 /srv/releases/populate-archives.sh
-/srv/releases/azure-sync.sh
+/srv/releases/batch-upload.bash
```
rtyler, line 49:
```suggestion
/srv/releases/batch-upload.bash || true
```

The `batch-upload.sh` script in its current form returns a non-zero exit, not sure why, but this causes the script to stop executing immediately and fail to propagate changes to the mirrors

olblak: i am not a big fan of || true, it makes it even more harder to detect when something get wrong, I prefer to use set t-e and exit 0 if nothing get wrong

olblak:
It doesn't make sense to merge this PR since [this](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/role/manifests/mirrorbrain.pp) puppet role is totally broken and should fix first

PR #1242, opened 2019-02-11
[INFRA-2007] Add ssh monitored endpoints

olblak:
[INFRA-2007](https://issues.jenkins-ci.org/browse/INFRA-2007) issues highlights that we still have endpoints that we don't monitor


rtyler (approved):
Having added a couple of these myself, I completely forgot about this monitoring we had in place :frowning:

PR #1252, opened 2019-02-15
Install python-certbot-apache for letsencrypt plugin

olblak:
This package is needed to use apache plugin


rtyler (approved):
This looks fine to me, I don't think it should conflict with the other packaged tool from the letsencrypt module

PR #1255, opened 2019-03-02
Remove the puppet-mirrorbrain module in favor of simply inlining resources

rtyler:
We don't really need to split our use of mirrorbrain into a separate module,
it's not worth the overhead :)

This pull request hasn't been properly tested outside of puppet-rspec since Vagrant is acting a fool on my machine :slightly_frowning_face:


olblak:
Let me test this in my *working* vagrant environment before breaking this machine again

olblak:
As expected, this role requires more works in order to be conifgured by puppet6 
* puppetlabs/postgresql from '4.7.1' to  '5.12.1'
* Package[libapache-mid-mirrorbrain] not found
* And probably other things
I can't do more investigation at the moment

olblak (dismissed):
In the current state, this role doesn't works with puppet6

olblak:
@rtyler  "Cannot find libapache2-mod-mirrorbrain" Is it the package that you built?

rtyler:
Hm, that package should exist. I'm provisioning a machine so I can get Vagrant going here and will take a looksee

rtyler:
I've been tinkering with this some more in Vagrant, and it looks like the openSUSE Build Service has pushed expired GPG keys which prevent packages from being installed properly on newer Debian-based systems.

rtyler:
Just an update on this, the mirrorbrain repository is now properly provisionable for Debian hosts. It looks like the `python-mb` package tries to do some database accesses in the package's postinstall script (:roll_eyes:), but I don't think that is a blocker for us per  se

rtyler:
Rats, I found some other packaging problems with `mod_asn`. Looks like more hacking on packaging is in store for me!

rtyler:
Alright, verified as much as I can in Vagrant. Things are looking as good as they can/should look at the moment.

rtyler:
@olblak I'm going to deploy this and see how it goes :smile_cat:

PR #1269, opened 2019-03-24
Switch to Azure Container Instances for testing.

rtyler:
This removes some now redundant parts of the Jenkins Pipeline. Based on my testing it looks like this should be finishing in less than 10 minutes every time. :clap:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -40,7 +40,7 @@ pipeline {
                  * Ignoring spec/fixtures which can include non-ASCII data that
                  * causes problems with our stash
                  */
-                stash includes: '.bundle/**,vendor/**,spec/fixtures/modules/**',
+                stash includes: '.bundle/*,vendor/**/*,spec/fixtures/**/*',
                           name: 'deps',
                       excludes: 'vendor/**/spec/fixtures/**'
             }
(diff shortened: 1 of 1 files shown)


rtyler:
The branch build succeeded, but because of how the expected checks work, it's still "waiting" on a "pr-merge" check which Jenkins will never send.

Jenkinsfile
```diff
@@ -5,53 +5,51 @@ pipeline {
...
+    environment {
+        LANG = 'en_US.UTF-8'
+        LC_CTYPE = 'en_US.UTF-8'
+        /* These environment variables make it feasible for Git to clone properly while
+         * inside the wacky confines of a Docker container
+         */
+        GIT_COMMITTER_EMAIL = 'me@hatescake.com'
+        GIT_COMMITTER_NAME = 'Hates'
+        GIT_AUTHOR_NAME = 'Cake'
+        GIT_AUTHOR_EMAIL = 'hates@cake.com'
```
olblak: Should we really use this email "hates@cake.com" this domain exit, is used and we have no control on it? 
It would be better to use ci@jenkins.io (which doesn't exist) or jenkins@example.com
Same for me@hatescake.com

olblak (dismissed):
It would be nice to change the two emails, in order to not use domain that we don't control even if those emails are not really pushed in the git history. It would better for clarity.

olblak:
But globally this PR will be very useful, nice work

PR #1273, opened 2019-03-26
[INFRA-1352][INFRA-2005] Migrate truste.ci to azure

olblak:
This PR paves the way for trusted.ci to be running on azure, it also disable anonymous access on trusted.ci

### DIFF
--- a/hieradata/clients/cert-ci.yaml
+++ b/hieradata/clients/cert-ci.yaml
@@ -1,43 +1,10 @@
 ---
 profile::buildmaster::ci_fqdn: 'cert.ci.jenkins.io'
+profile::buildmaster::groovy_d_agent_security: 'absent'
+profile::buildmaster::groovy_d_enable_ssh_port: 'absent'
+profile::buildmaster::groovy_d_lock_down_jenkins: 'absent'
+profile::buildmaster::groovy_d_pipeline_configuration: 'absent'
+profile::buildmaster::groovy_d_set_up_git: 'absent'
+profile::buildmaster::groovy_d_terraform_credentials: 'absent'
+profile::buildmaster::proxy_port: 1443
 profile::buildmaster::letsencrypt: false
-# These are plugins we need in our ci environment
-profile::buildmaster::plugins:
-  - ansicolor
-  - azure-container-agents
-  - azure-vm-agents
-  - blueocean
-  - build-timeout
-  - buildtriggerbadge
-  - checkstyle
-  - cloudbees-folder
-  - credentials
-  - credentials-binding
-  - docker-workflow
-  - embeddable-build-status
-  - findbugs
-  - git-client
-  - git
-  - github
-  - github-branch-source
-  - github-organization-folder
-  - groovy
-  - jacoco
-  - jobConfigHistory
-  - junit-attachments
-  - junit-realtime-test-reporter
-  - ldap
-  - lockable-resources
-  - mailer
-  - parallel-test-executor
-  - pipeline-githubnotify-step
-  - pipeline-stage-view
-  - pipeline-utility-steps
-  - pmd
-  - ssh-agent
-  - throttle-concurrents
-  - timestamper
-  - toolenv
-  - warnings
-  - workflow-aggregator
-  - workflow-multibranch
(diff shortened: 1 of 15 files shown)


manifests/site.pp
```diff
@@ -148,6 +148,12 @@
   include role::jenkins::master
 }
 
+node 'trusted.ci.jenkins.io' {
+  $hiera_role = 'trustedci'
+  sshkeyman::hostkey { ['trusted.ci.jenkins.io', 'ci.trusted.jenkins.io']: }
```
rtyler: What's the second hostkey for here? Are you planning to put more under "trusted.jenkins.io" in the future?

rtyler (commented):
Looks reasonable overall, one minor question, but don't let that hold up progress here.

hieradata/clients/trusted.ci.jenkins.io
```diff
@@ -0,0 +1,45 @@
+---
+
+profile::buildmaster::ci_fqdn: 'trusted.ci.jenkins.io'
+profile::buildmaster::proxy_port: 1443
+profile::buildmaster::anonymous_access: false
+profile::buildmaster::letsencrypt: false
+# These are plugins we need only in our trusted-ci environment
+profile::buildmaster::plugins:
```
daniel-beck: Could I get an empty list here please?

manifests/site.pp
olblak: It was a blind stupid copy/paste from above :D written by a tall guy :D

olblak:
@rtyler Can you take another look at this PR? Basically I am removing the dependency on jenkins puppet module which is not used anymore as we use Jenkins from a docker image and I am also conditionnaly installing groovy init  script in order to being able to remove them on cert.ci as demanded by @daniel-beck

hieradata/clients/cert-ci.yaml
```diff
@@ -1,44 +1,10 @@
 ---
 profile::buildmaster::ci_fqdn: 'cert.ci.jenkins.io'
+profile::buildmaster::groovy_d_agent_security: 'absent'
```
daniel-beck: This looks like opt-out, which is likely error prone…?

olblak: Puppet resources use the parameter ensure => "present or absent" to know if a resource must be installed or removed. In this case it's used [here](https://github.com/jenkins-infra/jenkins-infra/blob/72adc28ca884e47ae15f8ad9302ba8268b8b0580/dist/profile/manifests/buildmaster.pp#L145) and means that for the cert-ci instance, that groovy script must be absent

daniel-beck: When someone adds a new Groovy script for use by ci.jenkins.io, do they need to remember to edit this file? If so, that seems like a problem as that would easily be missed.

olblak: This puppet class only install/desinstall groovy script defined in this puppet repository. 
The approach is:
1) If you don't need templating, you can add your specific groovy script in this [directory](https://github.com/jenkins-infra/jenkins-infra/tree/b461b54dbb50101a9ee2552e37219775df278916/dist/profile/files/buildmaster)
2) If you need to templating, put your template in this [directory](https://github.com/jenkins-infra/jenkins-infra/tree/b461b54dbb50101a9ee2552e37219775df278916/dist/profile/templates/buildmaster) like [lockbox.groovy.erb](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/templates/buildmaster/lockbox.groovy.erb)
 3) Update the buildmaster puppet [manifest](https://github.com/jenkins-infra/jenkins-infra/blob/b461b54dbb50101a9ee2552e37219775df278916/dist/profile/manifests/buildmaster.pp)
with your groovy script either based on a [emplate](https://github.com/jenkins-infra/jenkins-infra/blob/b461b54dbb50101a9ee2552e37219775df278916/dist/profile/manifests/buildmaster.pp#L188-L178) or based on a [file](https://github.com/jenkins-infra/jenkins-infra/blob/b461b54dbb50101a9ee2552e37219775df278916/dist/profile/manifests/buildmaster.pp#L176-L166)

Next time puppet agent run it will read and apply the manifest.

Parameter are customized per node from this [directory](https://github.com/jenkins-infra/jenkins-infra/tree/b461b54dbb50101a9ee2552e37219775df278916/hieradata/clients)

PR #1279, opened 2019-03-30
Prohibit abusive crawler from accessing ci.jenkins.io

daniel-beck:
This crawler ignores `robots.txt` despite claiming support for it at http://www.sogou.com/docs/help/webmasters.htm#07


rtyler (approved):
The CI failure is some weird transient network issue

PR #1303, opened 2019-05-13
[INFRA-2099] Use only 4G Xmx on cert.ci

daniel-beck:


rtyler (approved):
outta be enough for anybody

PR #1306, opened 2019-05-27
[INFRA-2115] Prepare migration of ci.jenkins.io to Java 11

alecharp:
The GC logging was changed using [this reference](https://docs.oracle.com/javase/9/tools/java.htm#JSWOR-GUID-BE93ABDC-999C-4CB5-A88B-1994AAAC74D5).

Moving away from G1 to ZGC, which made some parameters useless as well. We could choose not to perform this "migration" at the same time, as G1 is still available. 

cc @jenkins-infra/java11-support

### DIFF
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -227,7 +227,7 @@ class profile::buildmaster(
     env              => [
       "HOME=${jenkins_home}",
       'USER=jenkins',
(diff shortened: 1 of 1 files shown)


dist/profile/manifests/buildmaster.pp
```diff
@@ -211,7 +211,7 @@
   ##############################################################################
 
   docker::run { 'jenkins':
-    image            => 'jenkins/jenkins:lts-alpine',
+    image            => 'jenkins/jenkins:lts-jdk11',
```
olblak, line 214: This will configure every jenkins instance, including trusted and cert. 
Instead we must parametrize this variable and specifically configure ci.jenkins.io client with that tag

olblak: This [PR](https://github.com/jenkins-infra/jenkins-infra/pull/1305) parametrize this puppet class

alecharp: This is definitively not what I wanted. Will I still be able to change the `JAVA_OPTS` then?

olblak: Variables defined [here](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/manifests/buildmaster.pp#L33-L17), can be over-ridden per client in the hieradata directory, ci.j.io client configuration is located [here](https://github.com/jenkins-infra/jenkins-infra/blob/staging/hieradata/clients/ci.yaml)  
Now that I parametrized the docker image, you can just add a new variable in the client hieradata.
We must also parametrized JAVA_OPTS in order to use different options per environment

PR #1314, opened 2019-06-05
Fix the redirect under the language of zh

LinuxSuRen:
@olblak I missed the setting before, I thought it's ok for the nginx.
If I visit https://jenkins.io from China, I'll be redirected to localhost/zh.


donhui (approved):
LGTM

zacker330 (approved):

olblak:
Yes I introduced that stupid mistake :/

LinuxSuRen:
@olblak No worries.

daniel-beck:
This feature prevents users with `zh-CN` Accept-Language from ever using jenkins.io unless they land directly on a different page than the homepage. Are there plans to address this? Otherwise I don't see why we would keep this automatic language redirect around.

PR #1355, opened 2019-09-03
Rotate ldap admin password

olblak:


oleg-nenashev (approved):
From what I see, all entries have been updated

PR #1373, opened 2019-10-29
Fix the language order issues

### LINKED ISSUE #38: Translate task: about/artwork

LinuxSuRen:
Fix jenkins-infra/cn.jenkins.io#38. Inspire from @muellert which in [here](https://github.com/jenkins-infra/cn.jenkins.io/issues/238#issuecomment-547366920).

### DIFF
--- a/dist/profile/templates/kubernetes/resources/jenkinsio/configmap.yaml.erb
+++ b/dist/profile/templates/kubernetes/resources/jenkinsio/configmap.yaml.erb
@@ -6,7 +6,7 @@ data:
   default.conf: |
     map $http_accept_language $lang {
       default '';
-      ~zh zh;
+      ~^zh zh;
     }
 
     server {


olblak:
I am not sure to understand how this regex will fix the issue, but it worth trying, thanks for this PR

olblak (approved):
\o/

muellert:
> I am not sure to understand how this regex will fix the issue, but it worth trying, thanks for this PR

I haven't tried it, but the nginx docs suggest that if the expression starts with '~', it is being interpreted as a regular expression. IOW, the new regular expression should only fire if 'zh' appears at the beginning of the line. Have a look at your browser's accept-language headers. If you have only one language configured, just add another one temporarily to see the effect.

LinuxSuRen:
@olblak It seems isn't working. I guess this fix might not be deployed yet. So could you let me know this? Or when it is ready.

olblak:
@LinuxSuRen you are right, for some reason the second PR from staging to production wasn't trigger, I just create it

LinuxSuRen:
It's working now! Thanks to @olblak  and @muellert .

PR #1377, opened 2019-10-31
[INFRA-2310] Move back to G1GC + use 16Gb of Heap

aheritier:
I reverted our changes of this morning because they aren't J11 compatible and it introduces a lot of changes at the same time which is making difficult to understand if they are helping or not.


rtyler:
This looks fine for me, but I've been very out of the loop this week so I'll let @olblak handle this

olblak:
Because of the limited time we had yesterday to revert this change, I disabled puppet agent on that machine and allowed @aheritier to do the quick changes needed to revert to a working state. @aheritier can you confirm that this PR represent the state of ci.jenkins.io and so can be merged

olblak (approved):
At least we know it worked

aheritier:
@olblak yes I copy/paster the prod settings. I am preparing a new PR based on @dohbedoh feedback

PR #1379, opened 2019-11-01
[INFRA-2316] Improve GC settings

aheritier:
[INFRA-2316](https://issues.jenkins-ci.org/browse/INFRA-2316)

* GC logs are stored under the subfolder gc
* no more timestamp in the log filename (only the rotation done by the JVM based on the file size)
* Increase the file size to 40Mb instead of 20Mb to get more history

### DIFF
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -20,7 +20,8 @@ profile::buildmaster::groovy_d_lock_down_jenkins: 'present'
 profile::buildmaster::groovy_d_pipeline_configuration: 'present'
 profile::buildmaster::groovy_d_set_up_git: 'present'
 profile::buildmaster::groovy_d_terraform_credentials: 'present'
-profile::buildmaster::java_opts: "-server -Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc-%t.log::filecount=5,filesize=20M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2" # Java11 specific configuration
+# ! java_opts needs to be java11 compliant
+profile::buildmaster::java_opts: "-server -Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc-%t.log::filecount=5,filesize=20M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2"
 # These are plugins we need in our ci environment
 profile::buildmaster::plugins:
   - ansicolor
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -20,7 +20,7 @@ profile::buildmaster::groovy_d_lock_down_jenkins: 'present'
 profile::buildmaster::groovy_d_pipeline_configuration: 'present'
 profile::buildmaster::groovy_d_set_up_git: 'present'
 profile::buildmaster::groovy_d_terraform_credentials: 'present'
-profile::buildmaster::java_opts: "-server -Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc-%t.log::filecount=5,filesize=20M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2" # Java11 specific configuration
+profile::buildmaster::java_opts: "-server -Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc/gc.log::filecount=5,filesize=40M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2" # Java11 specific configuration
 # These are plugins we need in our ci environment
 profile::buildmaster::plugins:
   - ansicolor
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):
Fine with me

timja (approved):

rsandell (approved):

daniel-beck:
> * no more timestamp in the log filename (only the rotation done by the JVM based on the file size)

TBH, I liked the timestamps, as they allow us to keep track of restarts, AFAIUI.

> * Increase the file size to 40Mb instead of 20Mb to get more history

Why this over `filecount=10`?

daniel-beck:
Would a Jenkins restart now also remove previous GC logs that would explain why a restart is necessary in the first place? If so, that seems like a bad idea…

aheritier:
> TBH, I liked the timestamps, as they allow us to keep track of restarts, AFAIUI.

Is it really the best approach we have to track restarts ? We should have them in datadog I imagine.
The others solution if we want to keep the timestamps is to configure a logrotate to purge them ..

> Why this over filecount=10?

Like you prefer. I have no real preference.  On CloudBees side we advice to our customers to use 2x 100mb ( https://support.cloudbees.com/hc/en-us/articles/222446987 ). Not sure where this recommandation comes from @ryansmith303 @omehegan

> Would a Jenkins restart now also remove previous GC logs that would explain why a restart is necessary in the first place? If so, that seems like a bad idea…

`gc.log.0` is the oldest file and when the JVM restarts it is the one to be replaced

More details in https://blog.gceasy.io/2016/11/15/rotating-gc-log-files/ (but my understanding is that there is no perfect solution)

daniel-beck:
> Like you prefer.

Mostly asking for clarification, if it's essentially the same, works for me.

> but my understanding is that there is no perfect solution

IMO would prefer we continue time stamping. <5GB in half a year doesn't look like it's enough to matter in the big picture, especially once they're out of scope for support-core.

Not a strong opinion though, just concerned about losing information, especially in the case of relatively quick successive restarts.

olblak:
> GC logs are stored under the subfolder gc
What about the suggestion to move the subfolder outside jenkins_home? 
We could create a new directory call /var/logs/jenkins/gc and mount it at the same location inside the container

aheritier:
@olblak it could be the best, but I'm not sure how to setup it without breaking everything :-)

olblak:
> Is it really the best approach we have to track restarts ? We should have them in datadog I imagine.
We could monitor processes as provided by [Datadog](https://docs.datadoghq.com/graphing/infrastructure/process/?tab=linuxwindows), it's just a configuration change

hieradata/clients/azure.ci.jenkins.io.yaml
```diff
@@ -20,7 +20,7 @@ profile::buildmaster::groovy_d_lock_down_jenkins: 'present'
 profile::buildmaster::groovy_d_pipeline_configuration: 'present'
 profile::buildmaster::groovy_d_set_up_git: 'present'
 profile::buildmaster::groovy_d_terraform_credentials: 'present'
-profile::buildmaster::java_opts: "-server -Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc-%t.log::filecount=5,filesize=20M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2" # Java11 specific configuration
+profile::buildmaster::java_opts: "-server -Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc/gc.log::filecount=5,filesize=40M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2" # Java11 specific configuration
```
aheritier:
```suggestion
profile::buildmaster::java_opts: "-server -Xlog:gc*=info,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc/gc.log::filecount=5,filesize=40M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2" # Java11 specific configuration
```

@ryansmith303 proposes that we remove the debug GC logs and thus use these settings. cc @olblak @daniel-beck @batmat @rsandell @oleg-nenashev

daniel-beck: The continued presence of this after someone identified it as a problem in the call confused me anyway.

daniel-beck:
Could you clarify what the GC settings are _right now_? Are we still on the original, bad settings, or did we manually apply these?

olblak:
```
-e "JAVA_OPTS=-server -Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc-%t.log::filecount=5,filesize=20M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2" \
```

hieradata/clients/azure.ci.jenkins.io.yaml
aheritier: yes it was never fixed. 

Originally we had (ref: https://github.com/jenkins-infra/jenkins-infra/blob/d6e9e1c8ab46b0c89816a604f57a5ed26a724437/hieradata/clients/azure.ci.jenkins.io.yaml#L23 ):

```
-Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc-%t.log::filecount=5,filesize=20M
``` 

We now proposes

```
-Xlog:gc*=info,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc/gc.log::filecount=5,filesize=40M
```

* Info level by default
* No timestamp in the filename
* Logs not directly under JENKINS_HOME (but @olblak is verifying if they could be outside of JENKINS_HOME instead in a subfolder)
* Larger files (40Mb instead of 20Mb)

daniel-beck: > if they could be outside of JENKINS_HOME instead in a subfolder

What's the benefit of doing that?

daniel-beck:
```suggestion
profile::buildmaster::java_opts: "-server -Xlog:gc+heap*=debug,ref*=debug,ergo*=trace,age*=trace:file=/var/jenkins_home/gc/gc.log::filecount=5,filesize=40M -XX:+UnlockExperimentalVMOptions -XX:+UseG1GC -XX:+ParallelRefProcEnabled -XX:+UnlockDiagnosticVMOptions -Xms16g -Xmx16g -Duser.home=/var/jenkins_home -Djenkins.install.runSetupWizard=false -Djenkins.model.Jenkins.slaveAgentPort=50000 -Dhudson.model.WorkspaceCleanupThread.retainForDays=2"
```

G1 seems hardly Java 11 specific.

aheritier: G1 is available even on J8 but for sure these settings at J11 specific (at least the GC ones are J9+)
Note: For sure this comment wasn't enough to avoid me to propose some erroneous GC settings last week :(

aheritier: I think it is to avid to have them grabbed by default with by support-core but my understanding is that in any case it doesn't work with J11 (@dohbedoh ?)
Also potentially to not archive/backup them but now if we don't use a timestamp it's maybe not an issue...

oleg-nenashev: I am surprised we have no class unloading in the list. IIRC it still needs an explicit flag if we use any groovy scripting on the instance

daniel-beck: > I think it is to avid to have them grabbed by default with by support-core

Already accomplished through the subdirectory.

> Also potentially to not archive/backup them

5GB so far don't even matter among all the other nonsense.

aheritier: I agree @oleg-nenashev . Probably another PR to do

aheritier:
I merged your change @olblak

PR #1397, opened 2019-11-28
[INFRA-2357] Disable pluginsite deployment from puppet

olblak:
As the plugins.jenkins.io is now deployed from [jenkins-infra/charts](https://github.com/jenkins-infra/charts), we can now remove it from puppet


timja (approved):

olblak:
Remark this does not remove the resources from the current cluster, and it will require some manual step

PR #1409, opened 2020-01-20
Add markewaite ed25519 public key

MarkEWaite:
Passphrase protected in my key store


olblak (approved):
As part of INFRA-2437

PR #1410, opened 2020-01-28
Report commonly used wiki urls

halkeye:
In an effort to not have split results for google, i'm starting to add redirects to the wiki for migrated content. #1411 will add apache redirects for all the existing plugins, but after that its harder to do programatically.

Like for example, https://wiki.jenkins.io/display/JENKINS/Building+Jenkins should be linked to https://github.com/jenkinsci/jenkins/blob/master/CONTRIBUTING.md

---

This script will report the top 100 non redirected urls

Implementation ideas:
1) GIST
* Will update the same gist($GIST_ID) every day, so it always has the latest update
* Will need to login to gist cli tool which is not included in this patch - https://github.com/defunkt/gist#login

2) Mailing list
* Email a mailing list (new?) with the daily report

3) File on disk (Current)
* Create a file on disk, and update apache implementation to be able to serve that file
* /.well-known/yesterdays-urls.txt or something

### DIFF
--- a/dist/profile/files/confluence/report_last_log.sh
+++ b/dist/profile/files/confluence/report_last_log.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+GIST_ID=
+# grab all log files but sort by date
+find /var/log/apache2/wiki.jenkins-ci.org -name 'access.log*' -type f -printf "%T+\t%p\n" | sort | \
+  # grab the two newest
+  tail -n 2 | \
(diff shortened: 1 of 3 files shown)


MarkEWaite (approved):
Looks good to me.  The shell script looks workable.

halkeye:
@rtyler updated, though I sorta like idea #3 better now

rtyler (approved):
This looks interesting and worth trying :smile_cat:

PR #1414, opened 2020-02-04
Ignore our reports on proxy and fix the directory name

rtyler:


halkeye (approved):

rtyler (commented):
This change was already applied in production to unblock @halkeye

olblak (approved):

PR #1421, opened 2020-02-07
Upgrade the wiki usage reporting script

halkeye:
* Record which filename was last used for future debugging
* Drop sorting by file time, but sort by filename, which shouldn't be affected by gzipping or anything else
* Record last run time to make sure we can tell its running
* Set all the bash error reporting to make sure it bails if anything goes wrong
* support running with FINALFILE=- for debugging

### DIFF
--- a/dist/profile/files/confluence/report_last_log.sh
+++ b/dist/profile/files/confluence/report_last_log.sh
@@ -1,16 +1,27 @@
 #!/bin/bash
+set -euxo pipefail
 # make sure all can read this file
 umask 100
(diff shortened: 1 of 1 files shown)


halkeye:
example output:
```
# LogFile: access.log.20200202000000.gz
# LastUpdated: Fri Feb  7 19:32:38 UTC 2020
      1 /display/JENKINS/foo
```

rtyler (approved):
I'M SURE ITS FINE

PR #1446, opened 2020-03-09
Push artifacts on azure file storage from sync.sh

olblak:
Update sync.sh to also push packages on an azure file storage that can be used by this [service](https://github.com/jenkins-infra/charts/pull/122)

Ps: Because puppet agent cannot run anymore on pkg.jenkins.io, once approved, I'll manually change the file


olblak:
Currently it's taking 15min to push artifacts from pkg.jenkins.io to the azure file storage, I was able to reduce it to 10min by not checking md5 hash on non json files

rtyler (approved):

oleg-nenashev (approved):

olblak:
I won't use this method to upload artifacts to this blob storage, based on my test this method is way to slow

PR #1447, opened 2020-03-16
Allow accessing the hosting plugins page

timja:
Oleg said there's some missing content to be migrated


halkeye:
I would just remove the redirect for now? @oleg-nenashev do you want full access for everyone, or just the wiki exporter to be able to dump content?

timja:
> I would just remove the redirect for now? @oleg-nenashev do you want full access for everyone, or just the wiki exporter to be able to dump content?

image

It's in a thread on gitter, you might not be able to see it in your irc bridge

halkeye:
No I saw it, I took it as "Remove the redirect so people can see the missing content" not "Allow content to be fetched from wiki exporter so we can move more content"

timja:
> No I saw it, I took it as "Remove the redirect so people can see the missing content" not "Allow content to be fetched from wiki exporter so we can move more content"

k, I would rather just migrate it, but I don't know what's left...

halkeye:
Yea, I much rather migrate too, but until it is migrated, I think it should be opened up.

olblak (approved):

oleg-nenashev:
Thanks all!

PR #1452, opened 2020-03-26
Redirect Groovy Hooks Wiki to the recent jenkins.io page

oleg-nenashev:
https://jenkins.io/doc/book/managing/groovy-hook-scripts/ contains more documentation than the historical Wiki page. So we can just add a redirect


olblak (approved):
When will we know that we redirected every requests going to wiki.jenkins.io? :D

oleg-nenashev:
> When will we know that we redirected every requests going to wiki.jenkins.io? :D

One small step at a time, but we will get there.
If there is some Google analytics for Wiki, it would help to prioritize the migration

PR #1463, opened 2020-04-14
Deploy to production

olblak:


oleg-nenashev (approved):
Maintaining ec2 plugin will hurt. But it looks like a way to go

PR #1467, opened 2020-04-28
Do not escape the '#' in Starting & Accessing redirect

MarkEWaite:
The 'NE' flag ([RewriteRule flag](http://httpd.apache.org/docs/current/mod/mod_rewrite.html#rewriteflags)) is used successfully with the other URL in the rewrite which uses an ID inside the page.  Use it here as well.

Fixes [jenkins.io issue 3097](https://github.com/jenkins-infra/jenkins.io/issues/3097)

@halkeye could you review this to confirm that it is safe to use the same technique for "Starting and Accessing Jenkins" as is used for "Disabling Security"?


halkeye:
Yea, there's no querystring or regex capture that could potentially do harm or anything. I think its a safe thing to do

Docs - https://httpd.apache.org/docs/2.4/rewrite/flags.html#flag_ne

halkeye (approved):

MarkEWaite:
@olblak or @rtyler would you be willing to merge this?

@halkeye and I have both confirmed it matches the other case of using a RewriteRule with a URL that contains a '#' and it matches the Apache documentation.

olblak (approved):

PR #1477, opened 2020-05-15
Fixed Wiki page redirect pull request #3212

bigdevlarry:
Wiki pages are migrated one at a time. When the content of a wiki page has been migrated from wiki.jenkins.io and merged into www.jenkins.io, we can "hide" the wiki page by redirecting all requests for that page to the replacement page on www.jenkins.io. A page redirect pull request is used to define that redirect.
Capture jenkins fix

### DIFF
--- a/dist/profile/templates/confluence/vhost.conf
+++ b/dist/profile/templates/confluence/vhost.conf
@@ -2998,6 +2998,8 @@ RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
 RewriteRule "^/display/JENKINS/Features\+controlled\+by\+system\+properties$" "https://jenkins.io/doc/book/managing/system-properties/" [NC,L,QSA,R=301]
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
 RewriteRule "^/display/JENKINS/Groovy\+Hook\+Script$" "https://jenkins.io/doc/book/managing/groovy-hook-scripts/" [NC,L,QSA,R=301]
+RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
+RewriteRule "^/display/JENKINS/CSRF\+Protection$" "https://www.jenkins.io/doc/book/managing/security/#cross-site-request-forgery" [NE,NC,L,QSA,R=301]
 
 ## Developer documentation
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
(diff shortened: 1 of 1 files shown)


MarkEWaite:
Thanks @LarrySul .  That looks very good.  

You can help the reviewers better understand the content of future pull requests by describing the content of the pull request more specifically in the title.  In this case, you might have called the pull request:

## Redirect CSRF protection wiki page to jenkins.io

That helps the reviewers understand more clearly the content of the pull request and makes it easier for them to find the pull request when scanning later lists of pull requests.

It also helps reviewers if you embed a link to other information, like the original issue which prompted this pull request.  See https://github.com/jenkins-infra/jenkins.io/issues/3212

dist/profile/templates/confluence/vhost.conf
```diff
@@ -2998,6 +2998,8 @@ RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
 RewriteRule "^/display/JENKINS/Features\+controlled\+by\+system\+properties$" "https://jenkins.io/doc/book/managing/system-properties/" [NC,L,QSA,R=301]
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
 RewriteRule "^/display/JENKINS/Groovy\+Hook\+Script$" "https://jenkins.io/doc/book/managing/groovy-hook-scripts/" [NC,L,QSA,R=301]
+RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
+RewriteRule "^/display/JENKINS/CSRF\+Protection$" "https://www.jenkins.io/doc/book/managing/security/#cross-site-request-forgery" [NE,NC,L,QSA,R=301]
```
MarkEWaite: Should follow the same URL pattern as elsewhere in the file, using `jenkins.io` instead of `www.jenkins.io`.

```suggestion
RewriteRule "^/display/JENKINS/CSRF\+Protection$" "https://jenkins.io/doc/book/managing/security/#cross-site-request-forgery" [NE,NC,L,QSA,R=301]
```

MarkEWaite (changes requested):
Needs a change in the redirect URL for consistency with other redirects.  May need discussion with @olblak  to assure that the redirect should point to `jenkins.io` rather than `www.jenkins.io`

dist/profile/templates/confluence/vhost.conf
olblak: While I totally agree that it should be set to 'www' to remove one redirection, for now, I agree with @MarkEWaite for consistency it's better to use the same base URL than the others,

olblak (approved):

PR #1502, opened 2020-05-29
Redirect Authenticating scripted clients from wiki to jenkins.io

zaycodes:
Redirect Authenticating scripted clients from wiki to jenkins.io


zaycodes:
Hello @MarkEWaite, kindly review this. I tried to do a merge with the upstream and that added the extra commits. Not sure how to handle that

PR #1541, opened 2020-09-24
Updated redirection for tfs plugin

hp77-creator:
As per the [issue](https://github.com/jenkins-infra/jenkins.io/issues/3723), Updated conf. file with redirection for tfs plugin.


markjacksonfishing (approved):

oleg-nenashev (commented):
https://github.com/jenkinsci/tfs-plugin/pull/237 needs to be fixed, integrated and released before we can merge it

MarkEWaite:
@oleg-nenashev I don't think that redirecting away from the

MarkEWaite:
TFS plugin has been removed from distribution because it includes components that are not open source.

PR #1590, opened 2020-10-22
Add redirects to plugin site

andreamussap:
I've updated section *Rewrite all plugin docs to plugins.jenkins.io if user agent is not the wiki exporter* to add three links/redirects, following examples already on the page.

I also fixed the following two entries to add `\+Plugin$` to them:
* display/JENKINS/violation\+comments\+to\+gitlab\+Plugin$
* display/JENKINS/IBM\+zOS\+Connector\+Plugin$

This PR fixes

- https://github.com/jenkins-infra/jenkins.io/issues/3777
- https://github.com/jenkins-infra/jenkins.io/issues/3773
- https://github.com/jenkins-infra/jenkins.io/issues/3759

PR #1591, opened 2020-10-23
Fix redirect links to plugin site

andreamussap:
Complements #1590


andreamussap:
@MarkEWaite I've addressed your comments on #1590, and I've converted this to DRAFT so it won't be merged until you review my changes. Thanks, Andrea

olblak (approved):

MarkEWaite (approved):
Thanks very much.  I checked the URLs and they are correct.

No need to mark a pull request as "draft", since most reviewers will see "Draft" and think "I don't need to review this until it is no longer a draft".

PR #1594, opened 2020-10-23
Add confluence URL checks in CI

MarkEWaite:
## Add confluence URL checks in CI

Keep the checks lightweight by recording the expected URLs in local files, one for the URLs that are known to exist in Confluence and one for the URLs that are known to not exist in Confluence.

Also includes minor fixes to content of vhost.conf so that the tests pass.


Jenkinsfile
```diff
@@ -63,6 +63,8 @@ pipeline {
                         // Check that Confluence rewrite rules don't duplicate the mistake of adding an extra
                         // '/' after the JENKINS portion of the URL.
                         sh 'if grep JENKINS// dist/profile/templates/confluence/vhost.conf; then echo "Extra / after JENKINS in Confluence URL"; exit 1; fi'
+                        // Check confluence URLs from vhost.conf file
+                        sh 'scripts/verify_confluence_urls'
```
halkeye: since its in a ruby agent, do you want me to rewrite it in ruby so it doesn't have so many magical seds in it?

halkeye:
Its a really cool idea. Essentially a unit test. Add the url your trying to redirect to the list, and that'll confirm it stays redirecting in the future.

markjacksonfishing (approved):

Jenkinsfile
MarkEWaite: @halkeye I would be delighted if it were rewritten in ruby.  I have no ruby skills and didn't think it was the right approach to write it in Python then call it from a Ruby container. 

I'm sure it would be much more readable and maintainable in ruby than the current ugliness that the Unix shell and utilities provide.  If you're willing, that would be great for me.  I've been able to do some limited maintenance on ruby but have not spent enough time with it to develop any real skills with it.

halkeye: I'm doing ruby full time again, so its a good excuse to do a bit of practice :)

halkeye:
So I'm very confused why drizzle is written the way it is

https://github.com/jenkinsci/database-drizzle-plugin/blob/master/pom.xml#L14 points to a different url, so `/display/JENKINS/Drizzle(+MySQL)%20Database%20Plugin` is not the redirect that is needed? I think? maybe?

New ruby version - https://github.com/halkeye/jenkins-infra/blob/check-vhost-urls/scripts/verify_confluence_urls.rb

I'm not sure its better, its longer, but it might be more readable.

I also made it sort the url files so merge conflicts are easier.

That being said, it did find a bunch of new urls for the files - https://github.com/halkeye/jenkins-infra/commit/453232d380b0affa1c0510fb477ed38fe316b1b8

MarkEWaite:
> So I'm very confused why drizzle is written the way it is
> 
> https://github.com/jenkinsci/database-drizzle-plugin/blob/master/pom.xml#L14 points to a different url, so `/display/JENKINS/Drizzle(+MySQL)%20Database%20Plugin` is not the redirect that is needed? I think? maybe?
> 
> New ruby version - https://github.com/halkeye/jenkins-infra/blob/check-vhost-urls/scripts/verify_confluence_urls.rb

I like the ruby version but I don't understand how to fix an error that happens when I run it.  When I run it, the output is:

```
$ ./scripts/verify_confluence_urls.rb
HEAD for https://wiki.jenkins.io failed
HEAD for https://wiki.jenkins.io failed
======================== New redirect addition needs update to test data file ========================
diff --git a/scripts/confluence_urls_without_a_page b/scripts/confluence_urls_without_a_page
index 0f9a1dc..4ca2afc 100644
--- a/scripts/confluence_urls_without_a_page
+++ b/scripts/confluence_urls_without_a_page
@@ -13,6 +13,7 @@
 /display/JENKINS/D+Listing+Coverage+Plugin
 /display/JENKINS/Devstack+Plugin
 /display/JENKINS/Disable+GitHub+Multibranch+Status+Plugin
+/display/JENKINS/Drizzle(+MySQL)%20Database%20Plugin
 /display/JENKINS/Dynatrace+Plugin
 /display/JENKINS/FlexTeam+Plugin
 /display/JENKINS/Github+Custom+Notification+Context+SCM+Behaviour
======================== End of redirect addition update for test data file ========================
```

I can resolve the complaint about Drizzle by adding it to the confluence_urls_without_a_page file (even though there is a page that seems like it should match that name).

I can't resolve the messages:
```
HEAD for https://wiki.jenkins.io failed
HEAD for https://wiki.jenkins.io failed
```

Any hints?

halkeye:
I've updated the script to report errors and fixed the typo in that string ($uri vs uri)

HEAD for https://wiki.jenkins.io/display/JENKINS/Drizzle(+MySQL)%20Database%20Plugin failed: 404 ""
HEAD for https://wiki.jenkins.io/display/JENKINS/Play%21+Framework+Plugin failed: 404 ""

MarkEWaite:
> I've updated the script to report errors and fixed the typo in that string ($uri vs uri)
> 
> HEAD for https://wiki.jenkins.io/display/JENKINS/Play%21+Framework+Plugin failed: 404 ""

https://wiki.jenkins.io/display/JENKINS/Play+Framework+Plugin is the rename destination of that URL

halkeye:
> https://wiki.jenkins.io/display/JENKINS/Play+Framework+Plugin is the rename destination of that URL

is that a problem with my script or the data?

MarkEWaite:
> > https://wiki.jenkins.io/display/JENKINS/Play+Framework+Plugin is the rename destination of that URL
> 
> is that a problem with my script or the data?

I haven't considered what we should do with wiki pages that are renamed.  I'm open to recommendations.

halkeye:
> I haven't considered what we should do with wiki pages that are renamed. I'm open to recommendations.

for now, mark it as no page, and keep an eye on top urls list?

MarkEWaite:
> for now, mark it as no page, and keep an eye on top urls list?

That is an excellent idea.  Yes, mark it as no page and confirm that it is not a frequently accessed page.  Considering that plugin has been suspended from distribution due to unresolved security issues, I doubt it is accessed frequently (except by people like you and me investigating it)

halkeye:
Does https://github.com/halkeye/jenkins-infra/blob/check-vhost-urls/scripts/make_rewrite_rule.rb add any value too? all escaped and stuff?

MarkEWaite:
@halkeye I've not been able to make the updated code work. I've combined your branch and mine in this pull request so that they can both run in CI and I can use them to check other pull requests

halkeye:
> I've not been able to make the updated code work

not work how? What do you need me to do, I can totally look into it.

MarkEWaite:
> > I've not been able to make the updated code work
> 
> not work how? What do you need me to do, I can totally look into it.

I had not understood the details of the failure mode enough to give an adequate description earlier.  I think I understand it now.

When I run the ruby version of the script on Linux, it reports:
```
$ ./scripts/verify_confluence_urls.rb
HEAD for https://wiki.jenkins.io/display/JENKINS/Drizzle(+MySQL)%20Database%20Plugin failed: 404 ""
HEAD for https://wiki.jenkins.io/display/JENKINS/Play%21+Framework+Plugin failed: 404 ""
=========== Ruby: New redirect addition needs update to test data file ===========
=========== Ruby: End of redirect addition update for test data file =========
```

The ruby script reports two problems but returns a zero error code so the CI job will pass when it should return a non-zero error code to cause the CI job to fail.  I pushed a change to the ruby script to return the error code to the operating system.  I think that resolves the first issue I was having.

The second issue was that the two failing HEAD requests (Drizzle and Play%21+Framework) seem to indicate some inconsistency between the "not a page" file and the ruby code.  The "Drizzle" URL is in the "not a page" file but is still being checked with a HEAD request.  The "Play%21+Framework" is not in the "not a page" file, is being checked with a HEAD request, but no change is added to the "not a page" file to indicate that a HEAD request should not be used in the future.

halkeye:
> The ruby script reports two problems but returns a zero error code so the CI job will pass when it should return a non-zero error code to cause the CI job to fail. I pushed a change to the ruby script to return the error code to the operating system. I think that resolves the first issue I was having.

not a page isn't being checked first.  Infact nothing is being done with it :) Porting error I guess.
I updated my branch, but its simply

```
diff --git a/scripts/verify_confluence_urls.rb b/scripts/verify_confluence_urls.rb
index 74c2d2d..cd628f0 100755
--- a/scripts/verify_confluence_urls.rb
+++ b/scripts/verify_confluence_urls.rb
@@ -55,6 +55,8 @@ File.foreach(filename_vhost) do |line|

   # Found pattern in the list of working confluence pages
   next if confluence_urls.any? { |url| match_regex.match(url) }
+  # Found pattern in the list of working confluence pages
+  next if confluence_urls_without_a_page.any? { |url| match_regex.match(url) }

   uri = regex_to_url(match)
   begin
```

halkeye:
> The ruby script reports two problems but returns a zero error code so the CI job will pass when it should return a non-zero error code to cause the CI job to fail. I pushed a change to the ruby script to return the error code to the operating system. I think that resolves the first issue I was having.

yea I totally forgot to port the return_code check around diff, totally my bad

MarkEWaite:
Thanks @halkeye !  I've merged your change into the pull request.  It passes my tests.  Thanks for writing the Ruby code!

MarkEWaite:
@halkeye I rebased the current pull request, removed some of my mistakes from the rebased history, and have force pushed the rebase.  It passes the tests on my development environment and I expect it will pass tests on the CI job.

Special thanks to @TobiX for the significant improvement in the vhost.conf file by placing the wiki-exporter exclusion once in the file instead of many times.  It is much easier to read now.

PR #1595, opened 2020-10-25
Redirect more pages from Wiki to Plugin site

andreamussap:
- Fixes https://github.com/jenkins-infra/jenkins.io/issues/3749
- Fixes https://github.com/jenkins-infra/jenkins.io/issues/3751
- Fixes https://github.com/jenkins-infra/jenkins.io/issues/3752


markjacksonfishing (approved):

dist/profile/templates/confluence/vhost.conf
```diff
@@ -2857,6 +2857,12 @@ RewriteRule "^/display/JENKINS/App\+Center\+Plugin$" "https://plugins.jenkins.io
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
 RewriteRule "^/display/JENKINS/XML\+Job\+to\+Job\+DSL$" "https://plugins.jenkins.io/xml-job-to-job-dsl/" [NC,L,QSA,R=301]
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
+RewriteRule "^/display/JENKINS/Task\+Scanner\Plugin" "https://plugins.jenkins.io/warnings-ng//" [NC,L,QSA,R=301]
```
MarkEWaite: Missing the `+` character that separates the word "Scanner" from the word "Plugin" in the wiki URL.  Detected by the tests in #1594, though the message when detected is not clear about the cause, just that the page `/display/JENKINS/Task+ScannerPlugin` does not exist.
```suggestion
RewriteRule "^/display/JENKINS/Task\+Scanner\+Plugin" "https://plugins.jenkins.io/warnings-ng//" [NC,L,QSA,R=301]
```

MarkEWaite (changes requested):
One URL needs to be corrected before this is merged.

andreamussap:
> One URL needs to be corrected before this is merged.

Done. I've committed your fix. Thanks.

dist/profile/templates/confluence/vhost.conf
```diff
@@ -2857,6 +2857,12 @@ RewriteRule "^/display/JENKINS/App\+Center\+Plugin$" "https://plugins.jenkins.io
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
 RewriteRule "^/display/JENKINS/XML\+Job\+to\+Job\+DSL$" "https://plugins.jenkins.io/xml-job-to-job-dsl/" [NC,L,QSA,R=301]
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
+RewriteRule "^/display/JENKINS/Task\+Scanner\+Plugin" "https://plugins.jenkins.io/warnings-ng//" [NC,L,QSA,R=301]
+RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
```
olblak:
```suggestion
```
This line is duplicated with 2857

dist/profile/templates/confluence/vhost.conf
```diff
@@ -2857,6 +2857,12 @@ RewriteRule "^/display/JENKINS/App\+Center\+Plugin$" "https://plugins.jenkins.io
...
+RewriteRule "^/display/JENKINS/Xray\+Connector\+Plugin$" "https://plugins.jenkins.io/xray-connector/" [NC,L,QSA,R=301]
+RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
```
olblak:
```suggestion
RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
```
```suggestion
```
This line is duplicated with 2857

dist/profile/templates/confluence/vhost.conf
```diff
@@ -2857,6 +2857,12 @@ RewriteRule "^/display/JENKINS/App\+Center\+Plugin$" "https://plugins.jenkins.io
...
+RewriteRule "^/display/JENKINS/FitNesse\+Plugin$" "https://plugins.jenkins.io/fitnesse/" [NC,L,QSA,R=301]
+RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
```
olblak:
```suggestion
```
This line is duplicated with 2857

dist/profile/templates/confluence/vhost.conf
andreamussap: @olblak I don't get your comments about the duplication of `RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$`. When I looked at all redirects that were there before, it seems that this line is required (N-plicated) to each new redirect. And, the way the redirect works is by adding **two** lines, one with the links of the page, and another, the `RewriteCond` line ??

e.g, each redirect will have:

* `RewriteRule "^/display/JENKINS/Zoom\+Plugin$" "https://plugins.jenkins.io/zoom" [NC,L,QSA,R=301]` 

> With **specific** link for each new entry

* `RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$`

> **Repeated for ALL new entries**

So why only the 3 updates I've added are duplicating `RewriteCond`, but all the other entries in the file aren't?

olblak: They are just a lot of copy-paste that we should clean, :D  a `RewriteCond` is applied to every following `RewriteRule`
In this case, we always have the same condition that says if the user agent is not the jenkins-wiki-exporter then rewrite the url to the destination.

andreamussap: (*condition that says if the user agent is not the jenkins-wiki-exporter then rewrite the url to the destination*) - So, because we know for sure that these URLs are indeed from the Jenkins Wiki, we don't need the condition to rewrite the URL?

dist/profile/templates/confluence/vhost.conf
```diff
@@ -2857,6 +2857,9 @@ RewriteRule "^/display/JENKINS/App\+Center\+Plugin$" "https://plugins.jenkins.io
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
 RewriteRule "^/display/JENKINS/XML\+Job\+to\+Job\+DSL$" "https://plugins.jenkins.io/xml-job-to-job-dsl/" [NC,L,QSA,R=301]
 RewriteCond %{HTTP_USER_AGENT} !^jenkins-wiki-exporter/(.*)$
+RewriteRule "^/display/JENKINS/Task\+Scanner\+Plugin" "https://plugins.jenkins.io/warnings-ng//" [NC,L,QSA,R=301]
```
MarkEWaite: No need for the extra trailing slash
```suggestion
RewriteRule "^/display/JENKINS/Task\+Scanner\+Plugin" "https://plugins.jenkins.io/warnings-ng/" [NC,L,QSA,R=301]
```

MarkEWaite (approved):

dist/profile/templates/confluence/vhost.conf
TobiX: @olblak Just a note: The [Apache docs](http://httpd.apache.org/docs/2.2/mod/mod_rewrite.html#RewriteCond) say:

> One or more RewriteCond can precede a RewriteRule directive. The following rule is then only used if both the current state of the URI matches its pattern, and if these conditions are met.

So I think just leaving out those lines is not the right way...

TobiX: Cleanup done with #1608, so no need to bother with jenkins-wiki-exporter conditions anymore!

TobiX (approved):

PR #1598, opened 2020-10-26
Don't use docker image digest for ci.j & cert.ci

olblak:
After a discussion with @daniel-beck we agreed to not use updatecli for ci.jenkins.io.
It  will drastically slow down @daniel-beck as he manually update ci.jenkins.io after a security release.

The updatecli workflow implies

* New docker image published
* Updatecli to run and open a PR on jenkins-infra/jenkins-infra with new docker image digest
* Someone to validate and merge to staging
* Someone to open a PR from staging to production
* Someone to validate and merge to staging
* Puppet agent run on ci.jenkins.io

The preferred workflow for ci.jenkins.io is:
* New docker image published
* ssh on the machine
* systemctl restart docker-jenkins


MarkEWaite (approved):

daniel-beck (changes requested):
There seems to be a misunderstanding here.

ci.j.io needs to be updateable quickly during maintenance while a security advisory is pending, but cert.ci can use this. So far I've relied on https://github.com/jenkins-infra/jenkins-infra/blob/04e618e6cb95320ff1d2147981c0db7201a539d1/spec/classes/profile/buildmaster_spec.rb#L30 while updating plugins, which seemed good enough, but I'm not opposed to this behavior for cert.ci either.

daniel-beck (approved):

PR #1608, opened 2020-10-31
Reduce wiki-exporter special case to one rule

TobiX:
Currently, the exception for the wiki-exporter has to be specified for every rule (due too RewriteCond only [applying to the next rule](http://httpd.apache.org/docs/2.2/mod/mod_rewrite.html#RewriteCond)) , which bloats the (already pretty large) file.

I propose this "magic" rule to (somewhat) restore the readability of the file.

I tried this on a local install of Apache 2.4.46, but since this is pretty basic usage of mod_rewrite, it should be okay on any recent Apache version.

The only danger I see in this change is that if sometimes in the future rewrite rules are added to, for example, "hide" something from the public, care must be taken or otherwise someone can circumvent this "protection" by posing as "jenkins-wiki-exporter"


MarkEWaite:
The wiki.jenkins.io server runs Apache 2.4.29.  I assume that is close enough to 2.4.46 for this change.  Thanks for the pull request!

MarkEWaite (approved):

timja (approved):

PR #1609, opened 2020-10-31
Redirect popular nginx wiki page to official nginx documentation

TobiX:
According to https://wiki.jenkins.io/.well-known/reports/top_urls.txt there is a quite "popular" wiki page describing how to run Jenkins behind a nginx reverse proxy.

As an nginx user I found nothing on that page which isn't covered by the official documentation, so redirect to that instead.

(Quite on the contrary: The official documentation contains support for modern features like websockets, whereas the configs in the wiki page do not)


MarkEWaite (approved):
Thanks very much!  I thought we detected all the reverse proxy pages when we created that destination on www.jenkins.io.  I was wrong.  Thanks again for finding it and fixing it.

MarkEWaite:
@TobiX was this specifically a Hacktoberfest contribution?

TobiX:
@MarkEWaite Due to DO's changed rules, this actually counted against my Hacktoberfest contributions 😄

It wasn't specifically for Hacktoberfest, just trying to improve things I noticed while verifying my own configuration

oleg-nenashev:
Whatever works :)
Thanks for the patch!

PR #1624, opened 2020-11-17
Fix #3748

dskouris:
Fixes issue #3748


MarkEWaite (approved):

markjacksonfishing (approved):

oleg-nenashev (approved):
Screenshots need to be updated to avoid "slave", DEV@cloud has been also shut down. So there is some cleanup needed in the doc, but this pull request does not make it worse. Hence +1

olblak (approved):

PR #1627, opened 2020-11-20
Enable lets encrypt on trusted-ci

timja:


dist/profile/manifests/letsencrypt.pp
```diff
@@ -11,11 +11,4 @@
   package { 'python-certbot-apache':
     ensure => present
   }
-
-  cron { 'letsencrypt-renew-reload':
-    ensure  => present,
-    command => '/opt/letsencrypt/letsencrypt-auto renew --quiet --renew-hook="service apache2 reload"',
```
timja, line 17: this command doesn't exist on ci.jenkins.io (although still present in the crontab)
I assume left over from a previous version of the puppet module

hieradata/clients/trusted-ci.yaml
```diff
@@ -12,7 +12,7 @@ profile::buildmaster::anonymous_access: false
 profile::buildmaster::admin_ldap_groups:
   - admins
   - trusted-admins
-profile::buildmaster::letsencrypt: false
+profile::buildmaster::letsencrypt: true
```
timja, line 15: not sure why we have trusted-ci.yaml and trusted.ci.jenkins.io.yaml

olblak:
trusted.ci is not publicly available so this PR won't work. First we must configure puppet to use dns method with azure dns.

olblak:
I am closing this PR as it needs to be rewritten from a new branch

PR #1628, opened 2020-11-21
[INFRA-2825] - Wiki: Prevent search indexing of pages listed by pageId + disable indexing of page details and search results

oleg-nenashev:
This change prevents indexing of direct page view links like https://wiki.jenkins.io/pages/viewpage.action?pageId=97091655 (extension points). After we started adding redirects in vshost.conf, these pages started appearing in search instead. Adding robots.txt should help to prevent that

Inspired by https://gist.github.com/childnode/d15cd5e906a93fc1e11dcbdd55233b75 and https://wiki.scn.sap.com/robots.txt . Anyone is welcome to setup full rules, just trying to improve the redirect experience.

CC @halkeye @MarkEWaite @timja @getJv @zbynek @vsilverman as active documentation migration contributors


timja (approved):

zbynek:
there are some URLs that should not be indexed and are not covered by this change:
* `https://wiki.jenkins.io/plugins/viewsource/viewpagesrc.action?pageId=107775372`
* `https://wiki.jenkins.io/pages/replycomment.action?commentId=138456992&pageId=46334385`
* `https://wiki.jenkins.io/spaces/flyingpdf/pdfpageexport.action?pageId=79921356`
* `https://wiki.jenkins.io/pages/viewrecentblogposts.action?key=SECURITY`
* `https://wiki.jenkins.io/dashboard.action?updatesSelectedTab=all`
* `https://wiki.jenkins.io/pages/listpages-alphaview.action?key=JENKINS&startsWith=&startIndex=362`
* `https://wiki.jenkins.io/pages/listpages-dirview.action?key=JENKINS&openId=58002803`
* `https://wiki.jenkins.io/pages/recentlyupdated.action?key=JENKINS`
* `https://wiki.jenkins.io/plugins/recently-updated/changes.action?theme=concise&startIndex=45&searchToken=1339496191182&spaceKeys=JENKINS&contentType=page`

(all from Google search)

Also possibly any URL that contains a semicolon, e.g.
`https://wiki.jenkins.io/display/JENKINS/Git+Plugin;this-part-is-ignored-by-confluence-but-circumvents-redirect`

markjacksonfishing (approved):

oleg-nenashev:
Thanks @zbynek! I will add filters

zbynek:
> To remove content or a URL from Google search permanently:
>
> 1) Take one of the following actions to remove the page permanently:
>   * Remove or update the content on your site (images, pages, directories) and make sure that your web server returns either a 404 (Not Found) or 410 (Gone) HTTP status code. Non-HTML files (like PDFs) should be completely removed from your server. (Learn more about HTTP status codes)
>   * Block access to the content, for example by requiring a password.
Indicate that the page should not be indexed using the noindex meta tag. This is less secure than the other methods.
>   * **Do not use robots.txt as a blocking mechanism.**
> 2) If you blocked the page before removing your content permanently (step 1), unblock and then reblock the page. This clears the page from the index, if it was recrawled after blocking.

https://support.google.com/webmasters/answer/9689846

halkeye:
> there are some URLs that should not be indexed and are not covered by this change:

is it worth telling apache just to return forbidden for those unless like on vpn (I know dbeck sometimes uses them). It would be a simple regex and I'm not sure robots can handle regexp.

Actually re-reading @zbynek's second response, I really think forbidden is the right answer.

> Also possibly any URL that contains a semicolon, e.g.

same with this one

halkeye:
Looking at https://stackoverflow.com/a/21795992/358037 it would be something like
```
RewriteRule .*\.action\b - [F,NC]
RewriteRule .*;.* - [F,NC]
```

then
```
RewriteCond %{REMOTE_ADDR} !^xxx\.xxx\.xxx\.xxx$
```

matching the ips VPN supports?

halkeye:
or https://serverfault.com/a/771523 and just have a very basic htpassword in place (readonly/readonly) so it keeps out search engines, but anyone who wants those urls could use them.

olblak:
Any remaining work on this? Or do we proceed with the "disallow" approach?

olblak (approved):

timja:
> Any remaining work on this? Or do we proceed with the "disallow" approach?

we need to update it to include an htpassword really I think

timja (approved):

PR #1663, opened 2021-02-09
Switch jenkins lts docker image from digest to fixed version

olblak:
Updatecli can now retrieve the latest stable version, so I propose to use that instead of relying on the docker image digest which is more complicated for reviews.


timja (approved):

markjacksonfishing (approved):

timja:
>  The HTTP response we received was '504 Gateway Time-out'

MarkEWaite (approved):

PR #1765, opened 2021-05-17
Feat/remove okra lvm mngmt

dduportal:
After #1763 and #1761 , and given https://github.com/jenkins-infra/runbooks/pull/32, the subject of managing LVM for okra through Puppet has been [raised in IRC](https://matrix.to/#/!cddybYRxFljbtgddGf:matrix.org/$16212660393980171lFAaO:matrix.org?via=matrix.org&via=g4v.dev), leading to the following:

* The LVM volumes changes are really sensitive for this machine
* The LVM volumes changes are not done often enough (once per year at most) so it's hard to trust
* The requirement to reproduce the exact Puppet setup (exact size, exact amount of volumes) are working but not complete enough on vagrant
* The feedback loop between a change for LVM and when its effective in production it too high and requires too much back and forth

So the proposal is to shift the volume management out from Puppet:

* We keep maintaining the runbook operation to ensure knowledge is kept
* We could start automating using Terraform as Rackspace is Openstack compliant: https://docs.rackspace.com/docs/fanatical-support-aws/managed-infra-as-code/using-terraform


MarkEWaite (approved):
Thanks.  I think this is the right change.

PR #1776, opened 2021-06-01
Update authorized rsync connection on archives.jenkins.io

olblak:

For a reason that I can't explain yet, pods behind get.jenkins.io use a different IP address (52.177.88.13) for outbound connection than get.jenkins.io(52.167.253.43)


dduportal (approved):
🚀

timja (approved):

dduportal:
> For a reason that I can't explain yet, pods behind get.jenkins.io use a different IP address (52.177.88.13) for outbound connection than get.jenkins.io(52.167.253.43)

In fact, there are two distinct public IPs: the ingress and the egress one. The egress one is to be used for allow-list like here. For info, you can setup such IP (instead of the default Azure allocated one), for instance  with Terraform: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#outbound_type

PR #1783, opened 2021-06-14
Remove /cli blackhole

daniel-beck:
See https://github.com/jenkinsci-cert/SECURITY-218

Additionally, since 2.165 and 2.176.1, the remoting mode of the CLI doesn't exist anymore. I don't think there's a reason to prohibit these anymore now.


timja (approved):

dduportal (approved):
👍

PR #1789, opened 2021-06-21
feat: define container memory limit instead of xms/xmx for buildmasters

dduportal:
- ci.jenkins.io has 30g (for 32 Gb physical)
- trusted.jenkins.io has 7g (for 8 Gb physical)
- cert.ci.jenkins.io has 14g (for 16 Gb physical)
- vagrant dev. environment has 1.5g (for 2 gb physical)

The rationale is that the puppet agent and Docker Engine got OOMed
on trusted.ci in the past 4 days.

Signed-off-by: Damien Duportal <damien.duportal@gmail.com>

### DIFF
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -31,8 +31,7 @@ class profile::buildmaster(
   $groovy_d_pipeline_configuration = 'absent',
   $groovy_d_lock_down_jenkins      = 'absent',
   $groovy_d_terraform_credentials  = 'absent',
-  $java_opts_xms                   = '4g', # -Xms4g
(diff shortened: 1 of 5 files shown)


daniel-beck:
Does this limit heap size to the default of 256M everywhere? Did the behavior of Java 11 change? What am I missing?

dduportal:
I realize that I never added the answer here:

TL;DR; By removing any memory management java opt + defining the container limit, then JDK11 is defining the heap size to 25% of the container limit by default AND ensure that OOM is never triggers.

The JVM in JDK11 (or JDK8 with the correct experimental java opts defined) is able to read the cgroup limit and adapt accordingly.

We added a java opt to print the memory status on each JVM startup to ease diagnose in the future.

PR #1792, opened 2021-07-02
Add new archives profile and several puppet module update

olblak:
The goal of this PR is to add a new role that could be used from the new arm machine running on oracle.
Because the ubuntu version must be 20.04 and it's arm architecture, several improvement needed to be done

1. Only install azure-cli on amd64 machines
2. Update apt puppet module
3. Update Letsencrypt  puppet module
4. Update firewall puppet module

To test this pull request, I run `vagrant up -d archives` for both ubuntu version 18.04 and 20.04


dist/role/manifests/archives.pp
```diff
@@ -0,0 +1,6 @@
+#
+# Okra is a tiny VM (1vCPU/4GB RAM) on Oracle
```
MarkEWaite: I thought that we already had a machine named `okra`.  Should this have a distinct name?
```suggestion
# Okra is a tiny VM (1vCPU/4GB RAM) on Oracle
```

spec/server/archives/archives_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+  it_behaves_like "a standard Linux machine"
+
+  describe service('apache2') do
+    it { should be_enabled }
+    it { should be_running }
+  end
+
+  describe port(80) do
+    it { should be_listening }
+  end
```
MarkEWaite, line 13: Should port 443 be checked in the same way?

dist/role/manifests/archives.pp
olblak: Good point

dist/role/manifests/archives.pp
```diff
@@ -0,0 +1,6 @@
+#
+# archives is a tiny VM (2vCPU/8GB RAM) on Oracle
```
dduportal:
```suggestion
# archives requires low resources (2vCPU/8GB RAM)
```

dist/profile/manifests/azure.pp
```diff
@@ -5,15 +5,11 @@
 class profile::azure (
   $cli = true,
 ) {
+  # azure-cli only works on amd64
```
dduportal:
```suggestion
```

dist/profile/manifests/azure.pp
```diff
@@ -5,15 +5,11 @@
...
+  include apt
 
-  if $cli {
-    ensure_packages(['python-pip'])
+  if ($cli == true) and ($facts['architecture'] == 'amd64') {
```
dduportal, line 11:
```suggestion
  # azure-cli only works on amd64
  if ($cli == true) and ($facts['architecture'] == 'amd64') {
```

hieradata/common.yaml
```diff
@@ -98,6 +98,10 @@ accounts:
 # Lets Encrypt settings
 letsencrypt::config::email: "tyler@monkeypox.org"
 letsencrypt::config::server: "https://acme-staging.api.letsencrypt.org/directory"
+letsencrypt::renew_cron_ensure: 'present'
+letsencrypt::renew_cron_minute: 0
+letsencrypt::renew_cron_hour: 6
+letsencrypt::renew_cron_monthday: '1-31/2'
```
dduportal:
```suggestion
```

dduportal: (default value is `*` which mean everyday)

PR #1805, opened 2021-07-22
Update archives rsyncd allowed list

olblak:
Add the public IP of the new archives machine deployed on oracle cloud to allow connection to the archives rsynd service


dduportal (approved):
🚀

PR #1807, opened 2021-07-22
[INFRA-2917] feat: manage jenkins agent with JCasC

dduportal:
This PR aims at managing ci.jenkins.io and trusted.jenkins.io's agents are managed using JCasC.

Depends on #1810 

It also improves the `idempotent-cli` of Jenkins VMs to use the `jenkins-plugin-cli`  when installing plugins (but keep using `jenkins-cli.jar` for groovy and safe-restart cases)


dist/profile/files/azure.ci.jenkins.io/agents.yaml
```diff
@@ -0,0 +1,4 @@
+---
+jenkins:
+  systemMessage: "Jenkins configured automatically by Jenkins Configuration as Code plugin\n\n"
+...
```
timja:
```suggestion

```

dist/profile/manifests/buildmaster.pp
```diff
@@ -239,24 +240,26 @@
     }
 
     # Define Casc directory through java opts to avoid conditional environment variable
-    $java_opts = "${java_opts} -Dcasc.jenkins.config=${jcasc_config_dir}"
+    $jcasc_java_opts = " -Dcasc.jenkins.config=${container_jenkins_home}/${jcasc_config_dir}"
```
timja, line 243: improvement for later:
provide an auto reload token as well, and either a file watcher or call the webhook on change

timja (dismissed):

dist/profile/files/azure.ci.jenkins.io/agents.yaml
```diff
@@ -0,0 +1,501 @@
...
+        diskType: "managed"
+        doNotUseMachineIfInitFails: true
+        enableMSI: false
+        enableUAMI: false
+        ephemeralOSDisk: false
+        executeInitScriptAsRoot: true
+        existingStorageAccountName: "jnl4vmebchyhh6cmbd4dfvpa"
+        imageReference:
+          galleryImageDefinition: "jenkins-agent-ubuntu-20"
+          galleryImageVersion: "0.2.2"
```
timja: needs to be updated in multiple places and AMIs updated
```suggestion
          galleryImageVersion: "0.2.4"
```

dist/profile/templates/azure.ci.jenkins.io/agents.yaml.erb
```diff
@@ -0,0 +1,497 @@
...
+        credentialsId: "jenkinsvmagents-userpass"
+        diskType: "managed"
+        doNotUseMachineIfInitFails: false
+        enableMSI: false
+        enableUAMI: false
+        ephemeralOSDisk: false
+        executeInitScriptAsRoot: false
+        existingStorageAccountName: "<%= @agents_azure_vms_storage_account_windows %>"
+        imageReference:
+          id: "/subscriptions/<%= @agents_azure_vms_gallery_subscription_id %>/resourceGroups/prod-packer-images-eastus/providers/Microsoft.Compute/images/jenkins-agent-win10"
```
timja, line 166: where does this image come from? I can't see it in packer, the reason I ask is that I was wondering why this still used a managed image instead of image gallery

dduportal: I have no idea where the Windows 10 image comes from, neither where it's used :| I would vote to remove it, but it might break something. Asked in IRC

olblak (approved):

dist/profile/manifests/buildmaster.pp
dduportal: https://github.com/jenkins-infra/jenkins-infra/pull/1817

dist/profile/templates/azure.ci.jenkins.io/agents.yaml.erb
dduportal: cc @MarkEWaite @oleg-nenashev @olblak @jglick : do you know which job are expected to use the Windows 10 agents or if we can "just" remove this in favor of Windows 2019 ?

dist/profile/templates/azure.ci.jenkins.io/agents.yaml.erb
```diff
@@ -0,0 +1,502 @@
+---
+jenkins:
+  ## No builds on master
+  numExecutors: 0
+  agentProtocols:
+  - "JNLP4-connect"
+  - "Ping"
```
jglick, line 7: This is the default IIRC, you should not need to override it.

dist/profile/templates/azure.ci.jenkins.io/agents.yaml.erb
jglick: I had not heard of Windows 10 agents. For CI I would certainly expect to be using the latest available version of Windows Server, either LTSC (2019) or SAC (whatever).

dist/profile/templates/azure.ci.jenkins.io/agents.yaml.erb
dduportal: I've put them here for the sake of knowledge sharing, to make sure that the reader/contributor knows it

dist/profile/templates/azure.ci.jenkins.io/agents.yaml.erb
dduportal: Thanks for your answers and your time @jglick !

PR #1810, opened 2021-07-27
Puppet code cleanup: chore and fixes

dduportal:
While working on #1807, I stumbled upon a lot of minor issues which scope is unrelated to JCasc.

The goal of this PR is to contributes theses issues as soon as posible:

- (major) Fix path issues for the jenkins controllers, where the path used for the default user in the container is the path of the Jenkins Home on the host (while the Jenkins containers are using `/var/jenkins-home` as path). ⚠️ this might have an impact on the user behavior in production
- (major) Ensure that all the required plugins are specified through hieradata
- (medium) Fix the Jenkins' `idempotent_cli` not working by updating the script to use the correct `CLI` jar file in the container
- (medium) fix an issue on the `docker` role which throw an error about `/etc/docker` not existing when running an initial puppet agent run on a scratch machine + missing unit tests for this
- (minor) allow contributors to run vagrant on non admin VMs
- (minor) fully support @olblak 's work to have Ubuntu 20.04 compliance, by fully supporting it on the Vagrant tooling
- (minor) Markdown linting of  `README`

### DIFF
--- a/dist/profile/manifests/jenkinsplugin.pp
+++ b/dist/profile/manifests/jenkinsplugin.pp
@@ -7,12 +7,14 @@ define profile::jenkinsplugin (
   validate_string($name)
 
   exec { "install-plugin-${name}":
+    ## Check for plugin presence on the HOST (e.g. with the jenkins home in "/var/lib/jenkins" on the filesystem)
+    unless    => "/usr/bin/test -f /var/lib/jenkins/plugins/${name}.jpi || /usr/bin/test -f /var/lib/jenkins/plugins/${name}.hpi",
+    require   => Docker::Run['jenkins'],
+    ## Install the plugin (if needed) in the container, e.g. with the jenkins home mounted in /var/jenkins_home
     command   => "/usr/share/jenkins/idempotent-cli install-plugin ${name}",
     tries     => 10,
     try_sleep => 10,
     path      => ['/bin', '/usr/bin'],
-    unless    => "/usr/bin/test -f /var/lib/jenkins/plugins/${name}.jpi",
-    require   => Docker::Run['jenkins'],
     notify    => Exec['safe-restart-jenkins-via-ssh-cli'],
   }
 }
--- a/README.md
+++ b/README.md
@@ -41,21 +41,16 @@ This repo just manages and configures the deployments.
 
 The amount of testing that can be done locally is as follows:
 
- * `bundle install` - To get the necessary gems to run tests locally, if you're
-   unfamiliar with Ruby development you may want to use [RVM](http://rvm.io/)
-   to create an isolated Ruby environment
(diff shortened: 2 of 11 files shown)


dist/profile/manifests/jenkinsplugin.pp
```diff
@@ -7,12 +7,14 @@
   validate_string($name)
 
   exec { "install-plugin-${name}":
+    ## Check for plugin presence on the HOST (e.g. with the jenkins home in "/var/lib/jenkins" on the filesystem)
+    unless    => "/usr/bin/test -f /var/lib/jenkins/plugins/${name}.jpi || /usr/bin/test -f /var/lib/jenkins/plugins/${name}.hpi",
+    require   => Docker::Run['jenkins'],
```
olblak, line 12: Can you install a plugin is Jenkins is still booting? 
I think we may face a racing condition where the container is started but the Jenkins instance inside is not ready yet.
I am wondering if we shouldn't have an additional check to be sure that Jenkins is ready

olblak (approved):
Great improvement, despite my minor comment, I think this or is ready to be ship

timja (approved):
i'm sure there's more plugins that need adding to the list for it to be fully managed but looks fine

dist/profile/manifests/jenkinsplugin.pp
dduportal: Good observation, Let me add a comment to explain it for later readers.

In the current state, the plugin will always be able to be installed unless the container is killed:

- There are no healthchecks defined, so the status "container running" is reached as soon as the process with the PID 1 is up
- The PID 1 is [`tini`](https://github.com/jenkinsci/docker/blob/master/11/debian/buster/hotspot/Dockerfile#L105) which means that even if Jenkins is not able to start successfully (example: JCasc issues, HTTP/500, etc.) , there should be always a java process
- If you run a `docker exec` command (such as this one), then `tini` manages all the sub processes: while the plugin installation is running, then tini should not exit (unless a kill from systemd or an host admin of course)

However you are correct that if we reach a status where Jenkins is unable to expose its API to localhost (not the case with JCasc failing, but the startup sequence is not an easy one and can have edge cases), then the plugin installation would fail.

I got a proposal (but subsequent PR) that would be to edit the `idempotent-cli` script (on the host) to detect the passed command: if it is `install-plugin <...>` then it should use the [`jenkins-plugin-cli`](https://github.com/jenkinsci/docker#plugin-installation-manager-cli) in the container with the argument `--plugins-directory=/var/jenkins-home/plugins`], so there would be no dependency to Jenkins's status for the plugins installation (but we have to keep the `jenkins-cli.jar` for running groovy and triggering a `safe-restart` though). WDYT?

PR #1812, opened 2021-07-28
Add user mirrorsync to archives

olblak:
The purpose of this user is to allow ssh connection from a third location
which can trigger a mirror synchronization by running mirrorsync on demand.
Remark the command `mirrosync` command is still execute twice per hour.
So it's an additional improvement to the idea of having mirror operating by default in pull mode with the ability to trigger a sync when needed such as when we do a release.

At the moment this [script](https://github.com/jenkins-infra/mirror-scripts/blob/master/populate-archives.sh) is executed from pkg.origin.jenkins.io which upload artifacts from pkg.origing.jenkins.io to archives.jenkins.io.
The goal for this PR is to create a new user on archives.jenkins.io who can only trigger mirror synchronization with a command similar to `ssh jenkins@ftp-osl.osuosl.org 'sh trigger-jenkins'` from [here](https://github.com/jenkins-infra/mirror-scripts/blob/master/sync-recent-releases.sh#L40).

Ps: Once this PR is merged, I'll update this [script](https://github.com/jenkins-infra/mirror-scripts/blob/master/populate-archives.sh) by changing from `exec rsync -avz /srv/releases/jenkins/ www-data@archives.jenkins-ci.org:/srv/releases/` to `ssh mirrorsync@archives.jenkins.io 'sh mirrorsync'`  


timja (dismissed):

dist/profile/manifests/archives.pp
```diff
@@ -2,18 +2,72 @@
...
   include profile::apachemisc
   include profile::letsencrypt
 
+  $apache_owner     = 'www-data'
+  $apache_group     = $apache_owner
+
+  ## Manage mirrorsync user and its home directory
+  user { 'mirrorsync':
+    ensure     => present,
+    shell      => '/bin/bash',
```
dduportal, line 22: Is `bash` mandatory for a rsync usage (e.g. instead of `/bin/nologin` ?)
If yes, could you add a comment explaining "why" it is mandatory?

dist/profile/manifests/archives.pp
```diff
@@ -2,18 +2,72 @@
...
+    managehome => true,
+  }
+
+  # The user mirrorsync is only used to trigger a synchronization
+  # between a remote a mirror and the directory as the user www-data
+  sudo::conf { 'mirrorsync':
+    ensure  => present,
+    content => 'mirrorsync ALL=(ALL) NOPASSWD: /usr/bin/mirrorsync',
```
dduportal, line 30:
```suggestion
    content => 'mirrorsync ALL=(ALL) NOPASSWD: /usr/bin/mirrorsync',
    require => User['mirrorsync'],
```

dduportal: (to avoid failing the sudo config validation if user has not been created yet)

dist/profile/manifests/archives.pp
```diff
@@ -2,18 +2,72 @@
...
 
+  $apache_owner     = 'www-data'
+  $apache_group     = $apache_owner
+
+  ## Manage mirrorsync user and its home directory
+  user { 'mirrorsync':
+    ensure     => present,
+    shell      => '/bin/bash',
+    managehome => true,
+  }
```
dduportal, line 24: Should this user associated to the `www-data` (`$apache_group`) group? Or the other way: `www-data` should be added to the group `mirrorsync` ?

olblak: I don't think the user should be associated with "www-data", the only think that the user "mirrorsync" should do is being allowed to run the script "mirrorsync" which then run a rsync command as the user www-data

dist/profile/manifests/archives.pp
olblak: Good comment, let me think about that

olblak: I thought we couldn't run the command mirrorsync via ssh if we can't login

olblak:
I just changed the ownership of the script `mirrorsync` from the user `mirrorsync` to `root` to be sure that the user `mirrorsync` can't modify its content

dduportal (approved):
good job!

PR #1821, opened 2021-08-12
feat: add maven and maven-11 kubernetes agent templates to ci.jenkins.io [INFRA-2918]

dduportal:
As per INFRA-2918, this PR introduces the following changes:

- Add maven and maven-11 kubernetes agent templates to ci.jenkins.io
- The local vagrant stack is using the same JCasC `agent.yml` configuration as ci.jenkins.io to provide manual local validation
- Enforces resources limits + requests on kubernetes templates:
  * ruby, maven and maven-jdk11 are using the same CPU + memory as their ACI counterparts
  * the default JNLP agent is restricted to 1CPU + 1G


timja (approved):
let's see how it goes, the reason for the ACI value we had was that was the limit for how much you can get.

PR #1829, opened 2021-08-16
fix: Set resources to 4 CPUs and 8 Gb per Kubernetes agent on ci.jenkins.io

dduportal:
This PR comes from https://github.com/jenkinsci/jenkins/pull/5667 (see. https://issues.jenkins.io/browse/INFRA-2918 for context on the Kubernetes agents).

It introduces the following new (empirical) changes:

* Set the resources of each pod template to 4 vCPUs and 8Gb to avoid timeouting on the Jenkins Core jobs
* Update the maximum pod limit accordingly (as we have 50 worker nodes max. See inline comment for the formula)
* Cleanup Puppet code remnants of previous PRs in this area

[DDU] Setting to draft as local vagrant test are currently being run


dduportal:
Ready to review: validated on a manual local vagrant (with `vagrant up --provision jenkins::master`): Jenkins starts (so now CasC syntax/semantic error), and the expected configuration is ✅ on the Cloud Config UI as well as local Casc export.

timja (approved):

olblak (approved):
I have concerned that pods have too much memory assigned to them so we may struggle to find nodes with enough capacity. It may be interested to specify in a comment the size of k8s nodes that you are expecting

dduportal:
@olblak as underlined in the comment, the worker node are currently 8 CPUs / 32 Gb of memory. It means that memory won't be the limit: instead it's the CPU that is limiting.
Since both limit and requests are set to the same values, the scheduler will never put more than 2 pods per node, e.g. using 2*4 = 8 vCPUs and 2*8 = 16 Gb: we're going to waste 16 Gb of memory, but pack the CPU usage.

The CPU superallocation is not a worry in itself: we are using `t3.` family instance which allows burst on the CPU on the host (thanks to the CPU credit system).

Since autoscaling is maxed to 50 VMs, we should be able to schedule 100 pods maximum. Since this is defined in the config, then Jenkins will never try to allocate more than 100 agents: the rest will wait patiently in the build queue (which is a nice resilient queue :) )

However you are correct: we might want to improve the sizing:
* Increase the size of the 2-3 instances that are always up to handle more containers
* Define at least another node pool on "spot instances" to ensure cost are decreased
* Maybe define different node pools such as "big" and "small" to pack more small pods on smaller nodes, while the big guns are reserved for BOM/Core PRs (like we do for the High Mem VMs)

PR #1830, opened 2021-08-16
Fix: get rid of Jenkins CLI usages [INFRA-3037]

dduportal:
This PR aims to fix the `reload-jcasc-configuration` errors seen in production and described in INFRA-3037.

Important: this PR might be breaking. Please wait for merging before we all agree and approve.

* Add a custom CasC setup for the local vagrant setup, to catch the current error (TL;DR; requires an authentication)
* Ensure that JDK11 is the default (to ensure that the java_opts flags are not broken depending on the instance)
* Do not setup an SSHD server for the Jenkins controller
  * We know that it was used for Jenkins CLI usage, but it was not working anymore (user UID mismatch, exposed SSH key, require configuration flag and the technical user `jenkins` was not an admin)
  * There **might** be other use case outside this repo: if it is broken, we have to fix and document!
  * Please note that the SSH key has not been put as "removed" in the Puppet code: a manual remove will be required in 1 week (to be sure that we can rollback quickly if needed in next days)
* Full removal of the script `idempotent-cli`
  * Jenkins CasC reload is now implemented wit the REST API with a token passed as encrypted data through hieradata (at least for ci.jenkins.io)
  * The operation `safe-restart` is implemented as a container restart (con: can cut connections, pros: faster - yes - and ensure full coherency on the plugin states: Azure-* and EC2, I'm looking at you)
  * The operation install plugin was already a docker exec: the command is now in the puppet code
* Removed the "create CLI user" logic and code
  * The user `jenkins` will have to be deletd manually in 1 week


timja (dismissed):

timja (dismissed):

dist/profile/manifests/buildmaster.pp
```diff
@@ -33,10 +34,21 @@
   $groovy_d_lock_down_jenkins      = 'absent',
   $groovy_d_terraform_credentials  = 'absent',
   $jcasc_configs                   = [],
+  $jcasc_reload_token                     = '',
```
timja: indenting looks off here, I won't use a suggestion as i'll probably get the spaces wrong

dduportal: Sounds like it's my IDE's Puppet formatter which is confused by the comment on the next line. Weird. Fixed, thanks!

timja (approved):

dduportal:
I'm on call, so let's go for merging!

PR #1833, opened 2021-08-18
Fix: [ci.jenkins.io] Define Kubernetes agents imagePullPolicy to 'IfNotPresent' [INFRA-3036]

dduportal:
This PR is related to INFRA-3036.

It's goal is to limit the amount of Kubernetes Agents for ci.jenkins.io to stay in state "ErrImagePullBackOff".

By setting up the image pull policy to `IfNotPresent`, it ensures that the pod images are only pulled one time on the worker nodes.

The consequence is that a new Docker image published under the tag `latest` won't be used unless on a new worker node.
This issue will be fixed in a subsequent PRs, but there is a commit here to lay the foundation by moving ALL the kubernetes pod templates into Hieradata.


timja (dismissed):

olblak (dismissed):
:shipit:

timja (approved):

PR #1840, opened 2021-08-25
Updatecli check that docker image tag exist

olblak:
Ensure that updatecli test that the new docker image exist before updating the hieradata

### Signed-off-by: Olivier Vernin <olivier@vernin.me>


dduportal (approved):
🚀

timja (approved):

PR #1848, opened 2021-08-31
feat: [ci.jenkins.io] Switch to Adoptium OpenJDK

dduportal:
This PR introduces the (many) following changes:

* chore: Improve templating for ci.jenkins.io Casc agent setup
* feat: define Jenkins tools installation with Casc for ci.jenkins.io
* Bump all tools installation to use Adoptium (versions bumps for JDK8 and 11)
* Bumpd the VM templates used on ci.jenkins.io to latest builds (to use Adoptium) - ref. https://github.com/jenkins-infra/packer-images/pull/78

PR #1849, opened 2021-09-01
fix: [ci.jenkins.io] remove unsupported JDK8 tools installations

dduportal:
After #1848 was merged and deployed, the ci.jenkins.io UI show the errror and warning below.

Capture d’écran 2021-09-01 à 08 36 56

Capture d’écran 2021-09-01 à 08 37 01

This PR removes the tools installations that are triggering this 2 messages:

- The `windowspacker` agent does not exist anymore
- OpenJDK8 is not provided for s390x by Adoptium project *except* for the Docker image: https://hub.docker.com/_/eclipse-temurin/

THis PR also fixes the configuration errors in the JDK11 tool configuration that were introduded in #1848

### DIFF
--- a/dist/profile/templates/azure.ci.jenkins.io/tools.yaml.erb
+++ b/dist/profile/templates/azure.ci.jenkins.io/tools.yaml.erb
@@ -38,23 +38,23 @@ tool:
           installers:
           - zip:
               label: "linux"
-              subdir: "jdk-<%= @tools["jdk8"]["version"] %>"
+              subdir: "jdk-<%= @tools["jdk11"]["version"] %>"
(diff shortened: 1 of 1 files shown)


timja (dismissed):

timja (dismissed):

timja (approved):

MarkEWaite:
If we need to build on s390x with Java 8, we'll need OpenJ9 runtimes from the [IBM Semeru project](https://developer.ibm.com/languages/java/semeru-runtimes/downloads#).   The hotspot JVM is not available on s390x with Java 8.  I think it is best if we can avoid building Java 8 on s390x and limit ourselves purely to build with Java 11.

PR #1865, opened 2021-09-17
chore: enable automatic update of AMI for EC2 agents on all Jenkins controllers

dduportal:
This PR enables automatic updates of the AMI references for EC2 agents on the infrastructure.
Please note that only ci.jenkins/io and the local vagrant are using it now, but it's available for the other Jenkins controller through the common hieradata.

- Defines the 3 AMIs (Ubuntu Intel, Ubuntu ARM and Windows Intel) in hieradata commons
  - Implies an update on the Casc agent template for ci.jenkins.io to use these commons AMIs
  - minor: A new tag is defined on the Ubuntu Linux agent: `ubuntu` (due to template update)

- Use the "shell" resource introduced in [`updatecli` 0.7.0](https://github.com/updatecli/updatecli/releases/tag/v0.7.0) to retrieve the AMI IDs
  - Requires an update on the Github Actions: cannot use the default `updatecli` action does not allow installing the `awscli` (non root user, no unzip). This should be rollbacked once https://github.com/updatecli/updatecli/pull/265 is ready
  - The shell script used as a special dummy case: when running `updatecli diff` (e.g. `DRY_RUN=true`), it never fails if the aws CLI is not configured, because GH action does not provides the secret for security reasons. This "hack" allows to keep checking the rest of updatecli and ensure we did not break anything on a given PR? without sacrificying the security

About the AWS credentials:
- A new IAM user had been defined and created in jenkins-infra/terraform-states, with a minimalistic IAM policy to only allow a minimum set of "read" actions such as "describe images" and "describe instances"
- The credentials of this user have been generated and added to the secrets of this repository (nothing in Jenkins)

There have been 4 validations for this PR:

- 1 validation is the usual Jenkins CI check that validates the Puppet code change
- 1 validation has been done manually: exeucting locally `vagrant up --provision jenkins::master` to ensure that the hieradata change did not break the CasC config
- 1 validation is the ususal GitHub workflow for updatecli: it works (proving that updatecli configuration is ok) but reports a dummy value for the AMI sources (as expected)
- 1 additional validation: a workflow triggered through an upstream branch, to validate that the credentials are correctly set up and reports correct AMI values
  - Ex. https://github.com/jenkins-infra/jenkins-infra/runs/3633008996?check_suite_focus=true#step:5:478


dduportal:
The GitHub Action Check has been validated using a branch here: https://github.com/jenkins-infra/jenkins-infra/runs/3632630030?check_suite_focus=true (branch deleted), in order to allow the AWS credentials secretsto be available

.github/workflows/updatecli.yaml
```diff
@@ -14,19 +14,35 @@ jobs:
...
+        if: github.event_name != 'pull_request'
+        uses: aws-actions/configure-aws-credentials@v1
         with:
-          command: diff
-          flags: "--config ./updatecli/updatecli.d --values ./updatecli/values.github-action.yaml"
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-2
+      - name: Diff
+        # uses: updatecli/updatecli-action@v1
```
timja, line 31: why all the commented out code?

dduportal: The updatecli default action (default docker image) does not allow to install `aws` CLI. Until either the GH action is updated OR https://github.com/updatecli/updatecli/pull/265 is implemented, then we have to use the `run: ` instruction to define custom shell commands on the worker (instead of relying on the default Docker command for updatecli)

Commented code is kept to mark that, but I don't mind removing it also :)

timja (approved):

PR #1897, opened 2021-09-29
fix [ci.jenkins.io] ensure that the azure VM template for windows 2019 can be provisionned

dduportal:
[New PR message]
This PR disable the ephemeral disk for the Azure VM template for win-2019 on ci.jenkins.io, until the packer image is shrunk below 100 Gb.

[Original PR message]
* With the current setup on ci.jenkins.io (`Standard_DS4_v3`), deployment fails with this error:

```text
OS disk of Ephemeral VM with size greater than 100 GB is not allowed for VM size Standard_D4s_v3 when the DiffDiskPlacement is CacheDisk.
```

Because the VM image (`0.5.0`) is defining a size of 130 Gb for the system disk (https://github.com/jenkins-infra/packer-images/blob/0.5.0/jenkins-agent.pkr.hcl#L244), which is more than the 100 Gb specified for this instance size as per https://docs.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series#dsv3-series.

* If forcing disk size less than 100 Gb, then the error changes to the following:

```text
The specified disk size 90 GB is smaller than the size of the corresponding disk in the VM image: 130 GB. This is not allowed. Please choose equal or greater size or do not specify an explicit size.
```

* So this Pull Request switches the size instance to DS8_v3 (200 Gb of caching instead of 100 Gb of cache) to fix the allocation issue for now

### DIFF
--- a/dist/profile/templates/buildmaster/casc/clouds.yaml.erb
+++ b/dist/profile/templates/buildmaster/casc/clouds.yaml.erb
@@ -19,7 +19,7 @@ jenkins:
         doNotUseMachineIfInitFails: true
         enableMSI: false
         enableUAMI: false
-        ephemeralOSDisk: true
+        ephemeralOSDisk: <%= agent["useEphemeralOSDisk"].nil? ? true : agent["useEphemeralOSDisk"] %>
         existingStorageAccountName: "<%= agent["storageAccount"] %>"
         storageAccountNameReferenceType: "existing"
         storageAccountType: "Standard_LRS"
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -174,6 +174,7 @@ profile::buildmaster::cloud_agents:
         maxInstances: 10
         useAsMuchAsPosible: true
         credentialsId: "jenkinsvmagents-userpass"
+        useEphemeralOSDisk: false
   azure-container-agents:
     aci-windows:
       credentialsId: "azure-credentials"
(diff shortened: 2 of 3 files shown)


timja (dismissed):
other options would be to set the osDisk (check exact name but its for the disk size) field or disable ephemeral disk for windows?

dduportal:
>other options would be to set the osDisk (check exact name but its for the disk size) field or disable ephemeral disk for windows?

@timja Good point!
The use of ephemeral disk is really a big kick of startup performance (~2 min difference in startup time).
But since this PR increases the cost of this workload by ~2x (on the hourly cost) and that Windows instance are billed hourly, it might be an option to consider.

- From https://docs.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series, I'm not sure if we pay the Azure storage of the 130 Gb disk when it's not ephemeral: could you help me on this?
- I was also planning to update the packer-image template to decrease the image disk size below 100 Gb and to decrease back to DS4_v3. WDYT?

timja:
From what I can tell a 128gb disk is $18 a month so likely very little with an ephemeral workload

> I was also planning to update the packer-image template to decrease the image disk size below 100 Gb and to decrease back to DS4_v3. WDYT?

Sure

dduportal:
OK, so I'm gonna disable the ephemeral disk for windows to avoid increasing the VM size right to fix the template right now, and then we'll enable it back once the size will be decreased on the image.

Thanks @timja

dduportal:
PR updated @timja , tested locally on Vagrant.

timja (approved):

PR #1903, opened 2021-10-01
Fix for ci.jenkins.io agents (more kubernetes agents, fix maven-11-windows)

dduportal:

- Fix the maven-11 windows ACI agent not allocated on ci.jenkins.io (typo: missing `s`)
- Adjust the Kubernetes agent size and capping on ci.jenkins.io to handle increased load after
  - Follows up https://github.com/jenkins-infra/aws/pull/31
  - New container cap is 150 (3 maven-jdk pods per worker max)
  - Please note that the non maven-jdk pods have a smaller size to "fill" the gap. As there is no container cap per template, this is only to avoid triggerring too much workers, but does not change anything when the 150 pods limit is reached
- cleanup chore: removing unused hieradata keys (container images definitions have moved to `hieradata/common.yaml`)

### DIFF
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -53,7 +53,6 @@ profile::buildmaster::cloud_agents:
           cpus: 4
           memory: 8
         - name: jnlp-maven-11
-          image: jenkinsciinfra/inbound-agent-maven@sha256:0ddce98de47930b7505418421b89cbbfa5378d1c8f9c01bee416ff8ffe1d0775
           labels:
             - maven-11
             - jdk11
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -192,7 +192,7 @@ profile::buildmaster::cloud_agents:
           cpus: 4
           memory: 8
           labels:
-            - maven-11-window
+            - maven-11-windows
 # These are plugins we need in our ci environment
 profile::buildmaster::plugins:
   - ansicolor
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -37,12 +37,12 @@ profile::buildmaster::cloud_agents:
   kubernetes:
     cik8s:
       credentialsId: "cik8s-kubeconfig"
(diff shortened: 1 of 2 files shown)


hieradata/clients/azure.ci.jenkins.io.yaml
```diff
@@ -53,15 +53,14 @@ profile::buildmaster::cloud_agents:
           cpus: 4
           memory: 8
         - name: jnlp-maven-11
-          image: jenkinsciinfra/inbound-agent-maven@sha256:0ddce98de47930b7505418421b89cbbfa5378d1c8f9c01bee416ff8ffe1d0775
```
MarkEWaite, line 56: I'm surprised to see the deletion of an image specification.  Is the image then selected purely by name?  Is that enough?

hieradata/clients/azure.ci.jenkins.io.yaml
```diff
@@ -53,15 +53,14 @@ profile::buildmaster::cloud_agents:
...
           labels:
             - maven-11
             - jdk11
           cpus: 4
           memory: 8
         - name: jnlp-node
-          cpus: 4
-          memory: 8
+          cpus: 2
+          memory: 4
```
MarkEWaite, line 63: I thought that we had known failures with Jenkins core the last time we reduced from 8 GB RAM to 4 GB RAM.  Don't we need to keep the maven-11 at the higher values?
```suggestion
          cpus: 4
          memory: 8
```

MarkEWaite (approved):
Approved but with questions about the decrease of the agent memory size and agent CPU count.  I thought that we had Jenkins core test failures the last time we reduced from 8 GB RAM to 4 GB RAM

hieradata/clients/azure.ci.jenkins.io.yaml
dduportal: the changed values are only for `jnlp-node` (frontend related pipelines) and `jnlp-ruby` (infra related pipelines).

You're right to check this: we should keep 4 cpus to maven workloads :)

hieradata/vagrant/common.yaml
```diff
@@ -157,7 +156,7 @@ profile::buildmaster::cloud_agents:
           cpus: 4
           memory: 8
           labels:
-            - maven-11-window
+            - maven-11-windows
```
timja, line 159: ah nice find

dduportal: 😅

PR #1905, opened 2021-10-01
 Fixes #3743

noviicee:
Redirect 'SCM Filter Branch PR Plugin' page from wiki to plugins site

See jenkins-infra/jenkins.io#3743


MarkEWaite (approved):
May not be necessary, but will depend on how we decide to handle redirects for content from wiki.jenkins.io.

Plugin has migrated its documentation to GitHub.

noviicee:
Thank you :)

PR #1907, opened 2021-10-03
Remove Wiki redirects in permalinks

oleg-nenashev:
Part of the Wiki cleanup follow-ups. All the links end up with error 404 at the moment. Some of security resources are affected too, e.g. the /advisories link

CC @halkeye @rtyler @olblak @daniel-beck @wadeck

### DIFF
--- a/dist/profile/manifests/catchall.pp
+++ b/dist/profile/manifests/catchall.pp
@@ -79,7 +79,7 @@ class profile::catchall(
   # used to advertise the project meeting
   Redirect /meetings/        https://www.jenkins.io/project/governance-meeting/
   # used from friends of Jenkins plugin to link to the thank you page
-  Redirect /friend        https://www.jenkins.io/donate
+  Redirect /friend        https://www.jenkins.io/donate/
   # used by Gradle JPI plugin to include fragment
   Redirect /gradle-jpi-plugin/latest    https://raw.github.com/jenkinsci/gradle-jpi-plugin/master/install
   # used when encouraging people to subscribe to security advisories
--- a/dist/profile/manifests/catchall.pp
+++ b/dist/profile/manifests/catchall.pp
@@ -69,34 +69,33 @@ class profile::catchall(
 
   # permalinks
   # - this one is referenced from 1.395.1 "sign post" release
-  Redirect /why            https://wiki.jenkins-ci.org/pages/viewpage.action?pageId=53608972
+  # TODO: Create a Jenkins Manifesto page?
+  Redirect /why            https://www.jenkins.io/
   # baked in the help file to create account on Oracle for JDK downloads
   Redirect /oracleAccountSignup    http://www.oracle.com/webapps/redirect/signon?nexturl=http://jenkins-ci.org/
-  # to the donation page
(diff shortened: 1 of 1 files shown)


timja (dismissed):

dduportal (dismissed):
LGTM ❤️

dist/profile/manifests/catchall.pp
```diff
@@ -69,34 +69,33 @@
 
   # permalinks
   # - this one is referenced from 1.395.1 "sign post" release
-  Redirect /why            https://wiki.jenkins-ci.org/pages/viewpage.action?pageId=53608972
+  # TODO: Create a Jenkins Manifesto page?
```
halkeye, line 72: I'm not a big fan of committing TODOS
http://web.archive.org/web/20200511124337/https://wiki.jenkins.io/pages/viewpage.action?pageId=53608972 <-- its been deleted for a couple years now
http://web.archive.org/web/20170821200700/https://wiki.jenkins.io/pages/viewpage.action?pageId=53608972 <-- even before that, it was pretty dated

I think drop the todo and just redirect but no hard blocker.

halkeye (dismissed):

daniel-beck (dismissed):
Reasonable stopgap fix.

Longer term, do we have access logs that tell us which of these are even used?

These should also just be redirect filed in jenkins-infra/jenkins.io, no need to have them here.
Would also prevent weird discrepancies like jenkins.io/license/ vs. jenkins.io/license

dist/profile/manifests/catchall.pp
```diff
@@ -69,34 +69,33 @@
...
-  Redirect /license        https://wiki.jenkins-ci.org/display/JENKINS/Governance+Document#GovernanceDocument-cla
-  Redirect /licenses        https://wiki.jenkins-ci.org/display/JENKINS/Governance+Document#GovernanceDocument-cla
+  Redirect /license        https://www.jenkins.io/project/governance/#cla
+  Redirect /licenses        https://www.jenkins.io/project/governance/#cla
   # used to advertise the project meeting
-  Redirect /meetings/        https://wiki.jenkins-ci.org/display/JENKINS/Governance+Meeting+Agenda
+  Redirect /meetings/        https://www.jenkins.io/project/governance-meeting/
   # used from friends of Jenkins plugin to link to the thank you page
-  Redirect /friend        https://wiki.jenkins-ci.org/display/JENKINS/Donation
+  Redirect /friend        https://www.jenkins.io/donate
```
daniel-beck:
```suggestion
  Redirect /friend        https://www.jenkins.io/donate/
```

dduportal: oh, good catch! the link `https://www.jenkins.io/donate` redirects to `https://wiki.jenkins-ci.org/display/JENKINS/Donation`, while `https://www.jenkins.io/donate/` is ok.

daniel-beck: Not an actual catch because https://github.com/jenkins-infra/jenkins-infra/pull/1907/files#diff-2b2dd788d03cea552c387130532e5f5b364b5617421e6c240f0c2fbdb45f3fbaL75-L76 would take care of that. This just saves an extra redirect.

dduportal (approved):

PR #1919, opened 2021-10-06
Revert "feat [buildmaster] define static paths for jdk tools"

MarkEWaite:
## Restore JDK tool installer temporarily

Static JDK tool location is not available on every agent type.  We need the tool installer to be available until the static JDK tool location is available on all agent types.

See

* https://ci.jenkins.io/job/Plugins/job/elastic-axis-plugin/job/PR-46/3/console

This reverts commit f841cdfda5982354c233c5a4a475f1b797965847.


dduportal (changes requested):
Release (+fix ) in progress. Reverting the change won't change the behavior though (this is the packer-image version changed to `0.6.0`)

PR #1932, opened 2021-10-14
Upcoming breaking change in azure vm agents plugin

timja:
https://github.com/jenkinsci/azure-vm-agents-plugin/pull/313

Once the plugin has been upgraded this needs to be merged


dduportal (approved):
(puppet agent is disabled: I'll update the plugins before enabling it again) after https://github.com/jenkins-infra/docker-inbound-agents/pull/8 and thge docker image update here

dduportal:
Applied successfully on both trusted and ci.j, thanks TIm! It seems that it also fixed the "need to save the configure cloud form on the UI to make it work" \o/

timja:
I did a bunch of cleanup in that PR so optional fields shouldn’t be needed to be set anymore, I guess it was one of those that was missing before, this file could probably be cleaned up a bit now too

PR #1936, opened 2021-10-19
fix: add 'docker-windows' label to hieradata templates (INFRA-3099)

lemeurherve:
Add a new 'docker-windows' label, which will replace the 'windock' one. (To be coherent with the 'docker-highmem' label)
See https://issues.jenkins.io/browse/INFRA-3099?focusedCommentId=414936


dduportal (approved):
🚀

PR #1946, opened 2021-10-27
fix(azure.ci.jenkins.io) ensure that s390x and ppc64 agents are using the same Maven version as the other agents

dduportal:
As for today, the 2 static agents `s390x` and `ppc64le` on ci.jenkins.io are setup to use their local Maven installation.

The issue is that we do not track these installation, resutling in no update: it's still an olde `3.6.3` version.

This PR switches from local installation to the generic maven tool installation which is automatically tracked and updated.


dduportal:
Ping @MarkEWaite @lemeurherve @olblak @timja on this one:

If this PR is applied, then there would be NO command `mvn` available for builds running on these 2 specific agents, **unless** the pipeline explicitly call the the `tools 'maven'` instruction.

Do you know if we could enable the "global tool" by default for these agents (that would result on downloading Maven for each machine)?

=> The alternative would be to manage these 2 machines under Puppet and defines the Maven version through this (easy to upgrade since its already on hieradata/common.yaml). But I'm not sure about the PPC64 or s390x support of the puppet agent.

=> Last alternative is to access to these machines as admin and manually upgrade Maven. Meh.

WDYT?

MarkEWaite:
I've updated the static maven version in #2012 for both s390x and ppc64le.  I think you're right that we need a better way to handle those agents.  I prefer assuring that the `mvn` command is available.

dduportal:
Closing at this PR is not current the initial problem should be re-assessed.

PR #1958, opened 2021-11-02
Automatic Update of Docker Engine using updatecli

aneveux:
Hello :wave: 

https://issues.jenkins.io/browse/INFRA-3074

Please find a first answer to this issue in that Pull Request allowing to:

- Check for the latest version of docker-ce with apt,
- Set that version in the Puppet's provisioning file.

Please note that for now, this PR requires updatecli to be executed on:

- A Debian or Ubuntu environment,
- With apt,
- With docker-ce GPG keys and repositories configured,

As per now, there are no particular verification done on the Puppet's provisionning file.

Here are some future improvements that could be planned around this PR:

- Relying on a Docker image for resolving the latest docker-ce version (allowing to run the script on any environment basically),
- Checking for the Puppet's provisionning file to exist (I believe it's not yet possible through conditions, but could be a thing)
- Check for the existence of docker::version key in the Puppet's provisionning file (https://github.com/updatecli/updatecli/issues/289 needed)

Please don't hesitate to let me know what you think :)


updatecli/scripts/docker-ce-updates.sh
```diff
@@ -0,0 +1,24 @@
...
+# We want to get the Candidate version (which is the latest available)
+# And we want it in a readable format
+# --
+# DEV MEMO: Line explanation:
+# 1. Retrieve information about docker-ce from apt
+# 2. Keep only the line about the Candidate version (latest available)
+# 3. Cut it so we only keep the version and remove title (version contains a :, hence keeping fields 2 and 3)
+# 4. Trimming the result (removing spaces before and after)
+# 5. Surrounding the result with simple quotes ('') as it'll be needed this way
+apt-cache policy docker-ce | grep 'Candidate' | cut -f2,3 -d':' | xargs | { read -r x ; echo "'${x}'"; }
```
dduportal:
```suggestion
apt-cache policy docker-ce `# 1. Retrieve information about docker-ce from apt` \
  | grep 'Candidate' `# 2. Keep only the line about the Candidate version (latest available)` \
  | cut -f2,3 -d':' `# 3. Cut it so we only keep the version and remove title (version contains a :, hence keeping fields 2 and 3)` \
  | xargs `# 4. Trimming the result (removing spaces before and after)` \
  | { read -r x ; echo "'${x}'"; } # 5. Surrounding the result with simple quotes ('') as it'll be needed this way
```

Proposal (e.g. nitpicking, not blocking and subject to yes or no answer) to provide comment per-instruction. it's a pattern I've learned a few years ago to help streamlining the cognitive load when reading comment on a complex shell pipeline set of instructions. 
It uses backticks to hold the comments (except for the last one of course).

WDYT?

updatecli/updatecli.d/docker-ce.yaml
```diff
@@ -0,0 +1,15 @@
...
+    kind: shell
+    spec:
+      command: bash ./updatecli/scripts/docker-ce-updates.sh
+targets:
+  default:
+    name: Update docker-ce version in common.yaml
+    kind: yaml
+    spec:
+      file: hieradata/common.yaml
+      key: "docker::version"
```
dduportal:
```suggestion
      key: "docker::version"

```
Nitpick: As per POSIX, all lines should end with an "endline" character (https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline -> https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)

updatecli/scripts/docker-ce-updates.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# This script uses apt to find the latest version of docker-ce available.
+set -eu -o pipefail
+
+# Upadting the list of latest updates available for installed packages on the system
```
dduportal:
```suggestion
# Updating the list of latest updates available for installed packages on the system
```

Typo :)

updatecli/scripts/docker-ce-updates.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# This script uses apt to find the latest version of docker-ce available.
+set -eu -o pipefail
+
```
dduportal, line 3:
```suggestion

for cli in apt-get apt-cache grep cut xargs
do
  if ! command -v $cli >/dev/null 2>&1
  then
    echo "ERROR: command line ${cli} required but not found. Exiting."
    exit 1
  fi
done

```

WDYT to check for required CLI (at least)?

dduportal (changes requested):
That is an excellent job! Thanks for working on this topic 🥳 

Please find 4 suggestions: 1 is strictly blocking (typo), 1 is a strong recommendation (check for CLIs), and 2 are nitpickings.

I've enabled the "trust and run": further commits should trigger a build automatically.

dduportal (changes requested):
Could you put the YAML file into `updatecli/weekly.d/` instead of `updatecli/updatecli.d/`? (the former is a remnant of what we had a few month ago).

This would allow the Github Workflow to include this file as part of its run (ref. https://github.com/jenkins-infra/jenkins-infra/blob/production/.github/workflows/updatecli.yaml#L30)

updatecli/scripts/docker-ce-updates.sh
olblak: That's a good idea but what is the meaning of the backticks here?

updatecli/updatecli.d/docker-ce.yaml
olblak: I agree

updatecli/scripts/docker-ce-updates.sh
olblak: It can only the script stronger +1

olblak (dismissed):
Nice pull request

updatecli/scripts/docker-ce-updates.sh
dduportal: @olblak the backticks allow to inline comments (adding comments outside a backslash are syntax error, and starting a comment will comment the next piped instruction, so backticks creates a subshell with only the comment so it won't spawn a child process and it's not overloading the reading)

aneveux: Nice! I didn't know about that, and I was actually sad I couldn't add inline comments after backslashes :) Thanks a lot!

dduportal (changes requested):
We should check for an empty value and fail if it is the case (in the shell script?), otherwise it will break the `hieradata/common.yaml` by setting the docker version to the string `'(none)'` :D 

- https://github.com/jenkins-infra/jenkins-infra/pull/1958/checks#step:5:392

aneveux:
> We should check for an empty value and fail if it is the case (in the shell script?), otherwise it will break the `hieradata/common.yaml` by setting the docker version to the string `'(none)'` :D
> 
>     * https://github.com/jenkins-infra/jenkins-infra/pull/1958/checks#step:5:392

That's a really good point :)

aneveux:
@dduportal I added a simple Dockerfile allowing to run the script in a container so it can be executed on any Docker environment instead of a properly configured Ubuntu machine.

Built a test image using something like: `docker build -t test/test-ubuntu-updatecli:2 . `

And execute it with: `docker run test/test-ubuntu-updatecli:2`

And it gives the expected result.

I have no idea how/where to build/store/run that image in the context of that project though, so don't hesitate to recommand anything :)

dduportal:
> I have no idea how/where to build/store/run that image in the context of that project though, so don't hesitate to recommand anything :)

I'm thinking about 2 hypothesis:

- As per https://github.com/hadolint/hadolint/wiki/DL3009, the apt package index should never be shipped in a Docker image. The goal is to avoid unexpected caching that would provide an outdated index.
- There is no constraint on the time/efficiency of the source's shell script here: wether it takes 1s or 1min or even 10 min, we do not care as it's executed once a week.

=> That leds me to doubt if a Docker image other than the stock `ubuntu:18.04` is expected: WDYT about migrating the shell code you've put in the `Dockerfile` into the shell script executed as part of the updatecli source? That would shift the dependencies of the script from "bash +  apt-get + apt-cache + execute in Ubuntu + pre-configure apt repo for docker" to "bash + docker".
That would mean executing a command like `docker run --rm --entrypoint=bash ubuntu:18.04 -c "instruction 1 && instruction 2 && etc."` as part of the script.

dduportal:
Thanks for adding a check for emptiness: https://github.com/jenkins-infra/jenkins-infra/runs/4091583727?check_suite_focus=true#step:5:462

aneveux:
> That would mean executing a command like `docker run --rm --entrypoint=bash ubuntu:18.04 -c "instruction 1 && instruction 2 && etc."` as part of the script.

But that would mean the shell script has to install the gpg keys and apt repository for docker-ce right?

(which is fine to me, just want to make sure I understand properly)

dduportal:
> But that would mean the shell script has to install the gpg keys and apt repository for docker-ce right?
>
> (which is fine to me, just want to make sure I understand properly)

That is correct: the shell script would have to do this on each execution.

aneveux:
Hey @dduportal :wave: 

I updated the PR so now the script directly runs in a Docker container.

I actually did 2 scripts, one for running the docker container, the other one which is mounted in the Docker container as a volume to be executed. I thought as there are multiple things to execute it would be either to maintain a separate script file rather than just relying on a massive oneliner.

I tested that locally and it does the job, let me know what you think :)

updatecli/scripts/docker-ce-updates.sh
```diff
@@ -0,0 +1,13 @@
...
+if ! command -v docker >/dev/null 2>&1
+then
+  echo "ERROR: command line 'docker' required but not found. Exiting."
+  exit 1
+fi
+
+SCRIPT_PATH=$(dirname "$0")
+SCRIPT_PATH=$(cd "$SCRIPT_PATH" && pwd)
+
+docker run --rm --name=updatecli-docker-ce --volume="$SCRIPT_PATH":/scripts/ --entrypoint=bash ubuntu:bionic -c "chmod +x /scripts/fetch-docker-ce-latest-version.sh ; /scripts/fetch-docker-ce-latest-version.sh"
```
dduportal:
```suggestion
docker run --rm --name=updatecli-docker-ce --volume="$SCRIPT_PATH":/scripts/ --entrypoint=bash ubuntu:bionic /scripts/fetch-docker-ce-latest-version.sh
```
Nitpicking: no need to add the executable bit to a shell script if you pass this script's location to the command interpreter.

dduportal:
Thanks a lot @aneveux !
That sounds awesome: I've just approved the workflow, let's see the result.

I've added 1 nitpick for the sake of continuous learning :)

aneveux:
> I've added 1 nitpick for the sake of continuous learning :)

That's nice! Thanks for the tip :)

I merged it so I guess we should be good to go :)

dduportal:
That is an excellent job, many thanks!

I've update the PR to match the latest `production` branch where the updatecli GitHub workflow uses the binary instead of the Docker image: it works perfectly now!

dduportal (approved):

PR #1968, opened 2021-11-05
fix(openvpn): ensure cloud-init doesn't manage network

lemeurherve:


dduportal (dismissed):
LGTM 🚀 tested locally with vagrant

dduportal (approved):

PR #1969, opened 2021-11-05
chore(updatecli) use updatecli binary in GitHub Workflow to allow using Docker

dduportal:
(instead of the github action)

The goal of this PR is to allow #1958 to work as expected (using the update-cli action restrict usage of the binaries presents in the Docker image of `updatecli` while we need a Docker Engine)

cc @aneveux @olblak 

It also introduces the following changes:
- Removes Dependabot (no longer needed to upgrade the gh-action version)
- Add an updatecli to upgrade its own version


.github/workflows/updatecli.yaml
```diff
@@ -22,19 +20,23 @@ jobs:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: us-east-2
+      - name: Install updatecli CLI
```
timja, line 23: probably covered somewhere else but there should be an installer action that just installs the binary doesn't try run it in a container, `setup-updatecli` that way dependabot can update it

dduportal: Yup, I should have some free time this week end for that \o/

PR #1981, opened 2021-11-10
fix(buildmasters) ensure that JDK11 is used to start agent process for Azure VM agents

dduportal:
This PR introduces the following changes for all buildmasters managed with JCasc (ci.jenkins and trusted.ci):

- Specifies the java binary used on Azure VMs to force the embeded JDK11 instead of the default java found in path (INFRA-3126)
  - It is still possible to overrides the java binary on a specific templates through node's hieradata
  - It supports both Windows and Linux
-  Factorize in hieradata/common.yaml the default JDK installation paths to ease updates if the upstream jenkins-infra/packer-images changes the path.
    - It supports both Windows and Linux

### DIFF
--- a/dist/profile/manifests/buildmaster.pp
+++ b/dist/profile/manifests/buildmaster.pp
@@ -40,6 +40,7 @@ class profile::buildmaster(
   $agent_images                    = {},
   $additional_tools                = {},
   $default_tools                   = {},
+  $agent_cli_tools                 = {},
   $memory_limit                    = '1g',
   $java_opts = "-server \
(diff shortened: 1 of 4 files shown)


lemeurherve (approved):

dduportal:
Given that there is a huge discrepancy between methods to specify the java binary to be used, I'm closing this PR in favor of https://github.com/jenkins-infra/packer-images/pull/129 (change at the VM template level).

Arguments:
- EC2 agents template does not provide a java bin specification like Azure VMs
- SSH agents provides a java bin but need to be configured differently
- Containers are anoth story
- inbound (JNLP) agent are another different story
- Easier to specify the "default" JDK on a single location, and overrides only when needed (example: for s390x or ppc snowflakes machines)

PR #1992, opened 2021-11-19
INFRA-3124:  Allow configuration of the network for VPN role

gcuisinier:
Based on https://issues.jenkins.io/browse/INFRA-3124 

- [x]  The file `90-cloud-init.yaml` should be changed to a template which values are specified through hieradata, in a "client-only" file 
- [x]  The file `90-cloud-init.yaml` should be renamed to `90-network-config.yaml`
- [x] Ensure that the parent directories exists

Questions :
-  As  the new name of  the network config  file is `90-network-config.yaml`,  should the absence of `90-cloud-init.yaml` must  be enforced (if present by default on VM ) ?


lemeurherve:
> should the absence of 90-cloud-init.yaml must be enforced (if present by default on VM ) ?

No need as it's not present by default on VM.

dist/profile/manifests/openvpn.pp
```diff
@@ -10,7 +10,8 @@
   $openvpn_ca_pem         = undef,
   $openvpn_server_pem     = undef,
   $openvpn_server_key     = undef,
-  $openvpn_dh_pem         = undef
+  $openvpn_dh_pem         = undef,
+  $networks               = undef
```
lemeurherve:
```suggestion
  $networks               = {}
```

dist/profile/manifests/openvpn.pp
```diff
@@ -60,13 +70,24 @@
...
-  file { '/etc/netplan/90-cloud-init.yaml':
+  file { '/etc/netplan/90-network-config.yaml':
     ensure => 'file',
     owner  => 'root',
     group  => 'root',
-    source => "puppet:///modules/${module_name}/openvpn/90-cloud-init.yaml",
+    content => template("${module_name}/openvpn/90-network-config.yaml.erb")
+  }
+
+    file { '/srv/demo/passwd':
```
lemeurherve:
```suggestion
  file { '/srv/demo/passwd':
```

lemeurherve: What's this file '/srv/demo/passwd'?

hieradata/clients/vpn.jenkins.io.yaml
```diff
@@ -0,0 +1,11 @@
...
+networks:
+  eth0: 
+    route-metric: 100
+    macaddress: 00:0d:3a:0e:4b:1c
+  eth1:
+    route-metric: 200
+    macaddress: 00:0d:3a:0e:4f:89
+  eth2:
+    route-metric: 300
+    macaddress: 00:0d:3a:0e:4e:7e
```
lemeurherve: There should be an EOL character at the end of this line.

lemeurherve (changes requested):

dist/profile/manifests/openvpn.pp
```diff
@@ -51,6 +52,15 @@
...
+  file { '/etc/cloud/cloud.cfg.d':
+    ensure  => 'directory',
+    owner   => 'root',
+    group   => 'root',
+    recurse => true
+  }
+
+
   # Ensure cloud-init doesn't manage network (netplan config + netplan apply + systemd, in Ubuntu Bionic)
   # We assume the parent folder already exists
```
lemeurherve: We're testing the existence of the folder above, this comment could be removed.
```suggestion
```

dist/profile/manifests/openvpn.pp
gcuisinier: Copy paste to get a "template"  snippet, forgot to clean.
 I remove it.

gcuisinier:
What do you want in term of commit in PR ?
Multiple commit and rebase at the end ?

or Amend each modification each time to always have one commit ?

dist/profile/manifests/openvpn.pp
gcuisinier: Ok will modify.

lemeurherve:
> What do you want in term of commit in PR ? Multiple commit and rebase at the end ?
> 
> or Amend each modification each time to always have one commit ?

We'll probably squash the commits, don't worry about amending or rebasing them ;)

dist/profile/manifests/openvpn.pp
gcuisinier: Done

dist/profile/manifests/openvpn.pp
gcuisinier: Done

dist/profile/manifests/openvpn.pp
gcuisinier: Done

hieradata/clients/vpn.jenkins.io.yaml
gcuisinier: Done

dist/profile/templates/openvpn/90-network-config.yaml.erb
```diff
@@ -0,0 +1,16 @@
+network:
+    version: 2
+    ethernets:
+    <%- if @networks && !networks.empty? -%>
```
dduportal:
```suggestion
    <%- if @networks && !@networks.empty? -%>
```

dduportal (dismissed):
A blocking typo in the template :)

Thanks a lot though, it's really useful and sounds good on the current vagrant VM!

dduportal (dismissed):
Excellent job! Thanks a lot! Let's ship it 🚀

hieradata/clients/vpn.jenkins.io.yaml
```diff
@@ -0,0 +1,11 @@
+--
```
dduportal:
```suggestion
---
```

dduportal:
I'm adding a few commits to enable testing locally with the Vagrant VM (`vagrant up --provision openvpn`) before merging

dduportal:
Added the following change before merging:
- Added a "vagrant"-specific hieradata to allow a "local" testing (that part is not easy and weird so no worry that it was not on the initial scope of the PR)
- Fixed an issue in the template (the local variables inside the loop should not have an `@` because it's local scope). it was failing the template when trying on the local vagrant (and hard to catch without the vagrant part: imporvement for the team on this repo to helps contributors: @gcuisinier you would have caught this easily if we had been able to provide you a sane testing environment. Sorry for that).
- Adding the requirements for parent directories

dduportal (approved):

dist/profile/manifests/openvpn.pp
```diff
@@ -82,6 +85,9 @@
     ensure  => 'file',
     owner   => 'root',
     group   => 'root',
+    require => [
+      File['/etc/netplan/'],
```
lemeurherve, line 89: @dduportal do we need to verify it there even with the requirement on the same folder above?

dduportal: The doc for puppet 6 specifies that the requirement to parent folder is not mandatory: https://puppet.com/docs/puppet/6/types/file.html#file-description (read the `autorequire` part).

I'm still not sure why, but in some cases it fails so better to make it explicit.

(using the `require` attribute describe the dependency, and Puppet does the test for us)

lemeurherve (approved):

lemeurherve:
Successfully shipped! 🥳

PR #1999, opened 2021-11-29
fix: limit informations returned by Apache to the strict minimum, and desactivate server signature

lemeurherve:


dduportal (approved):
Tested locally with `vagrant up --provision archives`.

Works as expected, but needs an Apache restart (that Puppet does not seem to do on itself)

PR #2003, opened 2021-11-30
fix: remove bandwidth module, as we're on a cloud with far lower bandwidth cost (Oracle)

lemeurherve:
Ref: https://matrix.to/\#/!CuJBSpFAVIYrdHrjnY:libera.chat/\?via\=libera.chat\&via\=matrix.org\&via\=g4v.dev


dduportal (dismissed):
👏

timja (dismissed):

dduportal (approved):

PR #2014, opened 2021-12-04
Use /opt/jdk-11 to run ppc64le Java agent

MarkEWaite:
## Use /opt/jdk-11 to run ppc64le Java agent

The Ubuntu Java 11.0.11 at /usr/bin/java does not complete the connection, while the Eclipse Adoptium Java 11.0.13 that is installed at /opt/jdk-11 completes the connection.

No idea why, but that's what I observe when comparing the working connection from my test environment to the ci.jenkins.io agent that is unable to complete the connection of the agent.

### DIFF
--- a/dist/profile/templates/azure.ci.jenkins.io/agents.yaml.erb
+++ b/dist/profile/templates/azure.ci.jenkins.io/agents.yaml.erb
@@ -12,6 +12,7 @@ jenkins:
         ssh:
           credentialsId: "ppc64le-ed25519-2021-12-03"
           host: "158.175.161.173"
+          javaPath: "/opt/jdk-11/bin/java"
           launchTimeoutSeconds: 713
           port: 22
           retryWaitTime: 175


timja (approved):

dduportal (approved):

MarkEWaite:
Unfortunately, that didn't work long term either.  The agent is again disconnected on the ppc64le agent

timja:
Didn’t we use to get disconnections because of the large MTU default setting? Has that been adjusted on the new VM?

MarkEWaite:
> Didn’t we use to get disconnections because of the large MTU default setting? Has that been adjusted on the new VM?

I remember that we used to get disconnects, but I don't know the cause of the disconnects.  Thanks for the suggestion to check the MTU.  Any location where I can learn more about how to adjust the MTU values and what value might be preferred?

I see the following:

```
$ ip address | grep -i mtu
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
2: ibmveth2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UNKNOWN group default qlen 1000
3: ibmveth3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UNKNOWN group default qlen 1000
4: docker0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default
```

That matches the values on Ubuntu 20.04 that is running the controller in Docker.

timja:
I think it's set in the vsphere console from memory (something for the IBM people to do)

PR #2059, opened 2022-02-10
Update updatecli github action used

olblak:
Let's switch to the github action

### DIFF
--- a/.github/workflows/updatecli.yaml
+++ b/.github/workflows/updatecli.yaml
@@ -1,4 +1,6 @@
+---
 name: updatecli
+
 on:
   # Allow to be run manually
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


lemeurherve:
How about including it in [the existing Jenkinsfile](https://github.com/jenkins-infra/jenkins-infra/blob/production/Jenkinsfile) instead?

olblak:
> How about including it in [the existing Jenkinsfile](https://github.com/jenkins-infra/jenkins-infra/blob/production/Jenkinsfile) instead?

I think that would be even better :)

dduportal:
That would mean changing agent + 

> How about including it in [the existing Jenkinsfile](https://github.com/jenkins-infra/jenkins-infra/blob/production/Jenkinsfile) instead?

That also means migrating to infra.ci (to have the credentials :))

.github/workflows/updatecli.yaml
```diff
@@ -7,44 +9,31 @@ on:
...
-        env:
-          UPDATECLI_VERSION: "0.19.2"
-        run: |
-          curl --silent --location --show-error --output /tmp/updatecli.tgz \
-            "https://github.com/updatecli/updatecli/releases/download/v${UPDATECLI_VERSION}/updatecli_$(uname -s)_$(uname -m).tar.gz"
-          tar xzf /tmp/updatecli.tgz -C /usr/local/bin/ updatecli
-          rm -f /tmp/updatecli.tgz
+
       - name: Diff
-        continue-on-error: true
```
dduportal, line 25: must be kept :)

.github/workflows/updatecli.yaml
```diff
@@ -7,44 +9,31 @@ on:
...
-          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          aws-region: us-east-2
-      ## This step allows to generate a short-lived token which is allowed to modify GitHub Workflow files (that the default GITHUB_TOKEN cannot)
-      ## Please make sure that the 2 secrets are defined (usually at organization level, with a per-repository access)
-      - uses: tibdex/github-app-token@v1.5
-        id: generate_token
-        if: github.ref == 'refs/heads/production'
-        with:
-          app_id: ${{ secrets.JENKINS_ADMIN_APP_ID }}
-          private_key: ${{ secrets.JENKINS_ADMIN_APP_PRIVKEY }}
```
dduportal, line 44: must be kept

.github/workflows/updatecli.yaml
```diff
@@ -7,44 +9,31 @@ on:
...
       - name: Apply
+        uses: updatecli/updatecli-action@v1.19.0
         if: github.ref == 'refs/heads/production'
-        run: |
-          updatecli apply --config ./updatecli/weekly.d --values ./updatecli/values.github-action.yaml
+        with:
+          command: apply
+          flags: "--config ./updatecli/weekly.d --values ./updatecli/values.github-action.yaml"
         env:
-          UPDATECLI_GITHUB_TOKEN: ${{ steps.generate_token.outputs.token }}
```
dduportal, line 50: Must be kept

dduportal (changes requested):
https://github.com/jenkins-infra/jenkins-infra/blob/production/updatecli/weekly.d/docker-ce.yaml#L21 => require docker CE, which is not present in the updatecli's action.


One solution is a feature request on the updatecli-action (ref. https://github.com/updatecli/updatecli-action/issues/21): allow the action to take care of the updatecli install in the current context.

Another idea from @lemeurherve : allows to use a custom Docker image (but won't solve the Docker CE in Docker problem here)

timja:
Imo update cli should just have another action, `setup-updatecli`  which installs update cli.
GitHub actions tend to have a lot of tools installed on the agent / you can add more easily there

dduportal:
Closing this one as we are working on stepping out of GitHub actions for running updatecli (and use our own Jenkins controller, now that it's good enough)

PR #2060, opened 2022-02-11
Use JDK 17.0.2, not JDK 17.0

MarkEWaite:
## Use JDK 17.0.2 rather than JDK 17.0

Extend updatecli for JDK 17 to match JDK 11

Makes progress on jenkins-infra/helpdesk#2758

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -255,7 +255,7 @@ profile::buildmaster::default_tools:
     version: "11.0.14+9"
     sourceURL: https://github.com/adoptium/temurin11-binaries/releases/download
   jdk17:
-    version: "17+35"
+    version: "17.0.2+8"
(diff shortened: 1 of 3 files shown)


timja (dismissed):

dduportal:
Great catch, thanks a lot! 

Currently, this PR works for ci.jenkins.io, but would fail the updatecli process (the conditions uses a shell script that need to be updated because it says `ERROR: unsupported JDK version (17.0.2+8).` for now - ref. https://github.com/jenkins-infra/jenkins-infra/pull/2060/checks#step:4:290).

@MarkEWaite do you mind if I push the fix directly in this PR?

MarkEWaite:
> @MarkEWaite do you mind if I push the fix directly in this PR?

You are always welcome to push directly to any PR that I submit.  Thanks for being willing to do so

dduportal (approved):

PR #2080, opened 2022-02-28
feat(OpenVPN) renew server-side certificate

dduportal:
This PR aims at fixing https://github.com/jenkins-infra/helpdesk/issues/2798.

- Renewing the server-side certificate for OpenVPN (ref. https://community.openvpn.net/openvpn/wiki/EasyRSA3-OpenVPN-Howto for the minimalistic way of doing it, even though we did it with tooling from https://github.com/jenkins-infra/docker-openvpn)
- Moving all the OpenVPN "server" specific hieradata into client-level (instead of "common") to ensure that future "multi-servers" for OpenVPN could work + clarity 
  - Please note that the Docker image tag is kept in commons so both vagrant and production can use it


dduportal:
Tested in production: the certificate + key had been already manually deployed for validation. This PR only allows the config as code to operate and overide manual changes

hieradata/clients/vpn.jenkins.io.yaml
```diff
@@ -9,3 +9,13 @@ profile::openvpn::networks:
...
+# DH comes from https://github.com/jenkins-infra/docker-openvpn
```
lemeurherve, line 20:
```suggestion
# DH comes from https://github.com/jenkins-infra/docker-openvpn, it should not have to be updated
```
Not sure if it should be added. WDYT?

hieradata/common.yaml
```diff
@@ -338,5 +330,5 @@ limits:
     nofile:
       soft: '65536'
       hard: '65536'
-
+profile::openvpn::image_tag: '1.0.28' ## Tracked by updatecli
```
lemeurherve:
```suggestion
profile::openvpn::image_tag: '1.0.28' # Tracked by updatecli
```
(nit)

lemeurherve (dismissed):

hieradata/clients/vpn.jenkins.io.yaml
dduportal: Disagree on this suggestion: the comment implies to never update DH while there are cases where it is required.

lemeurherve (approved):

dduportal:
Merging as admin: the CI error also appears on the production branch and is not caused by this change (but it has to be fixed)

PR #2142, opened 2022-05-02
feat(ci.jenkins.io,trusted.ci.jenkins.io) ensure that pipeline shared library are not cached when on a Pull Request branch

dduportal:
While testing https://github.com/jenkins-infra/pipeline-library/pull/333, we wasted some time before realizing that pipeline-library are cached on ci.jenkins.io.

This change ensures that the cache is ignored for the git references which name follows the pattern `pull/*` too allow using annotations such `@Library('pipeline-library@pull/333/head') _` to end-to-end test  pull requests.

### DIFF
--- a/dist/profile/templates/buildmaster/casc/global-libraries.yaml.erb
+++ b/dist/profile/templates/buildmaster/casc/global-libraries.yaml.erb
@@ -6,6 +6,9 @@ unclassified:
         includeInChangesets: false
         name: "pipeline-library"
         cachingConfiguration:
+          ## Exclude pull request git reference to allow end to end testing when using annotation
+          # such as '@Library('pipeline-library@pull/333/head') _'
+          excludedVersionsStr: "pull/"
           refreshTimeMinutes: 180
         retriever:
           modernSCM:


timja (approved):

jglick:
Sounds like a bug in `workflow-cps-global-lib` if you even need to do this—it ought to be smart enough to keep track of which commit it cached. (I wrote the original external library code but was not involved in the subsequent addition of the caching feature so I do not know much about it.)

dduportal:
@jglick Let me clarify to be sure that I did not write too quickly the initial intent of the PR: the "problem" we solved with this configuration change was when we "real-life" test a PR on the shared libraries, when we add new commits to the PR.

Usually the first test works well (branch is not cached, the system gets the HEAD of the PR for the library).
Obviously the test fails, so you push a new commit to the library PR and trigger a new pipeline build for testing it: the previous (failing) commit is used because it is cached.

timja:
Yes but the point is, the system should know which commit is cached previously and it should check the HEAD of the branch / PR whatever and it should automatically invalidate the cache if it doesn't match.

I assume the expensive part is the cloning not checking the head commit...

dduportal:
> 

Gotcha, thanks for the clarification and explanations!

I wouldn't have expected the caching mechanism to do it intuitively, so not sure if it qualifies as a bug.
But whatever "qualification" (bug/feeature request/other) we put, this behavior of checking the heead would be nice to have!

PR #2151, opened 2022-05-04
cleanup(ci.jenkins.io,trusted.ci.jenkins.io) remove terraform credential configuration (file does not exist on controllers)

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2908,

This PR stops trying to add the Azure k8s SSH credential to the Jenkins controllers because the private key does not exist on the file system.

Should remove the error message `java.io.FileNotFoundException: /var/jenkins_home/.ssh/azure_k8s.pub (No such file or directory)`.

Please note that the file `terraform-credentials.groovy` must be manually deleted on the VMs once the change is applied.


lemeurherve (approved):

dduportal:
Groovy file removed on both ci.jenkins.io and trusted.ci.jenkins.io's VMs.

Verified on trusted after puppet run, by restarting the controller: no more error message for this element.

PR #2156, opened 2022-05-05
feat(rating) removing rating from puppet

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/1627
removing rating from puppet


dduportal (approved):
🚀

PR #2158, opened 2022-05-05
cleanup catchall): remove unused role and profile catchall

smerle33:
During https://github.com/jenkins-infra/helpdesk/issues/1627 removing rating we found out remnants of a role and profile for catchall  that weren't used anymore.


dduportal:
@smerle33 can you:
- Give a bit more context in the PR message (why did you end up in this PR?)
- Update the title of the PR to express (look at my previous cleanup PRs as example):
  - The fact that it's a cleanup  (it is not a new feature. I think I tend to do the same mistake some time). I tend to use `cleanup` instead of `feat` because https://www.conventionalcommits.org/en/v1.0.0/ does not provide something that is clear enough (or maybe chore?).
  - The fact that both a role and a profile are removed

dduportal (approved):

PR #2167, opened 2022-05-11
feat(rngd): add rngd package to linux base VM

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/2912 add rngd package presence check in the puppet configuration.

### DIFF
--- a/dist/profile/manifests/rngd.pp
+++ b/dist/profile/manifests/rngd.pp
@@ -1,7 +1,5 @@
-#
 # Profile for managing rngd packet installation
 class profile::rngd {
-
   ['rng-tools'].each | $package | {
     package { $package:
       ensure => present,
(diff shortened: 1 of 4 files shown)


dist/profile/manifests/rngd.pp
```diff
@@ -0,0 +1,10 @@
+#
+# Profile for managing rngd packet installation
+class profile::rngd {
+
```
lemeurherve:
```suggestion
# Profile for managing rngd packet installation
class profile::rngd {
```

spec/classes/profile/rngd_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'spec_helper'
+
+describe 'profile::rngd' do
+  it { should contain_class 'profile::rngd' }
```
dduportal:
```suggestion
```

This is not needed since the test already tests the class itselve

dduportal (approved):
👍🏼👍🏼🎉

PR #2190, opened 2022-05-28
Load image with https

MarkEWaite:


dduportal (approved):

lemeurherve:
@dduportal Who owns xmission.com? Shouldn't we link to an image on jenkins.io instead of hotlinking there?

smerle33:
> @dduportal Who owns xmission.com? Shouldn't we link to an image on jenkins.io instead of hotlinking there?

I would say no, as it seems to be an outage page, so limiting the risk to have a dead image.

PR #2192, opened 2022-05-28
Update "mirrors.jenkins.io" references to use https

### LINKED ISSUE #2191: Installing Jenkins from pkg.jenkins.io via APT is failing due to https -> http redirect
See https://stackoverflow.com/q/72412100/433558 for a concrete instance.

The short version is that folllowing the instructions from https://pkg.jenkins.io/debian/ verbatim now fails with something like this:

```console
Err:28 https://pkg.jenkins.io/debian binary/ jenkins 2.349
  Redirection from https to 'http://mirrors.jenkins.io/debian/jenkins_2.349_all.deb' is forbidden [IP: 151.101.2.133 443]
```

I believe the underlying cause is https://www.jenkins.io/blog/2022/05/13/mirrors/, and I believe the fix is probably updating to `https` in the following line? (and maybe any/all other references to `mirrors.jenkins.io`?)

https://github.com/jenkins-infra/jenkins-infra/blob/df1a23b54bce3328f7466dc4fb4623a1173925ab/dist/profile/templates/pkgrepo/debian_htaccess.erb#L16

**Edit:** ah, https://github.com/jenkins-infra/helpdesk/issues/2957#issuecomment-1140098592 :see_no_evil: :heart:

tianon:
Per https://www.jenkins.io/blog/2022/05/13/mirrors/ 👍

Closes https://github.com/jenkins-infra/jenkins-infra/issues/2191
Refs https://github.com/jenkins-infra/helpdesk/issues/2957 (:pray:)

(Happy to rebase, amend, update, let you take over, etc :heart:  The goal is for this to be helpful and if it isn't, please feel free to close it!  No hard feelings :smile:)

### DIFF
--- a/dist/profile/templates/pkgrepo/debian_htaccess.erb
+++ b/dist/profile/templates/pkgrepo/debian_htaccess.erb
@@ -13,4 +13,4 @@ RewriteRule ^binary/(.*)\.deb$  /<%= @name %>/direct/$1.deb [L]
 # See also: https://issues.jenkins-ci.org/browse/INFRA-964
 RewriteCond %{HTTPS} on
 
-RewriteRule ^binary/(.*)\.deb$ http://mirrors.jenkins.io/<%= @name %>/$1.deb [R=302,L]
+RewriteRule ^binary/(.*)\.deb$ https://mirrors.jenkins.io/<%= @name %>/$1.deb [R=302,L]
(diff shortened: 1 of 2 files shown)


timja (approved):

timja:
Tests appear to be broken by the previous hotfix, PR.

yolo’ing this.

dduportal:
Many many thanks @tianon @timja for this fix!

- Tests have been fixed in #2194 
- The redirection in here has been applied (puppet agent was disabled temporarly given the incident) along with the public key fix in #2195

PR #2196, opened 2022-05-30
feat(pkg/update): puppet conf for apache MaxRequestWorkers

smerle33:
Related to https://github.com/jenkins-infra/helpdesk/issues/2960#issuecomment-1141083536.


hieradata/clients/pkg.yaml
```diff
@@ -0,0 +1,3 @@
+---
+apache::mod::event::serverlimit: '30' #default 25
```
dduportal: What is the reason to select 30 ?

hieradata/clients/pkg.yaml
```diff
@@ -0,0 +1,3 @@
...
+apache::mod::event::maxrequestworkers: '750' #ServerLimit * 25
```
dduportal: - Can you add a comment linking to the apache doc. with the formula?
- Where does the 25 value come from?

hieradata/clients/pkg.yaml
dduportal:
```suggestion
# https://www.liquidweb.com/kb/apache-performance-tuning-swap-memory/
# trashing point, with ~30 Gb of RAM and ~500Mb (worst-case full apache) means around 50.
# We are conservative and keep 30 (to be monitored and validated)
apache::mod::event::serverlimit: '30' #default 25
```

hieradata/clients/pkg.yaml
dduportal:
```suggestion
# https://httpd.apache.org/docs/2.4/fr/mod/mpm_common.html#maxrequestworkers
# maxrequestworkers = ServerLimit (see value above) * ThreadsPerChild (default: 25)
apache::mod::event::maxrequestworkers: '750'
```

dduportal (approved):

PR #2229, opened 2022-06-27
feat(EKS): temporarily disable EKS cluster during upgrade to 1.22

smerle33:
by setting the capacity to 0 on this cluster
as per https://github.com/jenkins-infra/helpdesk/issues/2930


lemeurherve (approved):
Let's go!

smerle33:
as a matter of fact it turns out not to set a 0 but empty which means no limit ....

PR #2235, opened 2022-06-30
chore: switch vagrant to Docker

dduportal:
This PR is an attempt at moving away from virtualbox to docker as a provider to ensure an easier acceptance testing of the puppet provisioning.

The main target are macOS ARM CPUs for the first iteration, to allow team members with such a setup to be able to test their system.

Ping @smerle33 @timja if you are able to test this on your M1 machines:

```shell
# Checkout this PR locally
gh pr checkout 2235

# Bootstrap dependencies (requires `bundle` command installed and in your PATH)
./vagrant-bootstrap

# Install vagrant - https://www.vagrantup.com/downloads or "bundle install vagrant" if no option

# Try to start the pkg VM
vagrant up pkg
```

### DIFF
--- a/dist/profile/manifests/docker.pp
+++ b/dist/profile/manifests/docker.pp
@@ -1,7 +1,12 @@
 #
 # Profile for managing basics of docker installation/configuration
 class profile::docker {
-  class { '::docker':
+  class { 'docker':
+  }
+
+  if ($facts['os']['architecture'] == 'aarch64') {
+    # 'https://github.com/puppetlabs/puppetlabs-docker/issues/494 (even with APT module at 8.4.1)
+    Apt::Source <| architecture == 'aarch64' |> { architecture => 'arm64' }
   }
 
   include datadog_agent::integrations::docker_daemon
@@ -10,8 +15,8 @@ class profile::docker {
   user { $datadog_agent::params::dd_user:
     ensure  => present,
     groups  => ['docker'],
-    require => Class['::docker'],
-    before  => Class['datadog_agent::integrations::docker_daemon']
+    require => Class['docker'],
+    before  => Class['datadog_agent::integrations::docker_daemon'],
   }
 
   firewall { '010 allow inter-docker traffic':
--- a/README.md
+++ b/README.md
@@ -62,14 +62,14 @@ catalog on a VM.
 ##### Pre-requisites for Vagrant
 
 * Install [Vagrant](https://www.vagrantup.com)
-* Install Vagrant plugins: `vagrant plugin install vagrant-serverspec`
+<!-- * Install Vagrant plugins: `vagrant plugin install vagrant-serverspec` -->
 * Run the `./vagrant-bootstrap` script locally to make sure your local
   environment is prepared for Vagranting
 
 To launch a test instance, `vagrant up ROLE` where `ROLE` is [one of the defined roles](dist/role/manifests).
 You can rerun puppet and execute tests with `vagrant provision ROLE` repeatedly while the VM is up and running.
-To just rerun serverspect without puppet, `vagrant provision --provision-with serverspec ROLE`.
-When it's all done, deprovision the instance via `vagrant destroy ROLE`.
+<!-- To just rerun serverspect without puppet, `vagrant provision --provision-with serverspec ROLE`.
+When it's all done, deprovision the instance via `vagrant destroy ROLE`. -->
 
 ### Updating dependencies
 
--- a/README.md
+++ b/README.md
@@ -50,26 +50,49 @@ The amount of testing that can be done locally is as follows:
   the rspec-puppet over and over, use `rake spec_standalone` to avoid
   re-initializing the Puppet module fixtures every time.
 
+### Pre-requisites for local development
+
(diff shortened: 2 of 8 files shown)


timja:
I get this error:
https://github.com/rapid7/metasploit-framework/issues/11368

Trying to run the suggested fix, (quite slow though)

timja:
Next error is around installing r10k, nothing seems to install it and running `gem install r10k` doesn't add it to the path / I haven't found where it installs to

dduportal:
> Next error is around installing r10k, nothing seems to install it and running `gem install r10k` doesn't add it to the path / I haven't found where it installs to

Can you try prepending `bundle exec` on the `r10k` command? I remember cases where bundler maintains its own path

timja:
`tried to create Proc object without a block`

I suspect it needs an old version of ruby? possible to update it?

dduportal:
> `tried to create Proc object without a block`
> 
> I suspect it needs an old version of ruby? possible to update it?

Gotta check, thanks for the feedback!

dduportal:
@timja we juste tested on @smerle33 's M1 machine, with an installation from scratch (so with the macOS default ruby) and did not had these issues: I'll need help for reproduction (to provide the correct fix).

We succesfully tried (and pushed) the following:
- Fixed the r10k issue: since the `Gemfile` specifies `r10k`, we prepended the `bundle exec` to make sure it's not required to be installed before. On Stephan's machine, we had to cleanup r10k with `sudo gem uninstall r10k` before as he already had it installed and it was conflicting. I did not had any conflict (tested successfully) on my intel mac.
- Succesfully started the `pkg` "vagrant machine" (`vagrant up pkg`) which provisionned successully the role, with an amd64 Ubuntu 18.04 container on my machine and an arm64 Ubuntu 18.04 container on @smerle's machine.
- We had the `jenkins::master` vagrant machine failing when installing the Docker in Docker installation. I was able to fix this and test it successfully on my machine (`vagrant up jenkins::master`) but I was not able to validate on a M1 CPU though

smerle33:
I can confirm that it's working perfectly : "vagrant up jenkins::master"
Capture d’écran 2022-07-04 à 10 16 02

timja:
I have a 3.x ruby not sure what the default one is.

dduportal:
> I have a 3.x ruby not sure what the default one is.

On my amd64 macOS 12.4, I got the following system ruby installation (which works) for reference:

```shell
$ which ruby && ruby --version
/usr/bin/ruby
ruby 2.6.8p205 (2021-07-07 revision 67951) [universal.x86_64-darwin21]
```


Let me try with a ruby 3.0 (using asdf). But it means that I should write the requirement in this PR.

dduportal:
OK, README updated to mention the Ruby constraints (I've double checked my machine AND the Docker image).

Tested with success the `vagrant up pkg` and `vagrant up jenkins::master` on @smerle33 's M1 macOS + my AMD64 machine.

As it incrases the amount of environments where the vagrant stuff can be done, I'm setting this PR ready to review as it should be usable.

In order to improve this project and share knowledge, I've opened some issues:
- https://github.com/jenkins-infra/jenkins-infra/issues/2247
- https://github.com/jenkins-infra/jenkins-infra/issues/2248
- https://github.com/jenkins-infra/jenkins-infra/issues/2249
- https://github.com/jenkins-infra/jenkins-infra/issues/2250

dist/profile/manifests/docker.pp
```diff
@@ -4,6 +4,11 @@
   class { '::docker':
   }
 
+  if ($::architecture == 'aarch64') {
+    # 'https://github.com/puppetlabs/puppetlabs-docker/issues/494
+    Apt::Source <| architecture == 'aarch64' |> { architecture => 'arm64' }
```
timja, line 9: fixed as of 8.4.0 apparently, I assume we're using an older version though?

timja (dismissed):

dist/profile/manifests/docker.pp
dduportal: Yep: https://github.com/jenkins-infra/jenkins-infra/blob/production/Puppetfile#L34

But that one should be quite easy to bump though

dduportal: https://github.com/jenkins-infra/jenkins-infra/pull/2251

dduportal:
Putting as draft until https://github.com/jenkins-infra/jenkins-infra/pull/2251 is merged + wip

dist/profile/manifests/docker.pp
dduportal: Currently trying without this code block (removed on the last push force)

timja (dismissed):

dist/profile/manifests/docker.pp
dduportal: Damn, sounds like that the fix on the apt module is not used in the docker module (https://github.com/puppetlabs/puppetlabs-docker/issues/494).

dduportal: Updated the code: the `if` condition is back with an improved comment (hoipint hat the docker module will solve issues one day..)

PR #2257, opened 2022-07-07
Revert "feat(ci.jenkins.io) enable spot instances for EC2 VM agents (INFRA-3101)"

lemeurherve:
Reverts jenkins-infra/jenkins-infra#1956

[Disable “spot” for highmem EC2 instances (so we can check the impact on budget and confirm that spot is the major culprit)](https://github.com/jenkins-infra/helpdesk/issues/3031#issuecomment-1177498315)

Ref: https://github.com/jenkins-infra/helpdesk/issues/3031

PR #2259, opened 2022-07-07
feat(EC2): add a 'spot' parameter and disable it for highmem EC2 instances

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3031

Replace https://github.com/jenkins-infra/jenkins-infra/pull/2257


hieradata/clients/azure.ci.jenkins.io.yaml
```diff
@@ -229,6 +229,7 @@ profile::buildmaster::cloud_agents:
             - docker
             - linux
           useAsMuchAsPosible: true
+          spot: true
```
lemeurherve, line 232: I've used the same parameter name as the one used for azure VMs

dist/profile/templates/buildmaster/casc/clouds.yaml.erb
```diff
@@ -127,9 +127,11 @@ jenkins:
         remoteAdmin: "<%= @agents_setup[agent["os"].to_s]["remoteAdmin"] %>"
         remoteFS: "<%= @agents_setup[agent["os"].to_s]["agentDir"] %>"
         securityGroups: "ci-agents"
+        <%- if agent["spot"] -%>
```
lemeurherve, line 130: Note: on azure VMs, the test is done on `agent["spotInstance"] || agent["spot"]`, I've kept the later only here.

dduportal: Keeping `spot` make sense 👍

dduportal (approved):

PR #2262, opened 2022-07-07
chore: no spot instances for highmem

lemeurherve:
Ref: jenkins-infra/helpdesk#3031, more precisely https://github.com/jenkins-infra/helpdesk/issues/3031#issuecomment-1177746463

### DIFF
--- a/hieradata/vagrant/common.yaml
+++ b/hieradata/vagrant/common.yaml
@@ -102,7 +102,7 @@ profile::buildmaster::cloud_agents:
           useAsMuchAsPosible: true
           spot: true
           idleTerminationMinutes: 30
-        - description: "High memory ubuntu 20.04 (spot)"
+        - description: "High memory ubuntu 20.04 (ondemand)"
           maxInstances: 20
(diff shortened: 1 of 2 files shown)


dduportal (approved):

hieradata/vagrant/common.yaml
```diff
@@ -112,17 +112,6 @@ profile::buildmaster::cloud_agents:
...
-          maxInstances: 50
-          instanceType: M54xlarge # 16 vCPUS / 64 Gb
-          os: "ubuntu"
-          architecture: "amd64"
-          labels:
-            - highmem-ondemand
-            - highram-ondemand
-            - docker-highmem-ondemand
-          useAsMuchAsPosible: false
           spot: false
```
lemeurherve, line 115: Note: I didn't found any "spot" parameter for azure VM in vagrant

dduportal: for sure it was because of a bad copy and paste :) We can do whatever we want with the vagrant template: the only goal is to use it to "acceptance" test a full puppet agent run.

Issue #2250, opened 2022-07-04
Acceptance Tests: allow "Per VM" tests


dduportal:
- Acceptance test harness is aimed at testing Puppet roles. Which should be done automaically (#2248 and #2249 )
- Infra team should be able to test "per node", e.g. allowing Vagrant to spin up a "copy" of the production node (or at least a node that is as close as possible) to increase the trust in changes.
  - We could bring back the `staging` branching model from a few years ago, as set up by Tyler and Olivier alternatively.

dduportal:
Won't be done as per https://github.com/jenkins-infra/helpdesk/issues/4714

Issue #2249, opened 2022-07-04
Bring back the Acceptance Test Harness


dduportal:
As underlined in https://github.com/jenkins-infra/jenkins-infra/pull/2235 Serverpsec tests are deprecated since Q1 2021 (the vagrant plugin is not maintained anymore and is not working with Vagrant 2.x).

We should bring back a test harness as part of the vagrant acceptance testing:

- either find a solution to run ServerSpec v2 inside the vagrant "instance", after the puppet provision, and returns an exit code  based on its status + test reports.
  - Pros: we **might** be able to reuse the existing serverspec logic + rpsec is the same language as for unit tests (but with different assertion resources though)
- OrsSwitch to [Goss](https://github.com/aelsabbahy/goss) as it can also be used for [Docker with the `dgoss` wrapper](https://github.com/aelsabbahy/goss/tree/master/extras/dgoss) and provides automatic adding tests

dduportal:
Won't be done as per https://github.com/jenkins-infra/helpdesk/issues/4714

Issue #2248, opened 2022-07-04
Bring back automated acceptance testing for roles


dduportal:
With https://github.com/jenkins-infra/jenkins-infra/pull/2235, Vagrant should be usable with Docker (instead of virtualbox).

Which means that we can enable automated acceptance tests as part of the ci.jenkins.io build by scheduling theses tasks in a VM with the `docker` label`.

Requires an update on the VM template (ref. https://github.com/jenkins-infra/packer-images/issues/198) to add
- Vagrant
- Ruby + bundler (check #2247 for the exact version)

[edit]
The target would be infra.ci.jenkins.io, to avoid competing for executors with other jobs such as BOM when it's an infra related task.

dduportal:
Won't be done as per https://github.com/jenkins-infra/helpdesk/issues/4714

Issue #2247, opened 2022-07-04
Bump Ruby dependency to 3.x (or at least 2.7)


dduportal:
As per https://www.ruby-lang.org/en/downloads/ , Ruby 2.6 is end of line.

We should update this project to run on ruby 2.7, at least (with bundler from 1.17 to 2.x!), or better: to the latest stable 3.x.

https://github.com/jenkins-infra/jenkins-infra/pull/2235 showed us that Ruby 3.x does not work out of the box (or maybe it's bundler 2.x).

The ci.jenkins.io is, for now, defined in the following Docker image: https://github.com/jenkins-infra/docker-inbound-agents/blob/main/ruby/Dockerfile which is quite outdated (Ruby 2.6, Debian stretch).

dduportal:
As mentioned in https://github.com/jenkins-infra/packer-images/pull/269#issue-1294220374: once we'll bump ruby minor (or major) version, we'll have to add an updatecli manifest to track ruby update (mandatory).

timja:
Ruby 3.x would be great. :)

dduportal:
Won't be done as per https://github.com/jenkins-infra/helpdesk/issues/4714

PR #2263, opened 2022-07-08
chore(unit-tests) generate test fixtures from Puppetfile

dduportal:
This PR is a big cleanup on the Puppet unit test to ensure that we can keep it working and upgrade dependencies in the future.  Had to untangle quite some elements but it's hard to split into small PRs, so here we go.

EDIT: please only take teh "ci.jenkins.io" check, not the infra.ci.jenkins.io one.

- Remove the file `.fixtures.yml` from the repository:
  - This file duplicates the content of `Puppetfile` and required manual copy and paste, leading to unit test running with different puppet modules than production
  - Added to gitignore and generated through the `generate-puppetfile` new ruby gem which reads `Puppetfile` and generates 
 
- Fix the Puppet code to support both Puppet v4 (version used for unit tests) and v6 (production server), using Puppet formatter for most of them:
  - `include` directive does not need  the`::` prefix anymore when loading a top-level module
  - `ensure` says `file` instead of `present` for resources of type `file`
  - `ensure` is deprecated for `concat::fragment` resource

- Fix Unit Tests Harness with the errors raised by the having an up-to-date `.fixtures.yml` file:
  - Correctly manage `facter`'s mock facts to support both Puppet v4 and v6 in `spec/spec_helper.rb`
  - Stop testing for default values of the `apache::mod::status` module as it's not available in this mocked context with recent versions (should be an acceptance test)
  - Fix `census` tests that were checking a deprecated attrbitue for the resource `concat::fragment` 

- Update the Pipeline (`Jenkinsfile`):
  - Update + cleanupo the shell scripts and  `Rakefile` to make sure that both pipeline and local development can behave the same
  - Only spawn 1 agent for the whole build and run steps in parallels. Consumes less executors, and the dependencies are solved once per build. it's not faster, but ensure our builds are not stuck when there is a hugh activity on ci.jenkins.io
  - No need for junit test results (at least with our outdated dependencies for ruby specs gems) 
  - Archive the `.fixtures.yml` file generated during the build for auditing purpose

- Documentation: README updated (was long due)

Please note that with this PR, we can start thinking about:

- Bumping dependencies on `Gemfile` to recent (and decent) versions including Puppet v4 to v6
- https://github.com/jenkins-infra/jenkins-infra/issues/2247
- https://github.com/jenkins-infra/jenkins-infra/issues/2248
- https://github.com/jenkins-infra/jenkins-infra/issues/2249
- https://github.com/jenkins-infra/jenkins-infra/issues/2250
- Bumping dependences on `Puppetfile`


Jenkinsfile
```diff
@@ -1,54 +1,39 @@
...
-        GIT_AUTHOR_NAME = 'Git'
-        GIT_AUTHOR_EMAIL = 'git@example.com'
-    }
-
     stages {
+        stage('Prepare Puppet Project') {
+            steps {
+                // Install Dependencies once for all
+                sh 'bash ./ci/00_setupgems.sh'
+                // For auditing purposes: if test are failing with "module not found" or "object not found" for instance
```
lemeurherve:
```suggestion
                // For auditing purposes: if tests are failing with "module not found" or "object not found" for instance
```

Puppetfile
```diff
@@ -87,3 +68,29 @@ mod 'erwbgy/limits', '0.3.1'
...
+  :git => 'https://github.com/jenkins-infra/puppet-irc.git',
+  :ref => '4e5e437'
+# Needed for managing our accounts in hiera, this fork contains the pull
+# request which adds support for multiple SSH keys:
+# <https://github.com/torrancew/puppet-account/pull/18>
+mod 'account',
+  :git => 'https://github.com/jenkins-infra/puppet-account.git',
+  :ref => '1deebe9'
+mod 'jenkins_keys',
+  :git => 'git@github.com:jenkins-infra/jenkins-keys.git',
```
lemeurherve, line 91: Why this one has a `git@github.com` url instead of an `https://github.com` one?

README.md
```diff
@@ -39,28 +35,26 @@ This repo just manages and configures the deployments.
...
 
 You can **always** check the Docker image that ci.jenkins.io uses to run the test harness for this project at <https://github.com/jenkins-infra/docker-inbound-agents/blob/main/ruby/Dockerfile> (Jenkins agent labelled with `ruby`).
 
+### Install Local Dependencies
+
+Run the script `./ci/00_setupgems.sh` to ensure that all the local dependencies are ready for local development, including:
+
+* Ruby Gems managed by `bundler` (through `Gemfile` and `Gemfile.lock`) to ensure development tools are available through `bundle exec <tool>` commands, installed to `vendor/gems`
+* Puppet modules retrieved from `./Puppetfile` and installed to `./modules`
+* Unit Tests fixtures generated from  `./Puppetfile` into `.fixtures.yml` but also other locations in `./spec/`
```
lemeurherve:
```suggestion
* Unit Tests fixtures generated from `./Puppetfile` into `.fixtures.yml` but also other locations in `./spec/`
```

README.md
```diff
@@ -146,6 +133,6 @@ See [this page](https://github.com/jenkins-infra/.github/blob/master/CONTRIBUTIN
 
 Channels:
 
-* `#jenkins-infra` on the [Freenode](http://freenode.net) IRC network
-* [INFRA project](https://issues.jenkins-ci.org/browse/INFRA) in JIRA.
```
lemeurherve, line 150: Oops, missed that one on the infra issues migrations, good catch!

ci/00_setupgems.sh
```diff
@@ -1,10 +1,13 @@
...
 
-gem install bundler --no-ri --no-rdoc
+# Show version to help debugging (or fail fast)
+yq --version
+ruby -v
+bundle -v
 
+# Install Unit Test Dependencies
 mkdir -p vendor/gems
+bundle install --without development plugins --path=vendor/gems
```
lemeurherve, line 10:
```suggestion
bundle install --verbose --without development plugins --path=vendor/gems
```
Not sure if needed anymore

spec/spec_helper.rb
```diff
@@ -20,22 +20,47 @@
 
   c.hiera_config = File.join(FIXTURES_PATH, 'hiera.yaml')
 
+  # Use the command `facter` on any production node to see real-life values
```
lemeurherve:
```suggestion
  # Use the `facter` command on any production node to see real-life values
```

lemeurherve (dismissed):
Great job! 🎉

ci/00_setupgems.sh
dduportal: The verbose mode for `bundle install` does not feel really useful. It happens when there is an issue with one of the dependencies.
But it adds a time penalty (twice the time on my local macOS :'( ) + it is not set on the actual Jenkinsfile on the principal branch, so I've removed it.

Puppetfile
dduportal: It's a constraint with the Git authentication: the puppetmaster uses SSH authentication with a "deploy key" defined on the repository jenkins-infra/jenkins-keys.

Rationale was because an Oauth token would have add too much permissions (no filtering by repository) while an SSH key is defined at a repository-level, with either read-only or read/write on the content (and no GitHub API access).

I'm not sure if using a GitHub app would be possible today: we would need to implement a hook in Puppetmaster's lifecycle, hence keeping the SSH.


Beside, it is a private repository that is removed from the `.fixtures.yml` so ci.jenkins.io never clones it.

smerle33 (approved):
sounds great
🚀

lemeurherve (approved):

PR #2270, opened 2022-07-11
chore: block remote access API only on ci.jenkins.io instance

### LINKED ISSUE #3045: Remote access API on every non-ci.j.io instance
### Service(s)

cert.ci.jenkins.io, trusted.ci.jenkins.io

### Summary

Could every private instance have the remote API accessible? 🥺

With authentication required on those instances, and Basic authenticated requests allowed by the reverse proxy rule, the only difference is interactive use, which I attempt occasionally. Not a blocker, but just one of those behaviors improperly applied to private instances because ci.j.io had it.

release.ci and infra.ci are unaffected (Kubernetes special?) but trusted.ci and cert.ci are.

### Reproduction steps

_No response_

lemeurherve:
Fixes https://github.com/jenkins-infra/helpdesk/issues/3045

### DIFF
--- a/dist/profile/manifests/jenkinscontroller.pp
+++ b/dist/profile/manifests/jenkinscontroller.pp
@@ -335,67 +335,7 @@ class profile::jenkinscontroller (
     require => File[$docroot],
   }
 
-  apache::vhost { $ci_fqdn:
-    serveraliases         => [
-      # Give all our jenkinscontroller profiles this server alias; it's easier than
-      # parameterizing it for compatibility's sake
-      'ci.jenkins-ci.org', $ci_resource_domain,
-    ],
-    require               => [
-      Docker::Run[$docker_container_name],
-      File[$docroot],
-      # We need our installation to be secure before we allow access
-      File[$groovy_d],
-    ],
-    port                  => 443,
-    override              => 'All',
-    ssl                   => true,
-    docroot               => $docroot,
-
-    access_log_pipe       => "|/usr/bin/rotatelogs -t ${apache_log_dir}/access.log.%Y%m%d%H%M%S 86400",
-    error_log_pipe        => "|/usr/bin/rotatelogs -t ${apache_log_dir}/error.log.%Y%m%d%H%M%S 86400",
-    proxy_preserve_host   => true,
-    allow_encoded_slashes => 'on',
-    custom_fragment       => "
-RequestHeader set X-Forwarded-Proto \"https\"
-RequestHeader set X-Forwarded-Port \"${proxy_port}\"
-RequestHeader set X-Forwarded-Host \"${ci_fqdn}\"
-
-RewriteEngine on
-
-RewriteCond %{REQUEST_FILENAME} ^(.*)api/xml(.*)$ [NC]
-RewriteRule ^.* \"https://jenkins.io/infra/ci-redirects/\"  [L]
-
-# Abusive Chinese bot that ignores robots.txt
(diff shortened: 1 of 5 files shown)


dist/profile/manifests/jenkinscontroller.pp
```diff
@@ -430,6 +415,16 @@
...
+# Loading our Proxy rules ourselves from a custom fragment since the
+# puppetlabs/apache module doesn't support ordering of both proxy_pass and
+# proxy_pass_match configurations
+ProxyRequests Off
+ProxyPreserveHost On
+ProxyPass / http://localhost:8080/ nocanon
+ProxyPassReverse / http://localhost:8080/
+"
+  if $block_api_json_path {
+    $custom_fragment = "${custom_fragment}
```
dduportal: Great job! I remember that a variable in Puppet can only be set once and is immutable then so this line seems weird.

Were you able to try this in the vagrant-docker stack? I'm not 100% sure of myself on this so acceptance testing looks the way to go ;)

(that would only mean renaming the first occurence of `$custom_fragment` to something like `$base_custom_fragment` and etvoilà)

lemeurherve: I also saw `unless $groovy_init_enabled {` elsewhere, but I'm not sure of the difference it would make instead of the `if`

lemeurherve: Added a $base_custom_fragment

dist/profile/manifests/jenkinscontroller.pp
```diff
@@ -334,31 +335,10 @@
...
-    access_log_pipe       => "|/usr/bin/rotatelogs -t ${apache_log_dir}/access.log.%Y%m%d%H%M%S 86400",
-    error_log_pipe        => "|/usr/bin/rotatelogs -t ${apache_log_dir}/error.log.%Y%m%d%H%M%S 86400",
-    proxy_preserve_host   => true,
-    allow_encoded_slashes => 'on',
-    custom_fragment       => "
+  $base_custom_fragment = "
 RequestHeader set X-Forwarded-Proto \"https\"
 RequestHeader set X-Forwarded-Port \"${proxy_port}\"
-RequestHeader set X-Forwarded-Host \"${ci_fqdn}\"
+RequestHeader set X-Forwarded-Host \"${ci_resource_domain}\"
```
daniel-beck: Is that the wrong domain?

dduportal: Yep, that line might have to be separated in another custom fragment

lemeurherve: I've separated it, I missed that one, good catch!

dist/profile/manifests/jenkinscontroller.pp
```diff
@@ -335,10 +335,17 @@
     require => File[$docroot],
   }
 
+  $ci_fqdn_x_forwarded_host = "
+RequestHeader set X-Forwarded-Host \"${ci_fqdn}\"
```
dduportal, line 339: Sounds like you might want to define a template, because the amount of `if` makes this really hard to read and maintain (ref. https://puppet.com/docs/puppet/6/lang_template_erb.html#erb_use-evaluating-ERB-template-strings) .

lemeurherve: As discussed together and as I don't see how to pass a value to the inline_template function, I'll keep this construction.

dduportal (approved):

PR #2274, opened 2022-07-12
fix: move the other api/xml redirection in the if too

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3045


dduportal (approved):
LGTM 👍

PR #2280, opened 2022-07-14
chore: migrate pipeline to infra.ci.jenkins.io

dduportal:
This PR migrates the pipelien from ci.jenkins.io to infra.ci.jenkins.io.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,8 @@
 #!/usr/bin/env groovy
 
 pipeline {
-    agent { label 'ruby' }
+    // All Linux agents have ruby + bundle installed through `asdf`
+    agent { label 'linux-amd64-docker' }
 
     options {
         buildDiscarder(logRotator(numToKeepStr: '10'))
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -1,17 +1,33 @@
...
 
     options {
         buildDiscarder(logRotator(numToKeepStr: '10'))
         timeout(time: 15, unit: 'MINUTES')
         timestamps()
     }
 
+    environment {
+        // To allow using ASDF shims
+        PATH = "${env.PATH}:/home/jenkins/.asdf/shims:/home/jenkins/.asdf/bin"
```
timja, line 15: why not add this to the path of the image?

Jenkinsfile
```diff
@@ -1,17 +1,33 @@
...
+    }
+
     stages {
         stage('Prepare Puppet Project') {
             steps {
+                // Install `yq` until https://github.com/jenkins-infra/packer-images/pull/277 is merged and available
```
timja, line 21: resolve this first? unless urgent?

Jenkinsfile
dduportal: I wasn't able to have a custom PATH working in the VMs. I assume it might be related to the current SSH configuration (infra.ci uses EC2 VM agents so SSH agents) that filters out.

None of the usual `profile` and `bashrc` are not loaded.

Jenkinsfile
dduportal: urgent no, but we can work in parallel

dduportal:
Gotta merge to finish the migration

PR #2290, opened 2022-07-16
fix(vagrant) ensure that LVM volumes are created without requiring udev 

dduportal:
This PRs allows the `lvm` module to be usable with the Vagrant Docker boxes.

The main issue comes from the fact that `udev` does not workin inside a container (ref. https://lwn.net/Articles/676831/).
Not having `udev` means that a lot of the LVM commands fails while expecting it (udev device filter rules, zeroing volumes, etc.).

Since the `lvm` module that we use is:

. Not recent (0.3.2 version)
. Hard to hack

I've set up this method that defines custom `exec` resources to run a set of shell commands (with idempotency to avoid re)fromatting on each provisionning of the container).
The scope is only the profile `vagrant`: all the "dummy devices set up to loopbacks" had been removed from production profiles and moved into the vagrant profile.

The following "nifty tricks" had been used to ensure it's successfull:

- When creating a logical volume, the flag `--zero n` is provided to disable the blocks zero-ing, removing the need for `udev` for this step
  - Since the logical volumes are not cleared due to this flag, the addition command `vgscan -mknodes` is used to initialize the volume groups 's nodes "as if" they were zero-ed

- The "dummy devices" which are mapped to the `/dev/loopXX` virtual loopbacks had been renamed to `/dev/dummyXX` to support multiples instances and follow usual device paths

- The size of the dummy devices is calculated by parsing the hieradata, calculating the total specified size for the LVs and divided by the amount of specified devices (plus a margin of 1M, the size of a block)

### DIFF
--- a/dist/profile/manifests/census.pp
+++ b/dist/profile/manifests/census.pp
@@ -13,16 +13,6 @@ class profile::census (
 
   $docroot = "${home_dir}/census"
 
-  if str2bool($facts['vagrant']) {
-    # during serverspec test, fake /dev/xvdb by a loopback device
-    exec { 'create /tmp/xvdb':
-      command => 'dd if=/dev/zero of=/tmp/xvdb bs=1M count=16; losetup /dev/loop0; losetup /dev/loop0 /tmp/xvdb',
-      unless  => 'test -f /tmp/xvdb',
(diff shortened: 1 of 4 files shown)


lemeurherve (approved):
👏

PR #2296, opened 2022-07-21
legacy fqdn in a variable to enable hiera overriding

smerle33:
in order to solve https://github.com/jenkins-infra/helpdesk/issues/2649
add the legacy fqdn as a variable and used hieradata to allow overriding of fqdn and legacy fdqn to allow temporary domain for test purpose like : oracle.updates.jenkins.io


smerle33:
For the vagrant check, the files within hieradata need to have the same name as the role to be taken in account so I had to rename oracle.updates.jenkins.io.yaml to updates.yaml but I kept oracle.updates.jenkins.io.yaml for the actual profile of the new VM.

smerle33:
the oracle.updates subdomain had been defined : 
Capture d’écran 2022-07-21 à 12 06 07

smerle33:
to be fully functional, the vagrant VM should be able to use eyaml and get a temporary access to the key in order to decrypt the ssl key cert and chain to provide within apache2 and then simulate correctly the startup (right now, apache2 start is locked if the certificates used are encrypted).
but this seems overkill just for vagrant ...

dduportal:
Closing as outdated

PR #2321, opened 2022-08-11
fix(jenkins::agent) install adoptium JDKs 8,11 and 17 to /opt

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3090

This PR ensures that any machine using the profile `jenkins::agent` (only 1: the `trusted-agent-1`) has the 3 major adoptium JDKs installed.

It uses the version from the common hieradata: any update on the JDK picked will also be installed on agents.

Side cleanups:

- Removed ruby management on `jenkins::agent` (not required anymore)
- Removed invalid hieradata from vagrant area


MarkEWaite:
I'm very nervous about this change.  I thought that there were still jobs on ci.jenkins.io that required an agent with the `ruby` label.  Does this remove the ruby label?

Have the job definitions been searched to check for jobs that depend on the `ruby` label?

I know that the check-agent-availability job uses the `ruby` label.  It is easy enough to fix that job to not require the ruby label, but the `ruby` label was added to that job because that label was available and believed to be in use.

dduportal:
> I'm very nervous about this change. I thought that there were still jobs on ci.jenkins.io that required an agent with the `ruby` label. Does this remove the ruby label?
> 
> Have the job definitions been searched to check for jobs that depend on the `ruby` label?
> 
> I know that the check-agent-availability job uses the `ruby` label. It is easy enough to fix that job to not require the ruby label, but the `ruby` label was added to that job because that label was available and believed to be in use.

No worries:

- Only the trusted-agent-1 permanent agent is controller by this profile. It is not using ruby at all, and is only connected to trusted.ci.jenkins.io.
- The only ruby usage was infra's puppet test, which has been migrated to infra.ci.jenkins.io (on VM agents because Docker is required)
- On ci.jenkins.io , the ruby usage is only through Kubernetes container templates: this change does not remove it, neither does it remove the `ruby` label
- About the check-agent-availability, I plan to update it tomorrow after removing the last ruby container template

MarkEWaite:
Thanks!

PR #2328, opened 2022-08-13
Switch s390x agent to exclusive

timja:
The SSH connection is timing out but node requests that don't specify a label try use this agent first, I had an ATH build that waited ages for this.

The PPC agent already uses the exclusive mode

(note: already manually changed)


dduportal (approved):
LGTM thanks!

PR #2333, opened 2022-08-17
feat(jenkinscontrollers) refactor profile to only use hieradata to specify JCasc configuration

dduportal:
Now that we have Hiera v5 #2325 , we can start deep-merging hash and arrays.

This PR utilizes this ability to move ALL the JCasc configuration for ci.jenkins.io, trusted.jenkins and cert-ci.jenkins.io into hieradata only (instead of spreading between puppet code, hieradata yaml hierachy and templates.

Most of the settings are moving to hashmap (instead of array when having a list) to allow overriding "common sane defaults".

It also factorizes the Adoptium (temurin) download URL generation and share it with the `jenkins::buildagent` profile (including future support for non linux / amd64 platforms.

Please note the following improvements:

- All the JCasc configuration has been moved under the `profile::jenkinscontroller:: jcasc` key to avoid spreading configuration items
- Any "default" JCasc template can be opt-out by adding `disabled: true` as a sub-key of its `jcasc.xxx` setup
- Simplified the JDK tools template configuration: more installers are supported (adding "comand" and "batchfile"), automatic URL generation, ability to override the "default" version
- Add "generic" tool ability (used for cert-ci.j for now
- Ability to customize pipeline-libraries
- Defined the following common default to ensure that ci.j and trusted.ci have the same "tooling):
  - Global library enabled with https://github.com/jenkins-infra/pipeline-library.git 
  - JDK8, 11 and 17 defined with zip installed using default "latest" versions (auto-updated when a new release of teumrin comes) for linux x64 and Windows x64 only
  - A default Maven installation (with latest version bla bla updated bla bla) named `mvn`
- Add the ability to specify custom permanent agents through hieradata and applied to ci.j and trusted.j
- Moved the "trusted-ci" specific labels to its own hierdata configuration
- Updated the cert-ci configuration to disable the following inherited behaviors:
  - No jdk17
  - No "default" mvn installation
  - Overrides the defaults jdk8 and jdk11
  - Added jdk-8 and jdk-11, with the exact same config as jdk8 and jdk11 (yaml anchors)
  - Added generci tool config for groovy
  - No shared library
- Updated the tools:
  - The "subdir" pattern is prefixed by `jdk-`
  - Removed the (unused :) ) `static` JDK tools
  - Fix Adoptium URLs for JDK17 (missing `U` in production today)

PR #2335, opened 2022-08-18
fix(jenkinscontroller) correct default subdir to fit adoptium archives naming + allow to override through hieradata for edge cases

dduportal:
This PR amends #2333 to correct the automatic configuration of the `subdir` of the JDK tools to use the correct adoptium conventions (which are ... exotic).

It also adds the ability to customize the value of "subdir" per JDK tool installer in hieradata.

Once merged (and deployed), this PR shoyld fix the failures for builds of [jenkins-infra/update_center2 on ci.jenkins.io](https://github.com/jenkins-infra/update-center2/pull/642#issuecomment-1219361972); but also the crawler on trusted.ci.jenkins.io, and many more that we have not seen yet.


lemeurherve (approved):
👏

PR #2337, opened 2022-08-18
amends 1e932b97 with the correct jdk values

dduportal:
Signed-off-by: Damien Duportal <damien.duportal@gmail.com>


dduportal:
Tested in production: https://ci.jenkins.io/job/Infra/job/update-center2/job/master/539/console

PR #2355, opened 2022-08-29
cleanup (ci.jenkins.io): remove jnlp-ruby agent

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3111
remove this unused agent from ci.jenkins.io


lemeurherve:
Not sure at all but I'm wondering if we shouldn't also get rid of the ruby script https://github.com/jenkins-infra/jenkins-infra/blob/cdc715088a1fddf3f55df2d3397b096d96b15928/dist/profile/files/staticsite/deploy-site

smerle33:
> Not sure at all but I'm wondering if we shouldn't also get rid of the ruby script https://github.com/jenkins-infra/jenkins-infra/blob/cdc715088a1fddf3f55df2d3397b096d96b15928/dist/profile/files/staticsite/deploy-site

I would create an issue for this to check with @dduportal  and all if it is a good idea, and in the meantime I would merge this one ... WDYT ?

lemeurherve (approved):

PR #2364, opened 2022-09-01
fix DO: replace plus by dot as its forbidden

smerle33:


lemeurherve:
Here is the corresponding error message:

> spec.containers[0].env[9].name: Invalid value: "PATH+JAVA_HOME": a valid environment variable name must consist of alphabetic characters, digits, '_', '-', or '.', and must not start with a digit (e.g. 'my.env-name',  or 'MY_ENV.NAME',  or 'MyEnvName1', regex used for validation is '[-._a-zA-Z][-._a-zA-Z0-9]*'),

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
```diff
@@ -210,7 +210,7 @@ jenkins:
                   key: "JAVA_HOME"
                   value: "<%= agent['javaHome'] %>"
               - envVar:
-                  key: "PATH+JAVA_HOME"
+                  key: "PATH.JAVA_HOME"
```
lemeurherve, line 213:
```suggestion
                  key: "PATH_JAVA_HOME"
```

lemeurherve (approved):

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
smerle33: The idea is to avoid the confusion with the _ of java_home, it's probably not the definitive solution but at least the dot is accepted by the regexp that make the check : 
```
Invalid value: "PATH+JAVA_HOME": a valid environment variable name must consist of alphabetic characters, digits, '_', '-', or '.', and must not start with a digit (e.g. 'my.env-name',  or 'MY_ENV.NAME',  or 'MyEnvName1', regex used for validation is '[-._a-zA-Z][-._a-zA-Z0-9]*')
```

daniel-beck:
Are you relying on the magic behavior of `VARNAME+` in Jenkins, or why are you setting this env var? If so, this change would break it.

smerle33:
> Are you relying on the magic behavior of `VARNAME+` in Jenkins, or why are you setting this env var? If so, this change would break it.

yes I think it's the idea but the + got rejected for now so I plan to change by "." while waiting to find a good way to go (and maybee some help) to at least check that the agent is spawning.

daniel-beck:
Right, but that means this value will no longer be merged into `PATH`, so ultimately this won't have any effect.

smerle33:
> Right, but that means this value will no longer be merged into `PATH`, so ultimately this won't have any effect.

which is already the case with the error ... I need to find the correct way later ... 

for the record how the + is used by jenkins : https://github.com/jenkinsci/jenkins/blob/01e637ec08fddc9a032837ebd838cccdefe4da83/core/src/main/java/hudson/EnvVars.java#L66-L71

PR #2365, opened 2022-09-01
DO agent (test allinone): fix pod path try with the full path

smerle33:
as the magic trick https://github.com/jenkinsci/jenkins/blob/01e637ec08fddc9a032837ebd838cccdefe4da83/core/src/main/java/hudson/EnvVars.java#L66-L71 is using a + and it's rejected here, I temporarily try with a full path variable (not incremental).

as per https://github.com/jenkins-infra/helpdesk/issues/2980


lemeurherve (approved):
🤷

PR #2367, opened 2022-09-02
chore(ci.j.io) : path environment variable mixing case

smerle33:
not sure but worth a try : 
```
For a pod running on Linux, defaults to sh, which should be in $PATH; for a pod running on Windows, defaults to cmd, which should be in %Path%. Should not generally be overridden.
```

https://www.jenkins.io/doc/pipeline/steps/kubernetes/


lemeurherve:
Why this PR?

smerle33:
> 

https://github.com/jenkins-infra/jenkins-infra/pull/2367#issue-1360169567

lemeurherve:
I don't see why you want to try to get a Windows env var name while the value content is for Linux (?)

smerle33:
> 

After hours searching everywhere, I felt like it could be a good try ....

lemeurherve (approved):
🙈

PR #2389, opened 2022-09-15
hotfix(jenkinscontroller) force locale to C.UTF-8 for linux kubernetes containers

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3130


jglick:
Why this is not the default for all Linux flavors and images already, I cannot fathom.

PR #2406, opened 2022-09-30
fix(windows ec2 vm): hack to avoid exit on error with scp

smerle33:
as we use ssh for windows, it crash on scp because of single slash.

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -194,7 +194,8 @@ profile::jenkinscontroller::jcasc:
   agents_setup:
     windows:
       agentDir: 'C:/Jenkins'
-      tempDir: 'C:\\Temp'
+      # Double Backslash is required (for EC2 plugin as we hackishly use the unix launched for Windows to use OpenSSH)
+      tempDir: 'C:\\\\Temp'
       remoteAdmin: Administrator
       osDiskSize: 128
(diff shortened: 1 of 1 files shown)


hieradata/common.yaml
```diff
@@ -194,7 +194,7 @@ profile::jenkinscontroller::jcasc:
   agents_setup:
     windows:
       agentDir: 'C:/Jenkins'
-      tempDir: 'C:/temp'
+      tempDir: 'C:\\Temp'
```
lemeurherve: I would add a comment above explaining why anti-slashes are used here

hieradata/common.yaml
```diff
@@ -194,7 +194,8 @@ profile::jenkinscontroller::jcasc:
   agents_setup:
     windows:
       agentDir: 'C:/Jenkins'
-      tempDir: 'C:/temp'
+      # Double Backslash is required (for EC2 plugin as we hackishly use the unix launched for Windows to use OpenSSH)
+      tempDir: 'C:\\\\Temp'
```
dduportal, line 198: For info @smerle33: with only 2 backslashes, you got the string `C:\Temp` because JCasc interpolates a double backslash into a single. So I've set it to 4 backslashes.

Tested locally

lemeurherve (approved):

PR #2413, opened 2022-10-07
feat(windows datadog): adding a windows initscript as initscript OR userData 

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3149 
this should provide 2 windows agents one on azure, and one on aws/ec2 that should launch the datadog agent at startup through cloudinit


smerle33:
after discussion with @lemeurherve we decide to avoid anchors that make the code harder to read

lemeurherve (approved):

PR #2544, opened 2022-12-14
fix(private-vpn): don't include 99-disable-network-config.cfg if there are only 2 network interfaces

lemeurherve:
Including this cloud-init config seems to break connectivity on Ubuntu Jammy VM like the private.vpn.jenkins.io one.

### DIFF
--- a/dist/profile/manifests/openvpn.pp
+++ b/dist/profile/manifests/openvpn.pp
@@ -62,16 +62,18 @@ class profile::openvpn (
   }
 
   # Ensure cloud-init doesn't manage network to ensure the order of eth1 and eth2 if this last one is defined (ie 3 interfaces) (netplan config + netplan apply + systemd, in Ubuntu Bionic)
+  $content = ''
   if $networks.length > 2 {
-    file { '/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg':
-      ensure  => 'file',
-      owner   => 'root',
-      group   => 'root',
-      require => [
-        File['/etc/cloud/cloud.cfg.d'],
-      ],
-      content => 'network: {config: disabled}',
-    }
+    $content = 'network: {config: disabled}'
+  }
+  file { '/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg':
+    ensure  => 'file',
+    owner   => 'root',
+    group   => 'root',
+    require => [
+      File['/etc/cloud/cloud.cfg.d'],
+    ],
+    content => $content,
   }
 
   file { '/etc/netplan/':
--- a/dist/profile/manifests/openvpn.pp
+++ b/dist/profile/manifests/openvpn.pp
@@ -61,15 +61,17 @@ class profile::openvpn (
     ],
   }
 
-  # Ensure cloud-init doesn't manage network (netplan config + netplan apply + systemd, in Ubuntu Bionic)
-  file { '/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg':
-    ensure  => 'file',
-    owner   => 'root',
-    group   => 'root',
-    require => [
-      File['/etc/cloud/cloud.cfg.d'],
-    ],
-    content => 'network: {config: disabled}',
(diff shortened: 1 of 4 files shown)


dist/profile/manifests/openvpn.pp
```diff
@@ -61,15 +61,19 @@
     ],
   }
 
-  # Ensure cloud-init doesn't manage network (netplan config + netplan apply + systemd, in Ubuntu Bionic)
+  # Ensure cloud-init doesn't manage network to ensure the order of eth1 and eth2 if this last one is defined (ie 3 interfaces) (netplan config + netplan apply + systemd, in Ubuntu Bionic)
+  $content = ''
+  if $networks.length > 2 {
+    $content = 'network: {config: disabled}'
```
dduportal: WDYT putting the `file { ...}` resource block inside the `if  { ... }` block instead? That would ensure that no file is created at all on `private.vpn`.

dduportal (changes requested):
this PR looks good, but it require to add rspec unit test (you can specify custom hieradata in `hieradata/rspec/...` for instance) to check the 2 cases:

- network more than 2
- networ with only 2

spec/classes/profile/openvpn_spec.rb
```diff
@@ -29,3 +29,34 @@
   it { expect(subject).to contain_firewall "100 allow routing from 172.19.0.0/24 to 10.0.0.0/16 on ports 80/443" }
   it { expect(subject).to contain_exec "addroute 10.0.0.0 through 192.168.100.1 (NIC eth1)" }
 end
+
+
+describe 'profile::openvpn' do
```
dduportal:
```suggestion
describe 'profile::openvpn with 2 interfaces' do
```

spec/classes/profile/openvpn_spec.rb
```diff
@@ -29,3 +29,34 @@
...
+      :rspec_hieradata_fixture => 'profile_openvpn_two_interfaces',
+    }
+  end
+
+  it { expect(subject).to contain_class 'stdlib' }
+  it { expect(subject).to contain_class 'profile::docker' }
+
+  it { expect(subject).to contain_package 'net-tools' }
+
+  it { expect(subject).to contain_file '/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg' }
```
dduportal:
```suggestion
  it { expect(subject).not_to contain_file '/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg' }
```

as per https://github.com/rspec/rspec-expectations/tree/v3.11.0#equivalence

lemeurherve: `.not_to` fails (as `should_not`):

```
Failures:

  1) profile::openvpn with 2 interfaces is expected not to contain File[/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg]
     Failure/Error: it { expect(subject).not_to contain_file '/etc/cloud/cloud.cfg.d/99-disable-network-config.cfg' }
     
     Puppet::ParseErrorWithIssue:
       Could not parse for environment rp_env: This Name has no effect. A value was produced and then forgotten (one or more preceding expressions may have the wrong form) (line: 115, column: 26) on node macbook-pro-de-herve.local
```

hieradata/rspec/profile_openvpn.yaml
```diff
@@ -10,6 +10,11 @@ profile::openvpn::networks:
     network_cidr: 192.168.100.0/24
     peered_network_cidrs:
       - 10.0.0.0/16
+  eth2:
+    name: Another Network
+    network_cidr: 192.168.200.0/24
+    peered_network_cidrs:
+      - 11.0.0.0/16
```
lemeurherve, line 17: Note: changed from `10.0.0.0/16` to fix this error:

> Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Firewall[100 allow routing from 127.0.10.0/24 to 10.0.0.0/16 on ports 80/443] is already declared at (file: ./spec/fixtures/modules/profile/manifests/openvpn.pp, line: 156); cannot redeclare (file: ./spec/fixtures/modules/profile/manifests/openvpn.pp, line: 156) (file: ./spec/fixtures/modules/profile/manifests/openvpn.pp, line: 156, column: 9)

PR #2559, opened 2022-12-21
Fix(openvpn) fix private

dduportal:
This PR fixes the private.vpn.jenkins.io new VM to allow accessing the new [privatek8s cluster](https://github.com/jenkins-infra/helpdesk/issues/2844) through the "private.vpn" service.

- fix the behavior of the openvpn profile to make sure that network interface setup is fully delegated to cloud init when using only 2 interfaces
- change the private.vpn network to 10.9.0.0/24 (new one) to ease routing diagnostics
- remove unneeded routes (to the eth0 -main- network) as not needed

⚠️ this PR fixes the "breaking reboot" of the VM (rebooting the private.vpn.jenkins.io VM before this change always ends up in the VM unreachable because of the cloudinit configuration competing with ours, leading to "offline" network interfaces

### DIFF
--- a/dist/profile/manifests/openvpn.pp
+++ b/dist/profile/manifests/openvpn.pp
@@ -64,8 +64,10 @@ class profile::openvpn (
     ],
   }
 
-  # Ensure cloud-init doesn't manage network to ensure the order of eth1 and eth2 if this last one is defined (ie 3 interfaces) (netplan config + netplan apply + systemd, in Ubuntu Bionic)
+  # When the VM has more than 2 interfaces (1 external and 1 internal) then it's a more complex setup where network interfaces
+  # must be managed by us instead of the usual cloud-init netplan generated configuration to keep the interface order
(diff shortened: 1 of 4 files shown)


lemeurherve (dismissed):

dduportal:
Unit test failing : i did not use tdd on it so now i have to fix it (hence the draft status)

lemeurherve (approved):

dduportal:
Diff in production for (legacy) `vpn.jenkins.io`:

```shell
$ puppet agent --enable && puppet agent --test --noop && puppet agent --disable "$USER is doing some manual maintenance"
Info: Using environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Applying configuration version '1671699460'
Notice: Applied catalog in 4.82 seconds
```

=> no changes as expected

PR #2576, opened 2023-01-12
chore(updatecli) : add conditions with tags for jenkins controller agents

smerle33:
This will avoid to launch updatecli PR if one of the source is not yet in the docker hub or aws ami store.


dduportal (approved):
LGTM (let's test for real on the `production` branch because of the AWS credentials not available on PRs, which does not allow conditions to be run as sources are failing)

PR #2662, opened 2023-02-24
hotfix(letsencrypt) ensure that pip list uses modern formatting for idempotency

dduportal:
Fixup of https://github.com/jenkins-infra/jenkins-infra/pull/2659

This PR ensures the `pip list` command does not error when piped to a non interactive process.
It adds the flag `--format=columns` to solve this issue.

PR #2664, opened 2023-02-24
fixup of #2662

dduportal:
- Tested in production
- `pip list` complains. Instead of an ugly `pip list | strings | grep`, using the `json` looks better has it is explicit for "processing" (I mean reading JSON is failing the turing test :trollface: )


lemeurherve (approved):
😛

PR #2667, opened 2023-02-27
hotfix(pkg,updates,tests) ensure that apache::mod::event custom setup uses the correct type

dduportal:
This PR is a fixup for #2538.

- The version [8.0.0](https://forge.puppet.com/modules/puppetlabs/apache/changelog#v800-2022-08-09) of the module adds a lot more of type checking.
- Since we only provide these values as pure hieradata for production setup, the unit tests cannot catch this typing error: we added a unit test (but the downside is duplicating the hieradata value).


smerle33 (dismissed):
🚀

smerle33 (approved):
🚀

PR #2668, opened 2023-02-27
fix(letsencrypt) ensure that the correct python version is used by default for the system (to avoid crashing apt)

dduportal:
While working on, we discovered that bionic install were broken: apt needs python3.6 while we force python 3.8 as default.

This PR fixes this issue :)


smerle33 (approved):
🚀  nice one

PR #2674, opened 2023-03-02
fix(artifact-caching-proxy): mirror all HTTP maven repositories

lemeurherve:
Fix the *"Blocked mirror for repositories: [maven.jenkins-ci.org (http://repo.jenkins-ci.org/public/, default, releases+snapshots)]"* error noticed in https://github.com/jenkins-infra/backend-extension-indexer/pull/48

Ref: https://github.com/jenkins-infra/helpdesk/issues/2752

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/artifact-caching-proxy.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/artifact-caching-proxy.yaml.erb
@@ -11,7 +11,7 @@ unclassified:
               <mirror>
                   <id><%= providerId %>-proxy</id>
                   <url>https://repo.<%= providerId %>.jenkins.io/public/</url>
-                  <mirrorOf>*,!incrementals,!elementary-releases,!chimera-releases,!chimera-snapshots</mirrorOf>
+                  <mirrorOf>*,external:http:*,!incrementals,!elementary-releases,!chimera-releases,!chimera-snapshots</mirrorOf>
               </mirror>
               <mirror>
                   <id><%= providerId %>-proxy-incrementals</id>


dduportal:
What is the reason of this change ? Can you explain the rationale either here or in https://github.com/jenkins-infra/backend-extension-indexer/pull/48 ?

The PR here does not explain what was the issue, neither why does it solve the issue (and the impact) ?

lemeurherve:
The "blocked mirror" error mentioned in this PR body and in the backend-extention-indexer is due to the use of an HTTP (not HTTPS) repository.

This PR should have solved this issue by adding this kind of maven repositories (using http instead of https) in the list of repositories to be mirrored in the Maven settings we're using with artifact caching proxy.

I've tested it locally and it fixed it, but the last log here seems to indicate it didn't work unfortunately: https://infra.ci.jenkins.io/job/reports/job/backend-extension-indexer/job/PR-48/7/consoleFull

lemeurherve:
> but the last log here seems to indicate it didn't work unfortunately: [infra.ci.jenkins.io/job/reports/job/backend-extension-indexer/job/PR-48/7/consoleFull](https://infra.ci.jenkins.io/job/reports/job/backend-extension-indexer/job/PR-48/7/consoleFull)

Oh, I made sure ci.jenkins.io settings.xml files were updated but not infra.ci.jenkins.io ones 😓

lemeurherve:
I realize they aren't managed as code at all on infra.ci.jenkins.io...

dduportal:
> The "blocked mirror" error mentioned in this PR body and in the backend-extention-indexer is due to the use of an HTTP (not HTTPS) repository.
> 
> This PR should have solved this issue by adding this kind of maven repositories (using http instead of https) in the list of repositories to be mirrored in the Maven settings we're using with artifact caching proxy.
> 
> I've tested it locally and it fixed it, but the last log here seems to indicate it didn't work unfortunately: https://infra.ci.jenkins.io/job/reports/job/backend-extension-indexer/job/PR-48/7/consoleFull

Oh, that make sense, thanks for the explanation! I guess the next step will be to mirror these "unsecured" maven repos in repo.jenkins-ci.org to get rid of the http allowance, but not priority for now 👍

dduportal:
> I realize they aren't managed as code at all on infra.ci.jenkins.io...

Good catch !

lemeurherve:
> > I realize they aren't managed as code at all on infra.ci.jenkins.io...
> 
> Good catch !

Good news, they are managed as code, just not up to date 😛 PR incoming, I'll revert this one as it seems unneeded.

PR #2659, opened 2023-02-24
fix(letsencrypt) use pip package for certbot and its plugins

dduportal:
This PR is related to https://github.com/jenkins-infra/helpdesk/issues/3399.

It fixes the certbot installation by using the Python pip package instead of snap, which is the only way to use the dns azure plugin for certbot which requires certbot 1.x as for today (see the issue above for details).

- Python 3.8 is required in order to properly install `certbot` (to ensure the certbot requirement of pyopenssl >= 40.x )
- Since Python 3.6 is the "default" installed version on Ubuntu 18.04, this script ensures that python 3.8 is installed and defined as the "alternative" for both `python` and `python3`.
  - For Ubuntu 20.04, Python 3.8 is the system default ✅ 
  - For Ubuntu 22.04: Python 3.10 is used ✅ 
- Pip is "prepared" in the context of python 3.8 to ensure that  `certbot` installation deos not fail:
  - Upgraded to the last version available when executed 
  - Add requirements for proper pyopenssl installation: `setuptools` and setuptools_rust`
- `certbot` is installed in version 1.32.0: it's the latest available version for the 1.x line in PyPip:  https://pypi.org/project/certbot/#history
- The plugin `certbot-dns-azure` is always installed (less conditional code + we can switch between HTTP-01 and DNS-01 without surprise)

In order to avoid any problem once this PR merged, the following preventive actions were taken:

- Backed-up any letsencrypt folder with certificates: if anything goes wrong on a machine, we can rollback the "current" certificate and keep going

```shell
for machine in ci.jenkins.io cert.ci.jenkins.io archives.jenkins.io edamame.jenkins.io lettuce.jenkins.io census.jenkins.io usage.jenkins.io trusted.ci.jenkins.io trusted-agent-1 vpn.jenkins.io private.vpn.jenkins.io bounce puppet.jenkins.io; do echo "=== Machine: ${machine}";ssh "${machine}" sudo cp -r /etc/letsencrypt /root/bkp-letsencrypt-20230224-01;done 
```

- Disabled the Puppet agent on each machine (to control, machine per machine, the new execution flow):

```shell
[...]

PR #2676, opened 2023-03-06
fix(letsencrypt,pkg) ensure that createrepo is usable

dduportal:
This PR is a fixup of https://github.com/jenkins-infra/jenkins-infra/pull/2659.

It ensures that the `createrepo` command line works as expected on the VM pkg.origin.jenkins.io in order to package Jenkins core when releasing a weekly or LTS.

https://github.com/jenkins-infra/jenkins-infra/pull/2659 did change the symlink `/usr/bin/python`from `python2.7` to `python3.x` using the `update-alternative.

This behavior is fixed here but only on pkg.origin.jenkins.io + removed from Let's Encrypt profiles to avoid the issue to reappear due to certbot (which needs `python3.8`).

Additionnaly, as `createrepo(8)` does not exist on Ubuntu 20.04 and later, the profile fails if not executed on Ubuntu 18.04. This is a safety measure to ensure that we'll fail fast (and we'll have to fix this) once we'll upgrade out from Ubuntu 18.04.
(NB. https://github.com/jenkins-infra/release/issues/202 and https://github.com/jenkinsci/packaging/pull/317 were tentative to move to `createrepo-c` instead, but wasn't done since it's still on pkg.origin.jenkins.io)

### DIFF
--- a/dist/profile/manifests/pkgrepo.pp
+++ b/dist/profile/manifests/pkgrepo.pp
@@ -14,7 +14,7 @@ class profile::pkgrepo (
   include profile::firewall
   include profile::letsencrypt
 
-  # 'Ubuntu 20.04+ are not supported (createrepo package absent).'
+  # Ubuntu 20.04+ are not supported (createrepo package absent).
   case $facts['os']['distro']['codename'] {
     'bionic': {
       # Needed so we can generate repodata on the machine
@@ -31,7 +31,7 @@ class profile::pkgrepo (
         unless  => '/usr/bin/python --version 2>&1 | grep --quiet "2\.7\."',
       }
     }
-    # TODO: add support of Ubuntu 22.04 Jammy with the createrepo-c package instead
+    # TODO: add support of Ubuntu 22.04 Jammy with the createrepo-c package instead. Ref. https://github.com/jenkins-infra/release/issues/202.
     default: {
       fail('[profile::pkgrepo] Unsupported Ubuntu distribution (ref. "createrepo" package).')
     }
--- a/dist/profile/manifests/pkgrepo.pp
+++ b/dist/profile/manifests/pkgrepo.pp
@@ -14,9 +14,27 @@ class profile::pkgrepo (
   include profile::firewall
   include profile::letsencrypt
 
-  # Needed so we can generate repodata on the machine
-  package { 'createrepo':
-    ensure => present,
+  # 'Ubuntu 20.04+ are not supported (createrepo package absent).'
+  case $facts['os']['distro']['codename'] {
+    'bionic': {
+      # Needed so we can generate repodata on the machine
+      ['createrepo','python2.7'].each |$pkg| {
+        package { $pkg:
+          ensure => present,
+        }
+      }
+
+      exec { 'Define python2.7 as the default system python':
+        require => Package['python2.7'],
+        # Last argument is the weight. Bigger weight wins
+        command => '/usr/bin/update-alternatives --install /usr/bin/python python /usr/bin/python2.7 1000',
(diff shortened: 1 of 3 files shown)


spec/classes/profile/pkgrepo_spec.rb
```diff
@@ -19,12 +19,32 @@
...
-    })
+  context 'Ubuntu 18.04 Bionic' do
+    # let(:facts) {
+    #   :os => {
+    #     :distro => {
+    #       :codename => 'bionic'
+    #     },
+    #   },
+    # }
+
```
dduportal:
```suggestion
```

dist/profile/manifests/pkgrepo.pp
```diff
@@ -14,9 +14,27 @@
   include profile::firewall
   include profile::letsencrypt
 
-  # Needed so we can generate repodata on the machine
-  package { 'createrepo':
-    ensure => present,
+  # 'Ubuntu 20.04+ are not supported (createrepo package absent).'
```
lemeurherve:
```suggestion
  # Ubuntu 20.04+ are not supported (createrepo package absent).
```
(nit)

dist/profile/manifests/pkgrepo.pp
```diff
@@ -14,9 +14,27 @@
...
+      }
+
+      exec { 'Define python2.7 as the default system python':
+        require => Package['python2.7'],
+        # Last argument is the weight. Bigger weight wins
+        command => '/usr/bin/update-alternatives --install /usr/bin/python python /usr/bin/python2.7 1000',
+        unless  => '/usr/bin/python --version 2>&1 | grep --quiet "2\.7\."',
+      }
+    }
+    # TODO: add support of Ubuntu 22.04 Jammy with the createrepo-c package instead
```
lemeurherve: Cf comment line 17 above, what about Ubuntu 20.04? Should it be mentioned in this comment too?

spec/classes/profile/pkgrepo_spec.rb
```diff
@@ -19,12 +19,24 @@
...
+      })
+
+      expect(subject).to contain_exec('Define python2.7 as the default system python').with({
+        :command => '/usr/bin/update-alternatives --install /usr/bin/python python /usr/bin/python2.7 1000',
+        :unless  => '/usr/bin/python --version 2>&1 | grep --quiet "2\.7\."',
+      })
+    end
   end
 
+
```
lemeurherve:
```suggestion
```

lemeurherve (dismissed):

dist/profile/manifests/pkgrepo.pp
dduportal: The comment here is about `createrepo-c` which is only available starting from Ubuntu 22.04.
The comment on L17 is about the "currently" used package `createrepo` which is only available on Ubuntu 18.04.

Ubuntu 20.04 is basically dead to us for this use case.

dduportal:
```suggestion
    # TODO: add support of Ubuntu 22.04 Jammy with the createrepo-c package instead. Ref. https://github.com/jenkins-infra/release/issues/202.
```

lemeurherve (approved):

smerle33 (approved):
🚀

dduportal:
- Puppet agent disabled on all machines as a safety measure (gotta run a `noop` on each and report here)

dduportal:
- No changes detected when applied to the virtual machines ci.jenkins.i, cert.ci.jenkins.io, archives.jenkins.io and pkg.origin.jenkins.io

PR #2679, opened 2023-03-07
feat(ci.jenkins.io) Increase Linux container agents capacity (150 in DigitalOcean)

lemeurherve:
As the droplets limit on DigitalOcean account has been increased to 100, we can now increase this cluster capacity to be the same as the AWS one.

Follow-up of #2655


smerle33 (approved):
🚀

PR #2709, opened 2023-03-23
fix(cert.ci): add missing `docker` labels for Azure VM agents

lemeurherve:


hieradata/clients/cert-ci.yaml
```diff
@@ -98,6 +99,7 @@ profile::jenkinscontroller::jcasc:
           architecture: amd64
           labels:
             - windows
+            - docker
```
dduportal:
```suggestion
            - docker-windows
```

Otherwise it will a good recipe for failure while trying to spin up docker linux container on a docker windows engine

dduportal (changes requested):

dduportal (approved):

PR #2711, opened 2023-03-24
fix(build_agent) change GitHub RSA SSH host key

dduportal:
Follow up of https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/

Cc @daniel-beck @wadeck @MarkEWaite .

Thanks @lemeurherve for catching this!

### DIFF
--- a/dist/profile/manifests/buildagent.pp
+++ b/dist/profile/manifests/buildagent.pp
@@ -109,7 +109,7 @@ class profile::buildagent (
     ensure       => present,
     host_aliases => ['github.com'],
     type         => 'ssh-rsa',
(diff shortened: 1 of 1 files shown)


lemeurherve (approved):

MarkEWaite:
Thanks very much.  I saw the message this morning in one of my feeds.  I'm surprised that we don't already have Jenkins issues reported due to the change of RSA host key at github.com.  I'm sure those reports are coming.

The [git client plugin configuration as code sample](https://plugins.jenkins.io/git-client/#plugin-content-configuration-as-code-sample) uses the ED-25519 host key instead of the RSA host key.  The ED-25519 host key is shorter and it is at least as strong as the RSA host key.

I've submitted a draft pull request proposing a change from the RSA host key to the ED-25519 host key for github.com.  See 

* #2714

PR #2714, opened 2023-03-24
feat(build_agent): Use ED-25519 host key for github.com

MarkEWaite:
## Use ED-25519 host key for github.com

The ED-25519 host key is a shorter string and was not rotated by GitHub in their 2023-03-24 host key rotation.  It was not rotated because the ED-25591 private host key was not exposed in a public repository like the RSA private host key was.

Consider this a prototype for discussion of the idea of switching to ED-25519 rather than as a firm claim that this is what we should do.

The Jenkins git client plugin documentation includes the ED-25519 host key for github.com in its configuration as code sample because it is much easier to read and seems to be at least as strong as the RSA key.

https://plugins.jenkins.io/git-client/#plugin-content-configuration-as-code-sample

https://security.stackexchange.com/questions/90077/ssh-key-ed25519-vs-rsa and https://news.ycombinator.com/item?id=12575358 compare ED25519 and RSA keys for those who might be interested.


dduportal (approved):
LGTM thanks!

PR #2717, opened 2023-03-24
fix(pkg) track mirror-scripts requirements.txt packages (with their versions)

dduportal:
Fixup of #2710 which did not check for the pip package version status (if the file is updated)


dduportal:
Puppet agent disabled on the machine

PR #2730, opened 2023-03-31
chore(updatecli update Maven version in rspec too

lemeurherve:
Ref: https://github.com/jenkins-infra/jenkins-infra/pull/2729#issuecomment-1491891419

### DIFF
--- a/updatecli/weekly.d/jenkinscontroller-tools-maven.yaml
+++ b/updatecli/weekly.d/jenkinscontroller-tools-maven.yaml
@@ -59,7 +59,9 @@ targets:
       - addprefix: /home/jenkins/tools/apache-maven-
       - addsuffix: /bin
     spec:
-      file: hieradata/clients/azure.ci.jenkins.io.yaml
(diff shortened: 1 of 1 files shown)


dduportal (commented):
The rspec dataset is only there to test different cases, and is not mapped to a real life environment.

You could set the version to "bork.bark.birk" it would still pass the unit tests.

This PR is not a problem, but I wonder if it is not making the manifest here more "fragile" to file changes.
Given the intrinsic nature of unit tests (prone to change often and not "real life") that might be a reason to not merge this PR.

On the other side, merging the PR would avoid making the impression of "a file was forgotten".


I can't decide :D

hervelemeur:
As we've already updated this file to keep maven version up to date more than once, I would be in favor of using updatecli to continue doing that.

dduportal:
Merged as it does not do any harm.

But keep in mind that `rspec` is for unit test, where the goal is to test code path, not "reality".

PR #2744, opened 2023-04-04
fix: use emptyDir volumes instead of overlay for `/tmp` and `/home/jenkins/.m2/repository`

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3423


dduportal (dismissed):

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
```diff
@@ -259,6 +259,13 @@ jenkins:
         <%- end -%>
           slaveConnectTimeout: 100
           yamlMergeStrategy: "override"
+          volumes:
+          - emptyDirVolume:
+              memory: false
+              mountPath: "/tmp"
+          - emptyDirVolume:
+              memory: false
+              mountPath: "/home/jenkins/.m2"
```
dduportal:
```suggestion
              mountPath: "/home/jenkins/.m2/repository"
```

To avoid having `settings.xml` with sensitive data to accidentally be persisted on the underlying host.

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
```diff
@@ -259,6 +259,13 @@ jenkins:
         <%- end -%>
           slaveConnectTimeout: 100
           yamlMergeStrategy: "override"
+          volumes:
+          - emptyDirVolume:
+              memory: false
```
dduportal:
```suggestion
              memory: true
```

- `/tmp` in a `tmpfs` should be a sane default 
- The memory used in `/tmp` is controlled by the pod limit so it is safe

dduportal (approved):

dduportal (approved):

PR #2749, opened 2023-04-07
chore: disable az-cli telemetry on every controller

lemeurherve:
https://learn.microsoft.com/en-us/cli/azure/azure-cli-configuration

We might want to configure that when installing az-cli instead.


timja (commented):
Why?

lemeurherve:
I thought it would be a good idea to avoid sending info. Not neeeded.

PR #2763, opened 2023-04-13
feat(datadog) define custom hostname for each VM and enable EC2 tags collection when needed

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3494

This PR defines through Puppet hieradata configuration system, custom datadog agent configurations for our machines:

- Defining a custom hostname to ease our datadog metrics researches (because... `ip-172-17-345-18` is not really useful. And its FQDN `ip-172-17-345-18.ec2.internal` isn't either)
- For the remaining EC2 VMs, enable automatic EC2 tags collection

This PR was tested on a local VM vagrant machine for the presence of both settings in `/etc/datadog_agent/datadog.yaml` configuration file


hieradata/vagrant/common.yaml
```diff
@@ -16,6 +16,9 @@ profile::letsencrypt::dns_azure:
   tenant_id: "<subscription id>" # subscription id
   zones:
     trusted.ci.jenkins.io: "/subscriptions/xxx/dnszone/yyyy"
+# Per-host Datadog configuration
+datadog_agent::host: "vagrant.local"
```
hervelemeur, line 20: Was that necessary?

dduportal: yes:

> This PR was tested on a local VM vagrant machine for the presence of both settings in /etc/datadog_agent/datadog.yaml configuration file

hervelemeur: Oups sorry, I looked only the modified files, I've missed the pr description

dduportal: no problem, I assumed you were on your phone ;)

dduportal:
Post deployment note: the PR #2764 surfaces the following warning message in the datadog-agent logs:

```
(pkg/util/ec2/ec2_tags.go:104 in fetchEc2TagsFromAPI) | unable to get tags using default credentials (falling back to instance role): NoCredentialProviders: no valid providers in chain. Deprecated.
``` 

=> It's caused by the `datadog_agent::collect_ec2_tags: true` of the current PR.
It has to be removed (letting the puppet module `datadog-agent` setting it to its default value `false`).
=> The root cause is because these VMs are not using instance metadata (IMDSv1) and don't need to as they are going to be migrated to Azure soon.

PR #2764, opened 2023-04-13
fix(apachemisc) ensure datadog apache integration works as expected 

dduportal:
The Apache integration for datadog is broken on our VMs.

This PR aims to fix it with the following changes:

- Remove the `datadog_check` profile which creates an empty and invalid file in `/etc/datadog-agent/process.yaml`, which creates unused logs messages and disable some automatic agent setups for datadog
- Remove the `puppetagent` profile as it was only including the `datadog_check` profile
- In the `apachemis` profile, stop using the `datadog_agent::integrations::apache` class because it cannot set up `logs`: https://github.com/DataDog/puppet-datadog-agent/blob/main/templates/agent-conf.d/apache.yaml.erb.
  - Use, instead, a custom static file which was tested on `trusted.ci.jenkins.io` to collect logs.


dduportal:
Puppet agent disabled, let's roll

dduportal:
**Post-deployment notes**

The main change of this PR is:


```
Notice: /Stage[main]/Profile::Apachemisc/File[/etc/datadog-agent/conf.d/apache.d/conf.yaml]/content: 
--- /etc/datadog-agent/conf.d/apache.d/conf.yaml        2023-04-13 17:39:09.992401840 +0000
+++ /tmp/puppet-file20230414-16194-13b9uac      2023-04-14 07:30:45.733889730 +0000
@@ -1,7 +1,16 @@
-### MANAGED BY PUPPET
-
 init_config:
-
 instances:
-    -   apache_status_url: http://localhost/server-status?auto
+  - apache_status_url: http://localhost/server-status?auto
+
+logs:
+  - type: file
+    path: /var/log/apache2/*.log
+    source: apache
+    service: apache
 
+  - type: file
+    path: /var/log/apache2/*/*.log.*
+    exclude_paths:
+      - /var/log/apache2/*/*.gz
+    source: apache
+    service: apache
```

and was delivered with success to all machines with an Apache server ✅ 


Next steps in https://github.com/jenkins-infra/helpdesk/issues/3514#issuecomment-1508173355

PR #2769, opened 2023-04-16
fix(trusted.ci/trusted-agent-1) rotate agent key and refresh configuration

dduportal:
Ref. jenkins-infra/helpdesk#3517 which was caused by https://github.com/jenkins-infra/jenkins-infra/pull/2760

This PR fixes the problem by specifying a new SSH key only for the `trusted.ci.jenkins.io` -> `trusted-agent-1` SSH connection with the user `jenkins`:

- It ensures the "(`cucumber.jenkins.io` was a former VM of the infra many years ago) which was already marked as "to be rotated"
Capture d’écran 2023-04-16 à 09 15 44

- It uses a modern `ed25519` algorithm for the key
- It updates the naming of the `ssh_keys` variable in the profile `buildagent` to avoid confusion between private and public key
- It moves the current private key of the `jenkins` user , defined in Hieradata, to the `trusted-agent-1` scope (not sure if this SSH key is even used)


dduportal:
- Puppet agent disabled on  both `trusted.ci.jenkins.io` and `trusted-agent-1`
- New SSH private key  inserted in trusted.ci.jenkins.io
Capture d’écran 2023-04-16 à 09 18 53
- Tested manually the change: agent connects using the new key

PR #2779, opened 2023-04-21
feat(azure.ci.jenkins.io) add a global build discarder policy

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3495

Tested on a local vagrant VM with 2 cases:

- No `jcasc.unclassified` section: the file is created empty (case for trusted.ci and cert.ci)
- With the hieradata defined for azure.ci.jenkins.io: worked as excpected (no JCasc reload error and the config is present)

### DIFF
--- a/dist/profile/manifests/jenkinscontroller.pp
+++ b/dist/profile/manifests/jenkinscontroller.pp
@@ -174,7 +174,7 @@ class profile::jenkinscontroller (
       'jenkinscontroller/casc/tools.yaml.erb',
       # Opt-out with `profile::jenkinscontroller::jcasc.artifact_caching_proxy: false`
       'jenkinscontroller/casc/artifact-caching-proxy.yaml.erb',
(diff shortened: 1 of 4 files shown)


dduportal:
Puppet agent disabled on ci.jenkins.io

lemeurherve (approved):
I would have named this puppet section "build-discarder" or something similar while keeping "unclassified" in the jcasc

timja (approved):

dduportal:
> I would have named this puppet section "build-discarder" or something similar while keeping "unclassified" in the jcasc

Good point.

It's a bit different than the `controller.JCasc.configs` collection of JCasc snippet in the Jenkins Helm Chart: the puppet profile `jenkins::controller` does not have a "dynamic" templating system, so your proposal would involve adding a list of files in addition of the hieradata definition.

Instead of creating a new feature in this Puppet profile, I chose the compromise of "let's pass through the whole `unclassified` config data since the target is the 3 controllers managed by Puppet .

PR #2782, opened 2023-04-22
feat(ci.jenkins.io) use AWS S3 Artifact Manager to store artifacts and stashes

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3496
Requires https://github.com/jenkins-infra/aws/pull/395

This PR adds the required JCasc configuration for ci.jenkins.io (and the local vagrant VM setup to allow testing) to enable Artifact Manager in S3.

Please note that stash deletion i enabled through the Java option (https://plugins.jenkins.io/artifact-manager-s3/#plugin-content-delete-stash) but not artifact deletion (a custom GC would be preferred).

### DIFF
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -50,6 +50,19 @@ profile::jenkinscontroller::jcasc:
               discarder:
                 logRotator:
                   numToKeepStr: "5"
+  artifact-manager:
+    data: |-
+      aws:
+        awsCredentials:
+          credentialsId: "aws-s3-artifact-manager"
+          region: "us-east-2"
+        s3:
+          container: "ci-jenkins-io-artifacts"  # Resource defined in jenkins-infra/aws/ci.jenkins.io.tf
+          disableSessionToken: false            # For non AWS services only
+          prefix: "ci.jenkins.io/"              # Rule of thumb: Controller's hostname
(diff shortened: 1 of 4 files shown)


lemeurherve (dismissed):
🚀

dduportal:
- Tested locally (on the vagrant setup) with the production S3 buckets (but the prefix `localhost/` to differentiate) and IAM key: success
- Credential created manually (not managed as code) on ci.jenkins.io, with an explicit description which links to the Terraform project, and scope only to controller and nodes

hieradata/clients/azure.ci.jenkins.io.yaml
```diff
@@ -50,6 +50,19 @@ profile::jenkinscontroller::jcasc:
...
                   numToKeepStr: "5"
+  artifact-manager:
+    data: |-
+      aws:
+        awsCredentials:
+          credentialsId: "aws-s3-artifact-manager"
+          region: "us-east-2"
+        s3:
+          container: "ci-jenkins-io-artifacts"  # Resource defined in jenkins-infra/aws/ci.jenkins.io.tf
+          disableSessionToken: false            # For non AWS services only
```
jglick, line 61: Note that some of these are set to the default value and _could_ be omitted.

dduportal: We've got bitten by sudden changes on defaults values in JCasc (for instance `ec2` plugin) so we prefer being explicit 😅

PR #2821, opened 2023-05-09
fix(jenkins::controller) do not use ACP for maven repositories with id `local`

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3567#issuecomment-1539563324

Needs a twin in jenkins-infra/kubernetes-management

This PR adds exlcusions from the ACP configurations file for any Maven repository which id would be `in-project` or `local` to allow loading JAR from source code local files.

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/artifact-caching-proxy.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/artifact-caching-proxy.yaml.erb
@@ -11,7 +11,7 @@ unclassified:
               <mirror>
                   <id><%= providerId %>-proxy</id>
                   <url>https://repo.<%= providerId %>.jenkins.io/public/</url>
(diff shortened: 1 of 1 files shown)


lemeurherve:
Where does `local` comes from?

Can't we keep only one unique id for local/in project maven repository?

dduportal:
> Where does `local` comes from?
> 
> Can't we keep only one unique id for local/in project maven repository?

I proposed `local` as an alternative name because `in-project` is confusing for me.
Also, It does not feel good for me to ask the contributor to change their id in the current pom.xml of their.


Hence the 2 names excluded.

dduportal:
Testing with https://github.com/jenkins-infra/helpdesk/issues/3567#issuecomment-1540465092 that it works as expected with the exception set up manually for `in-project`.

Then let's use only `local` and we'll ask the user to update their pom.xml

lemeurherve (approved):

PR #2828, opened 2023-05-12
feat(ci.jenkins.io) use new instance kind for highmem Linux: `Standard_D8ads_v5`, spot with ephemeral disks

dduportal:
As pointed in https://github.com/jenkins-infra/helpdesk/issues/3551#issuecomment-1545726927, this PR introduces the following configuration change to the "High Memory" instances on ci.jenkins.io:

- Change instance type from `Standard_D8s_v3` to `Standard_D8ads_v5` (ref. https://learn.microsoft.com/en-us/azure/virtual-machines/dasv5-dadsv5-series).
  - Same amount of vCPUs and memory, but CPU is way more powerfull, and new instance has up to 300 Gb of temp storage instead of 200
  - Price is a bit more on-demand ($0.4120 vs $0.3840 before) BUT spot price is way cheaper ($0.0412 vs $0.1695 before).

- Use ephemeral OS disk: wether we stick to old instance type or new one, the local storage or temp storage are able to use 150 Gb of OS disk. It avoid paying $0.05 / hour for a Managed Premium SSD 

- Enable Spot: the new instance type is 9x cheaper and the eviction rate is 0-5 % (it was 10 to 15 % with the DS_v3 serie)
  - Please note that both `hihgmem` pipeline users now uses the `retry()` method with agent failure detection: https://github.com/jenkinsci/jenkins/blob/6cd05ada0dab757c009581643602bf260d17c0ec/Jenkinsfile#L201C1-L202 and https://github.com/jenkinsci/acceptance-test-harness/blob/d8d38af2189ca3beac5502fd9700e59a3337805a/Jenkinsfile#L107C1-L108

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
@@ -19,7 +19,7 @@ jenkins:
         doNotUseMachineIfInitFails: true
         enableMSI: false
         enableUAMI: false
-        ephemeralOSDisk: false
+        ephemeralOSDisk: "<%= agent['ephemeralOSDisk'] ? agent['ephemeralOSDisk'] : false %>"
       <%- if agent['initScript'] && !agent['initScript'].empty? -%>
         executeInitScriptAsRoot: true
         initScript: |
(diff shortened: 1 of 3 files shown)


dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
```diff
@@ -48,7 +48,7 @@ jenkins:
         location: "<%= agent['location'] %>"
         noOfParallelJobs: 1
         osDiskSize: <%= agent['osDiskSize'] ? agent['osDiskSize'] : @jcasc['agents_setup'][agent['os'].to_s]['osDiskSize'] %>
-        osDiskStorageAccountType: <%= agent['osDiskStorageAccountType'] ? agent['osDiskStorageAccountType'] : @jcasc['agents_setup'][agent['os'].to_s]['osDiskStorageAccountType'] %>
+        osDiskStorageAccountType: <%= agent['ephemeralOSDisk'] ? "Standard_LRS" : (agent['osDiskStorageAccountType'] ? agent['osDiskStorageAccountType'] : @jcasc['agents_setup'][agent['os'].to_s]['osDiskStorageAccountType']) %>
```
lemeurherve, line 51: IIUC the use of ephemeral disk implies this Standard_LRS type?

lemeurherve (approved):

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
dduportal: Not from the Azure SDK. I had issue when testing in infra.ci.jenkins.io from the existing config: setting it to `Standard_LRS` is needed other wise it fails deploying the VM in Azure. => hence this code that ensures any user-custom config is overrident (but I did not found how to trigger a notice to puppet logs from the templates :'( )

dduportal:
Verified with success in https://ci.jenkins.io/job/Infra/job/acceptance-tests/job/check-agent-availability/2408/console.

IOPS are stille the same, but no managed disk \o/

PR #2834, opened 2023-05-15
feat: allow configuring Datadog plugin as code on Jenkins controllers

lemeurherve:
This PR allows the possibility to configure Datadog plugin parameters as code on Jenkins controllers if the section `profile::jenkinscontroller::jcasc['datadog']` is defined.

### DIFF
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -611,6 +611,8 @@ profile::jenkinscontroller::jcasc:
               - maven-19-windows
   artifact_caching_proxy:
     disabled: false
+  datadog:
+    host: "ci.jenkins.io"
 # These are plugins we need in our ci environment
 profile::jenkinscontroller::plugins:
   - ansicolor # Provides ANSI color in the UI when checking log outputs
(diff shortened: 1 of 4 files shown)


dduportal:
Started to run local tests with the vagrant VM on this PR.

While doing so, May I ask to either update the title and body to explain it adds the "datadog" feature to the  `jenkins::controller` provider with opt-in AND it also enable the feature for ci.jenkins.io (or split in 2 PRs)?

hieradata/clients/azure.ci.jenkins.io.yaml
```diff
@@ -617,6 +617,8 @@ profile::jenkinscontroller::jcasc:
               - maven-19-windows
   artifact_caching_proxy:
     disabled: false
+  datadog:
```
dduportal: Don't forget to add the plugin (same file, a list of plugins) `datadog` (I'm not sure if it is easy to make it automatic)

dduportal (approved):
Tested locally, both cases (default with no datadog, and with datadog) 

LGTM 👍

dduportal:
Confirmed that puppet agent behaved as expected:

- Did create an empty file `/var/lib/jenkins/casc.d/datadog.yaml` in the controllers

PR #2835, opened 2023-05-16
feat(*.trusted.ci.jenkins.io) add the 3 Azure VMs for trusted.ci.jenkins.io under puppet management puppet management

dduportal:
This PR "only" adds the 3 new machines for trusted (ref. https://github.com/jenkins-infra/helpdesk/issues/3486).

It uses the "puyppet hostnames" from https://github.com/jenkins-infra/azure/pull/344

Merging this PR won't trigger a puppet agent runs on this VM: they need an initial binding from an admin on the puppet master, so no impact expected

### DIFF
--- a/hieradata/clients/trusted.ci.jenkins.io.yaml
+++ b/hieradata/clients/trusted.ci.jenkins.io.yaml
@@ -1,0 +1,143 @@
+---
+lookup_options:
+  "^profile::jenkinscontroller::jcasc$":
+    merge:
+      strategy: deep
+      merge_hash_arrays: true
+  "^letsencrypt":
+    merge:
+      strategy: deep
+      merge_hash_arrays: true
+
+accounts:
+  wfollonier:
+    ssh_keys:
+      wfollonier:
+        key: AAAAB3NzaC1yc2EAAAADAQABAAABAQC3oj0NN9UL1dIfBP44JDsOj/bGX/DG/GIv82imhgWbCQvsKcPczb32+W+zVo+OF3mADX4EoBG681GGopjYCKLreo8D3nFSP/+kdAt0lEqbufzoLvSyyxa0RUHDwzVQIiMiNlzDiWqLRkF2TdeFDl5u+bbdPTYCGil5/qZ3Ro8sG9ayWXMxh+f+s0MAU9qFIwau838RF2R9OCMjmPodW/zf+Mcq+SqrbZyYfYha5jOWxoN8IdrGuAOQeYks7mrXC6qq5N9ojUtMKONvayFwNOsuC0U8PYUtukkHVnm2IK9KM1A38HDlV9eTF8ac7yCUHLlttmoIBdLUuLJAw72BQkK/
+    groups:
+      - sudo
+  notmyfault:
+    ssh_keys:
+      notmyfault_rsa:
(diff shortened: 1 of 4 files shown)


manifests/site.pp
```diff
@@ -111,3 +111,14 @@
 node 'bounce' {
   include role::bounce
 }
+
+## New VMs (Azure) for trusted.ci.jenkins.io
```
lemeurherve, line 118:
```suggestion
# New VMs (Azure) for trusted.ci.jenkins.io
```

lemeurherve: The above "bounce" node L110-113 will be removed after? If so, I would add a TODO comment.

hieradata/clients/trusted.ci.jenkins.io.yaml
```diff
@@ -0,0 +1,143 @@
...
+      strategy: deep
+      merge_hash_arrays: true
+
+accounts:
+  wfollonier:
+    ssh_keys:
+      wfollonier:
+        key: AAAAB3NzaC1yc2EAAAADAQABAAABAQC3oj0NN9UL1dIfBP44JDsOj/bGX/DG/GIv82imhgWbCQvsKcPczb32+W+zVo+OF3mADX4EoBG681GGopjYCKLreo8D3nFSP/+kdAt0lEqbufzoLvSyyxa0RUHDwzVQIiMiNlzDiWqLRkF2TdeFDl5u+bbdPTYCGil5/qZ3Ro8sG9ayWXMxh+f+s0MAU9qFIwau838RF2R9OCMjmPodW/zf+Mcq+SqrbZyYfYha5jOWxoN8IdrGuAOQeYks7mrXC6qq5N9ojUtMKONvayFwNOsuC0U8PYUtukkHVnm2IK9KM1A38HDlV9eTF8ac7yCUHLlttmoIBdLUuLJAw72BQkK/
+    groups:
+      - sudo
```
lemeurherve, line 18: What this group is for?

hieradata/clients/trusted.ci.jenkins.io.yaml
```diff
@@ -0,0 +1,143 @@
...
+            type: "zip"
+            label: "(linux && amd64) || updatecenter || census"
+  permanent_agents:
+    agent-1:
+      remoteFS: /home/jenkins/agent-workspace
+      numExecutors: 2
+      mode: EXCLUSIVE
+      labels:
+        - updatecenter
+        - census
```
lemeurherve, line 66: What this label is for?

manifests/site.pp
dduportal: The 2 characters comments means it's aimed at a "group" of directive (while 1 character is "only the next line/directive)

It's purposed willingly

dduportal: > The above "bounce" node L110-113 will be removed after? If so, I would add a TODO comment.

Good idea, even though it should not be part of this PR because we do not know, in the current state, if the former bounce directive will have to be removed or not (what if we rollabck the  change?).

It's the responsibility of the parent issue to track this. "TODO comment" are not a guarantuee that it will be treated

lemeurherve: Maybe not a TODO but at least a comment on either of them explaining why there are two trusted "bounces"? It feels weird to me, it will certainly feel weirder later ^^

hieradata/clients/trusted.ci.jenkins.io.yaml
dduportal: It adds this user to the `sudo` group on the machines

hieradata/clients/trusted.ci.jenkins.io.yaml
dduportal: I dont know : it is the same value as currently migrated VM so I won't change or remove it.

I assume there might be a job using it, or none. But it's not the subject of this PR neither of the VM migration.

If you think it should be searched and cleaned up, don't hesitate to open an issue with your researches

manifests/site.pp
dduportal: TODO added with an explanation that **should** content both point of view: is that ok for you?

dduportal:
Merging as the puppet agents are enabled and running successfully on the 3 VMs.

PR #2837, opened 2023-05-16
fix: expose ports for Datadog agent

lemeurherve:
This PR exposes the two localhost ports needed to communicate between the docker container from which Jenkins controller runs and the Datadog agent, running directly on the VM.


dduportal (changes requested):
The datadog-agent runs on the host so publishing the ports of the container won't work alas

dduportal:
Closing this PR as we worked together and the subject and that will need another angle with datadog agent setup

PR #2829, opened 2023-05-12
feat(ci.jenkins.io) use spot + ephemeral disk for all Azure VM Agents

dduportal:
As per https://github.com/jenkins-infra/helpdesk/issues/3551#issuecomment-1545726927, using ephemeral disk (and spot) would help decreasing prices.

It's a follow up of #2828 but scoped to all the other VM kinds.

It introduces the following changes: 

- Linux x86_64 VM agent (normal, eg. not "highmem"):
  - Switch from `Standard_D4s_v3` to `Standard_D4ads_v5` instance (cheaper both on demand and spot)
  - Use ephemeral disk (new instance size allows for 150 Gb)
 - Linux ARM64 VM agent (normal, eg. not "highmem"):
  - Switch from `Standard_D4ps_v5` to `Standard_D4pds_v5` instance (to allow ephemeral disk and avoid paying 0.05 per hour for the OS disk)
  - Use ephemeral disk (new instance size allows for 150 Gb)
- Windows 2019 VM agent:
  - Switch from `Standard_D4s_v3` to `Standard_D4ads_v5` instance (cheaper both on demand and spot)
  - Use ephemeral disk (new instance size allows for 150 Gb)
  - Enable spot
- Windows 2022 VM agent:
  - Switch from `Standard_D4s_v3` to `Standard_D4ads_v5` instance (cheaper both on demand and spot)
  - Use ephemeral disk (new instance size allows for 150 Gb)
  - Enable spot

PR #2896, opened 2023-06-07
feat(trusted.ci) Update instance type for Azure VM ephemeral agents

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3486

This PR updates the instance type for trusted.ci.jenkins.io's VM agents to use the same serie as ci.jenkins.io (introduced in https://github.com/jenkins-infra/jenkins-infra/pull/2829).

- These VMs are using ephemeral disks instead of managed disk: faster to boot, cheaper, for the same performances (local NVMe)
- Used the same instance size as the "highmem" on ci.jenkins.io: same amount of vCPUs but 32Gb of memory instead of 28Gb.
- Enabled datadog agent on these machines to collect system metrics:
  - Permanent agent already has the agent enabled
  - We have to be sure that these machines aren't overkill for the tasks on trusted.ci.jenkins.io, so we need metrics

Please note that;
- This PR has been tested with the local vagrant setup (see the hieradata/vagrant change)
- The ci.jenkins.io and cert.ci.jenkins.io controller does not benefit from the "generic init script change": should be subsequent PRs


smerle33 (approved):
Perfect !

PR #2898, opened 2023-06-08
feat(cert.ci) update to ubuntu 22.04

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2982, this PR ensures that puppet agent run does not fail once on Ubuntu 22.04 for cert.ci.jenkins.io


dduportal:
Self merging as tested in production (the check failure is due to updatecli)

PR #2909, opened 2023-06-20
feat(ci.jenkins.io) allow containers to use dogStatsD protocol with the host datadog agent

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3573

- https://forge.puppet.com/modules/datadog/datadog_agent/readme#configuration-variables shows that a map can be used to pass additional extra options to the datadog agent conf
- A firewall rule is added to the Docker profile to allow containers to communicate with the dogstatsd


dduportal:
Tested manually: let's roll

PR #2912, opened 2023-06-21
feat(ci.jenkins.io) enable datadog full integration with logs and traces

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3573

This PR enables the full Jenkins integration for ci.jenkins.io.

It is an "all-in-one" PR because the JCasc datadog plugin and the datadog agent plugin are pretty tied in this case.

- Ports for metrics, traces and logs collection are defined as top-level variable of the profile `jenkinscontroller` to allow reusability across different profiles
  - Direct consequence: the firewall definition controlling the dogstats UDP moved from profile `docker` to `jenkinscontroller`
- A few code cleanups along the way (ensuring proper owner and mode for datadog related changes, removing deprecated puppet hierdata syntaxes, YAML JCasc bgufix for datadog binding field)

### DIFF
--- a/dist/profile/manifests/jenkinscontroller.pp
+++ b/dist/profile/manifests/jenkinscontroller.pp
@@ -30,6 +30,11 @@ class profile::jenkinscontroller (
   String $groovy_d_set_up_git                  = 'absent',
   String $groovy_d_lock_down_jenkins           = 'absent',
   Hash $jcasc                                  = {},
+  Hash $datadog                                = {
+    metrics_collection_port => 8125,
+    traces_collection_port  => 8126,
+    logs_collection_port    => 8127,
+  },
   Boolean $block_remote_access_api             = false,
   String $memory_limit                         = '1g',
   String $java_opts = "-server \
@@ -240,6 +245,45 @@ class profile::jenkinscontroller (
     }
   } else {
     $jcasc_java_opts = ''
+  }
+
+  if $jcasc_final_config['datadog'] and $jcasc_final_config['datadog']['collectBuildLogs'] {
+    firewall { '901 accept datadog local metrics collection':
+      proto   => 'udp',
+      dport   => $datadog['metrics_collection_port'],
+      iniface => ['! lo', '! docker0'],
+      action  => 'reject',
+    }
+
+    firewall { '902 accept datadog local jenkins logs collection':
+      proto   => 'tcp',
+      dport   => $datadog['logs_collection_port'],
+      iniface => 'docker0',
+      action  => 'accept',
+    }
+
+    firewall { '903 accept datadog local trace collection':
+      proto   => 'tcp',
+      dport   => $datadog['traces_collection_port'],
+      iniface => 'docker0',
+      action  => 'accept',
+    }
+
+    file { "${datadog_agent::params::conf_dir}/jenkins.d":
+      ensure  => directory,
+      owner   => $datadog_agent::params::dd_user,
+      group   => $datadog_agent::params::dd_group,
+      mode    => '0755',
+      require => Class['datadog_agent'],
+    }
+    file { "${datadog_agent::params::conf_dir}/jenkins.d/conf.yaml":
+      ensure  => file,
+      owner   => $datadog_agent::params::dd_user,
+      group   => $datadog_agent::params::dd_group,
+      mode    => '0644',
+      require => File["${datadog_agent::params::conf_dir}/jenkins.d"],
+      content => template("${module_name}/jenkinscontroller/datadog_jenkins_conf.yaml.erb"),
+    }
   }
 
   ##############################################################################
--- a/dist/profile/manifests/jenkinscontroller.pp
+++ b/dist/profile/manifests/jenkinscontroller.pp
@@ -247,7 +247,7 @@ class profile::jenkinscontroller (
     $jcasc_java_opts = ''
   }
 
-  if $jcasc_final_config['datadog'] and $jcasc_final_config['datadog']['collectBuildLogs'] {
+  if $jcasc_final_config['datadog'] {
     firewall { '901 accept datadog local metrics collection':
       proto   => 'udp',
       dport   => $datadog['metrics_collection_port'],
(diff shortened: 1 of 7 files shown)


dduportal:
Partially tested on ci.jenkins.io with @lemeurherve but the deployment will be done controller by controller once this PR is approved and merged

dist/profile/manifests/jenkinscontroller.pp
```diff
@@ -242,6 +247,45 @@
     $jcasc_java_opts = ''
   }
 
+  if $jcasc_final_config['datadog'] and $jcasc_final_config['datadog']['collectBuildLogs'] {
```
lemeurherve: `collectBuildLogs`, not sure about the "Build" part: here firewall rules are added for metrics, logs and tracing, not only the build logs.

hieradata/clients/azure.ci.jenkins.io.yaml
lemeurherve, line 1: > It is an "all-in-one" PR because the JCasc datadog plugin and the datadog agent plugin are pretty tied in this case.

I suppose these modifications couldn't be done in a following PR?

hieradata/vagrant/common.yaml
```diff
@@ -348,6 +353,9 @@ profile::jenkinscontroller::jcasc:
     disabled: false
   datadog:
     host: "vagrant.local"
+    targetHost: "172.18.0.1" # docker0 interface in vagrant is non standard (because docker in docker)
+    collectBuildLogs: true
+    emitConfigChangeEvents: true
```
lemeurherve, line 358: Found https://javadoc.jenkins.io/plugin/datadog/index.html?org/datadog/jenkins/plugins/datadog/DatadogGlobalConfiguration.html but not sure about what this parameter is about

lemeurherve: Related to
> YAML JCasc bgufix for datadog binding field

?

dist/profile/manifests/jenkinscontroller.pp
dduportal: The value has been extracted from an export of the JCasc setup we tried yesterday on ci.jenkins.io + tested on a local vagrant VM.

The attribute `collectBuildLogs` maps to the checkbox "Enable Log Collections":

Capture d’écran 2023-06-22 à 12 16 03


This name makes sense when you think about it: the datadog plugin **only** collects build logs in Jenkins. I would argue the UI of the plugin is badly named as it misses the "Build" word.

hieradata/clients/azure.ci.jenkins.io.yaml
dduportal: Nope. I tried to split but it is highty coupled (as I define the datadog port in the `jenkinscontroller` profile).

I can totally start with a first PR with the port value "as it" and then update the PR here with the factorization if you prefer, but since the datadog setup has already been done manually on ci.jenkins.io (and is not overriden by the JCasc setup because disabled in puppet), it should not be a problem.

hieradata/vagrant/common.yaml
dduportal: About `emitConfigChangeEvents`:

- It is set to `true` on ci.jenkins.io so I've adapted the config as code setup here to map to it (that is the reason of the change). I don't remember changing it, but I can't be 100% it's not me though.

- This setting is useful if we want to track the config change history (like the configHistory plugin used to do with local git repositories) : it gives visibility on "what is happening" inside Jenkins.

> YAML JCasc bgufix for datadog binding field

Absolutley not. The bugfix is to add double quotes, in  the ERB template of the datadog JCasc YAML file, around the value associated to `targetHost`, as you dont want YAMl parser to mistake an Ip address with a number. Using double quotes forces type to string.

dist/profile/manifests/jenkinscontroller.pp
lemeurherve: OK. I'm a little bit confused by this condition, shouldn't firewall rules be defined even if this parameter is false?

hieradata/vagrant/common.yaml
lemeurherve: > This setting is useful if we want to track the config change history (like the configHistory plugin used to do with local git repositories) : it gives visibility on "what is happening" inside Jenkins.

Shouldn't this change be in a distinct PR?

dist/profile/manifests/jenkinscontroller.pp
dduportal: - You are correct for the 901 (UDP/8125) and 903 (TCP/8126) firewall rules.

- But the 902 (TCP/8127) is bound to the `$jcasc_final_config['datadog']['collectBuildLogs']` presence (along with the file `${datadog_agent::params::conf_dir}/jenkins.d/conf.yaml` (and it parent dir) => there is no reason to create the file and the firewall rule if Jenkins is not set up to collect build logs.

I'm gonna nest the `if` blocks to make it less confusing then.

hieradata/vagrant/common.yaml
dduportal: > Shouldn't this change be in a distinct PR?

IF the PR would change the current production setup: yes it should be a distinct PR.

But it is not the case, due to the way JCasc works for this particular Jenkins plugin (e.g. once you have set up through the UI, if JCasc does NOT change the value, then the value persists).

Not including this change in the PR here would change the setup once applied.

dduportal: Additionally, avoiding too much ci.jenkins restart is a constraint: it adds another compelling reason to have an all in one PR (since the scope is clearly defined)

dist/profile/manifests/jenkinscontroller.pp
dduportal: Done in https://github.com/jenkins-infra/jenkins-infra/pull/2912/commits/77c1ccb70c7d3f533e0c77ea5985c9a7e4506813

dist/profile/manifests/jenkinscontroller.pp
```diff
@@ -242,6 +247,47 @@
...
+
+    firewall { '902 accept datadog local trace collection':
+      proto   => 'tcp',
+      dport   => $datadog['traces_collection_port'],
+      iniface => 'docker0',
+      action  => 'accept',
+    }
+
+    if $jcasc_final_config['datadog']['collectBuildLogs'] {
+      firewall { '905 accept datadog local jenkins logs collection':
```
lemeurherve, line 266:
```suggestion
      firewall { '903 accept datadog local jenkins logs collection':
```

lemeurherve (approved):

dist/profile/manifests/jenkinscontroller.pp
lemeurherve: (Restore rule index specified before https://github.com/jenkins-infra/jenkins-infra/commit/77c1ccb70c7d3f533e0c77ea5985c9a7e4506813)

dduportal: I chose to change it to avoid confusion:

- The rules are not created on ci.jenkins.io yet, so no constraint on, the order (besides, iptables priorities are easier to manage than Azure NSG priorities)
- We might have to create a few other rules for the metric and traces, So i've let the indexes available (`903` and `904`)
- The rule was `902` previously (and `903` for the traces). Now traces have the `902` priority. So using `905` avoids any confusion on the order and nested conditions.

PR #2922, opened 2023-06-24
fix(openvpn) add ports 22 and 5432 to iptables nat rules + ensure default forward policy is accept

dduportal:
While [upgrading Ubuntu to 22.04](https://github.com/jenkins-infra/helpdesk/issues/2982#issuecomment-1604546958) on vpn.jenkins.io, the routing from VPN network (`10.0.8.0/24`) to the ci.jenkins.io / cert.ci.jenkins.io network (`10.0.2.0/24`) was broken.

This PR follows up https://github.com/jenkins-infra/docker-openvpn/pull/283 (applied in #2921) and fixes the following elements:

- Ensure the default FORWARD policy is `accept` in iptables (fixed manually on the VM with the command `iptables --policy FORWARD ACCEPT`) otherwise no packet are forwarded between networks
  - With Ubuntu 18.04, this setting was not blocking packages but I don't understand why. I guess iptables version in Ubuntu 22.04 is stricter with rules order.
- Ensure the (now unused, because we are migrating to `public-vnet`) network `10.0.1.0/24` is not set up. The network interface was removed from the VM manually in Azure
- Fix the `10.0.8.0` netmask which was not `/24` (same as https://github.com/jenkins-infra/docker-openvpn/pull/283)
- Since recent iptables is stricter, explicitly allow ports 22 and 5432 when crossing networks from VPN nets

### DIFF
--- a/dist/profile/manifests/openvpn.pp
+++ b/dist/profile/manifests/openvpn.pp
@@ -119,6 +119,12 @@ class profile::openvpn (
     iniface => 'eth0',
   }
 
+  # Set default FORWARD CHAIN policy to ACCEPT
+  firewallchain { 'FORWARD:filter:IPv4':
+    ensure => present,
+    policy => 'accept',
+  }
+
   # Create firewall rules and route for each specified NIC to allow routing from VPN virtual networks to different networks
   lookup('profile::openvpn::networks').each |$network_nic, $network_setup| {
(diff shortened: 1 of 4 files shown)


dduportal:
Tested with https://github.com/jenkins-infra/jenkins-infra/pull/2923 (as much as possible), let's break things now

PR #2925, opened 2023-06-26
feat(ci.jenkins.io) switch azure-vm agents to inbound mode in the new public-vnet network

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3535 and https://github.com/jenkins-infra/helpdesk/issues/3481#issuecomment-1497021548

This PR switches ci.jenkins.io's Azure VM agents templates to inbound agent mode in the public-vnet network.

- Both changes (inbound mode + new vnet) are coupled but it's required: the inbound connection in the current (legacy network) results in "TCP connection errors" due to the network overlaps :( 
- Cleaned up the init script to:
  - Ensure it's defined before the JCasc definition, to avoid breaking anchor references when reordering JCasc agent templates (was done last month on trusted.ci, using the same trick here)
  - Made the script instructions safe: using double quotes (to avoid YAML parsing errors) and escaping of "double quote" and "back slashes" (on Windows)
  - Note that the AWS EC2 init script is kept aside. Not only we ought to remove it in another PR, but it also is NOT ready for inbound mode

- Tested the scripts on ci.jenkins.io with a test azure vm cloud manually first, both Linux and Windows
  - Then transplanted the settings to the local vagrant testing
  - Cleaned up and checked the result on the local vagrant
  - But I did NOT closed the loop due to the risks of exposing encrypted credentials
  - Hence the puppet agent is kept disabled on ci.jenkins.io: I'll run a noop first once merged + I'll check all agents and report back here

### DIFF
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -49,7 +49,7 @@ x-inbound-agent-init-shell-script: &InboundAgentInitShellScript
   - "#!/bin/bash"
   - "set -eux"
   - "systemctl stop datadog-agent.service"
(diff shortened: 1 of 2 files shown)


dduportal:
Dang, I commited the datadog api key in the vagrant hierdata.
Credential rotate (and updated in the 2nd commit)

dduportal:
Self Merging to deploy

dduportal:
Fixup in https://github.com/jenkins-infra/jenkins-infra/pull/2926 (missing commit)

PR #2929, opened 2023-06-26
cleanup(ci.jenkins.io) remove EC2 agent template

dduportal:
Since https://github.com/jenkins-infra/jenkins-infra/pull/2826, we do not need any EC2 agent at all on ci.jenkins.io as ARM64 CPUs are now handled by Azure VMs.

This PR removes last EC2 remnants managed as code.

Please note that, once applied, 2 credentials will have to be cleaned up manually from ci.jenkins.io and inside AWS console:

- `ec2-agent-ssh` 
- `aws-credentials-jenkins-ci`

Also, the EC2 plugin will have to be removed from the UI once applied

### DIFF
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -489,39 +489,6 @@ profile::jenkinscontroller::jcasc:
               - default
             cpus: 1
             memory: 1
-    ec2:
-      aws-us-east-2:
-        region: us-east-2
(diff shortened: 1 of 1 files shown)


lemeurherve (approved):

timja (approved):

dduportal:
- Deployed
- Removed the 2 credentials (pre-checked on the JCasc export confirmed not used anywhere else)
- Removed the `ec2` plugin and all the AWS-SDK plugin not locked (AWS SDK::Minimal is required for S3 artifact manager for instance)

PR #2934, opened 2023-06-29
chore(datadog): don't collect container logs from `ci.jenkins.io`

lemeurherve:
As Jenkins logs are now collected by the Datadog agent, we don't need anymore to collect them as container logs.
Since Jenkins is the only container running on ci.jenkins.io machine, this PR disable the container logs collection.

### DIFF
--- a/hieradata/clients/azure.ci.jenkins.io.yaml
+++ b/hieradata/clients/azure.ci.jenkins.io.yaml
@@ -30,6 +30,7 @@ datadog_agent::apm_enabled: true
 datadog_agent::apm_non_local_traffic: true # Allow jenkins container to contact the agent for traces and logs
 datadog_agent::agent_extra_options:
   bind_host: "0.0.0.0" # All hosts interfaces to allow container access
+datadog_agent::container_collect_all: true # Don't collect Jenkins container logs (the only container running) as we're already collecting these logs via the Datadog plugin
 profile::jenkinscontroller::anonymous_access: true
 profile::jenkinscontroller::admin_ldap_groups:
   - admins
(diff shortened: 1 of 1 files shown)


hieradata/clients/azure.ci.jenkins.io.yaml
```diff
@@ -30,6 +30,7 @@ datadog_agent::apm_enabled: true
 datadog_agent::apm_non_local_traffic: true # Allow jenkins container to contact the agent for traces and logs
 datadog_agent::agent_extra_options:
   bind_host: "0.0.0.0" # All hosts interfaces to allow container access
+datadog_agent::container_collect_all: true # Don't collect Jenkins container logs (the only container running) as we're already collecting these logs via the Datadog plugin
```
dduportal:
```suggestion
datadog_agent::container_collect_all: false # Don't collect Jenkins container logs (the only container running) as we're already collecting these logs via the Datadog plugin
```

Ref.
- https://forge.puppet.com/modules/datadog/datadog_agent/readme 

> container_collect_all | A boolean to enable logs collection for all containers.

- https://github.com/DataDog/puppet-datadog-agent/blob/57ffbd188b0b32678328e3e4ac743b44b8c6e3c7/manifests/init.pp#L662

dduportal (changes requested):
(edit) I would expect it to be false already (as default value). There is something fishiy in their module 🤔

dduportal:
@lemeurherve I can try on a local vagrnat VM if you want?

hieradata/clients/azure.ci.jenkins.io.yaml
lemeurherve: 🤦

lemeurherve:
No need, my mistake 🤦 

Should be good now 🙂

dduportal (approved):

PR #2964, opened 2023-07-06
fix: disable `artifact-manager-s3` stashes deletion on Jenkins controllers

lemeurherve:
This PR is disabling the stashes deletion by the `artifact-manager-s3` plugin, currently reporting errors for ci.jenkins.io as noted in https://github.com/jenkins-infra/helpdesk/issues/3643

As the associated storage cost is low, there is no need for immediate cleanup of stashes, it can be done outside ci.jenkins.io.


timja (approved):

dduportal (approved):
Note: this require a controller restart to be effective

PR #2993, opened 2023-08-02
feat(ci.jenkins.io) set highmem default to spot and add an highmem-nonspot agent template

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3673#issuecomment-1660606208

This PR ensures that there are 2 High memory agent templates in ci.jenkins.io available for developers.

- The spot is the default (reverting the bevhaior of #2983 + #2986) 
- The "non spot" instance can be requests with the following labels: `highmem-nonspot`, `highram-nonspot`, docker-highmem-nonspot` and `linux-amd64-big-nonspot`


MarkEWaite (approved):

smerle33 (approved):
🚀

PR #3084, opened 2023-09-26
feat: initial support of the archives.do.jenkins.io VM

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3760

- Stop trying to install the LVM2 package on both `archives.jenkins.io` and `archives.do.jenkins.io`
  - It was only used by `archives.jenkins.io` when it was running in rackspace, before Oracle
  - It won't remove the package from `archives.jenkins.io`, but at least it won't install it in `archives.do.jenkins.io`
- Add the node `archives.do.jenkins.io` under puppet management
  - Custom data volume mounted (UUID retrieved from the spawned VM)
- Add a hieradata client manifest for `archives.do.jenkins.io`
  - Includes a custom datadog naming to allow discriminating the metrics/logs between old and new VM
  - Customized allow list for rsync

### DIFF
--- a/dist/profile/manifests/archives.pp
+++ b/dist/profile/manifests/archives.pp
@@ -73,11 +73,6 @@ class profile::archives (
     }
   }
 
-  #
-  package { 'lvm2':
-    ensure => present,
-  }
-
   package { 'libapache2-mod-bw':
     ensure => present,
   }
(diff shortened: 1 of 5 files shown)


dduportal:
Self merging to bootstrap the first puppet run (which will most probably fails...).

For info, the new VM (`archives.do.jenkins.io`) already has:

- The 750Gb volume mounted to `/srv` but manually (rebooting the VM unmount it today). Proof that the puppet management works as expected will be to reboot the VM and still have the volume mounted.
- The `/etc/letsencrypt` directory from `archives.jenkins.io`  copied through rsync to ensure a successfull bootstrap. However, we'll have to clean it up afterward to avoid the same issue as https://github.com/jenkins-infra/helpdesk/issues/3740#issuecomment-1701415305

PR #3113, opened 2023-10-13
feat(*.ci.jenkins.io) use JDK21 GA (exception of `s390x` sticking to EA)

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3783

paired with @gounthar


lemeurherve:
I don't have exactly the same here: https://github.com/jenkinsci/docker-agent/pull/521/files#diff-5ccf4f9636b549b76a1ae14b28ab90bbf9a4ba82774e416679d214444cef80ceR33-R35

updatecli/scripts/check-jdk.sh
```diff
@@ -30,10 +30,10 @@ function get_jdk_download_url() {
       echo "https://github.com/adoptium/temurin19-binaries/releases/download/jdk-${jdk_version}/OpenJDK19U-jdk_${platform}_hotspot_${jdk_version//+/_}";
       return 0;;
     21*)
-      # JDK version (21+35-ea-beta)
-      ##    https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21%2B35-ea-beta/OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21-0-35.tar.gz
+      # JDK version (21+35)
+      ##    https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21%2B35/OpenJDK21U-jdk_aarch64_linux_hotspot_21_35.tar.gz
       urlEncodedJDKVersion="${jdk_version//+/%2B}"
-      echo "https://github.com/adoptium/temurin21-binaries/releases/download/jdk-${urlEncodedJDKVersion}/OpenJDK21U-jdk_${platform}_hotspot_ea_21-0-$(echo ${jdk_version} | cut -d '+' -f 2 | cut -d '-' -f 1)"
+      echo "https://github.com/adoptium/temurin21-binaries/releases/download/jdk-${urlEncodedJDKVersion}/OpenJDK21U-jdk_${platform}_hotspot_${jdk_version//+/_}"
```
lemeurherve: ~~According to https://github.com/adoptium/temurin21-binaries/releases/tag/jdk-21%2B35, the links contain OpenJDK21U-**debugimage**_<...> (ex: `https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21%2B35/OpenJDK21U-debugimage_aarch64_alpine-linux_hotspot_21_35.tar.gz`)~~

~~This will certainly evolve, I would add a TODO about removing it when it will change.~~

updatecli/scripts/check-jdk.sh
```diff
@@ -30,10 +30,10 @@ function get_jdk_download_url() {
       echo "https://github.com/adoptium/temurin19-binaries/releases/download/jdk-${jdk_version}/OpenJDK19U-jdk_${platform}_hotspot_${jdk_version//+/_}";
       return 0;;
     21*)
-      # JDK version (21+35-ea-beta)
-      ##    https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21%2B35-ea-beta/OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21-0-35.tar.gz
+      # JDK version (21+35)
+      ##    https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21%2B35/OpenJDK21U-jdk_aarch64_linux_hotspot_21_35.tar.gz
       urlEncodedJDKVersion="${jdk_version//+/%2B}"
```
lemeurherve:
```suggestion
      underscoredJDKVersion="${jdk_version//+/_}"
```

updatecli/scripts/check-jdk.sh
lemeurherve:
```suggestion
      echo "https://github.com/adoptium/temurin21-binaries/releases/download/jdk-${underscoredJDKVersion}/OpenJDK21U-jdk_${platform}_hotspot_${underscoredJDKVersion}"
```

smerle33: Why do we need to track those debug image ?

lemeurherve: Nevermind, yesterday they were the only ones available, didn't check the link again today.

lemeurherve: Edited my suggestion 🙂

dduportal: nope, the `debugimage` series is only a subset of all the elements. These are not the binary we want (it was the case for previous JDK btw)

For Linux aarch64 (not Alpine) we want the following:

```
https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21%2B35/OpenJDK21U-jdk_aarch64_linux_hotspot_21_35.tar.gz
```

which means the code in this line looks good

dduportal: we don't want to use the debug images. See my comment in https://github.com/jenkins-infra/jenkins-infra/pull/3113#discussion_r1358029572

updatecli/scripts/check-jdk.sh
dduportal: I remember weird issues when trying to download from the URL with the `_`. Let's keep the encoded version (unless you are able to test your suggestion and prove that it works)

updatecli/scripts/check-jdk.sh
timja: debug image is a different image and not what we want

lemeurherve: Quoting my previous comment:

> Nevermind, yesterday they were the only ones available, didn't check the link again today.

updatecli/scripts/check-jdk.sh
```diff
@@ -52,7 +52,8 @@ case "${1}" in
   19.*+*)
     platforms=("x64_linux" "x64_windows" "aarch64_linux" "s390x_linux");;
   21*+*)
-    platforms=("x64_linux" "x64_windows" "aarch64_linux" "s390x_linux");;
+    # No s390x support for JDK21 yet
```
lemeurherve:
```suggestion
    # No s390x support for JDK21
```
According to https://github.com/adoptium/temurin/issues/8 there won't be any s390x support for JDK21.

updatecli/weekly.d/jenkinscontroller-tools-jdk21.yaml
```diff
@@ -12,25 +12,19 @@ scms:
...
-    scmid: temurin21-binaries
     spec:
+      owner: "adoptium"
+      repository: "temurin21-binaries"
+      token: "{{ requiredEnv .github.token }}"
+      username: "{{ .github.username }}"
       versionfilter:
         kind: regex
-        pattern: ".*-ea-.*"
+        pattern: "^jdk-21.(\\d*).(\\d*).(\\d*)+(\\d*)$"
```
lemeurherve: NB: these dots in the regexp match any character, not the `.` character.

Not a problem, just a remark.

lemeurherve (dismissed):

updatecli/scripts/check-jdk.sh
gounthar: Quoting Tim Ellison from Adoptium [in their Slack](https://adoptium.slack.com/archives/C09NLQQAV/p1697121672337069?thread_ts=1697112748.718609&cid=C09NLQQAV):

> As George and Stewart said above, the reason there is no Temurin 21 for ppc64 AIX/s390x Linux/arm32 Linux at the moment is because the OpenJDK 21 GA implementation on those platforms is not passing our release tests. When the fixes are available in OpenJDK source then we can proceed with a release.

So we may see s390x JDK21 builds in the near future (with 21.0.1 for example).
We won't see s390x JDK21+35 builds though.

updatecli/weekly.d/jenkinscontroller-tools-jdk21.yaml
smerle33: we need to match `jdk-21+35` so if we escape the dots, I think we will not catch it, don't you think ?. 
I didn't question the pattern as it is the one used for all our version of jdk but maybe we should.

lemeurherve: Proposition:
```suggestion
        pattern: "^jdk-21(\\.\\d+)*\\+(\\d+)$"
```
jdk-21 followed by zero to N groups of a dot and one to N numbers, followed by a plus and one to N numbers

lemeurherve: image

lemeurherve: I can propose this change in a distinct PR if you prefer.

gounthar (approved):

updatecli/scripts/check-jdk.sh
```diff
@@ -30,10 +30,10 @@ function get_jdk_download_url() {
       echo "https://github.com/adoptium/temurin19-binaries/releases/download/jdk-${jdk_version}/OpenJDK19U-jdk_${platform}_hotspot_${jdk_version//+/_}";
       return 0;;
     21*)
-      # JDK version (21+35-ea-beta)
-      ##    https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21%2B35-ea-beta/OpenJDK21U-jdk_aarch64_linux_hotspot_ea_21-0-35.tar.gz
+      # JDK version (21+35)
+      ##    https://github.com/adoptium/temurin21-binaries/releases/download/jdk-21%2B35/OpenJDK21U-jdk_aarch64_linux_hotspot_21_35.tar.gz
```
dduportal:
```suggestion
      ## JDK21 URLs have an underscore ('_') instead of a plus ('+') in their archive names
```

Nitpick: Let's keep the same pattern as other JDKs

updatecli/scripts/check-jdk.sh
```diff
@@ -52,7 +52,8 @@ case "${1}" in
   19.*+*)
     platforms=("x64_linux" "x64_windows" "aarch64_linux" "s390x_linux");;
   21*+*)
-    platforms=("x64_linux" "x64_windows" "aarch64_linux" "s390x_linux");;
+    # No s390x support for JDK21 yet
+    platforms=("x64_linux" "x64_windows" "aarch64_linux");;
```
dduportal, line 58: What happens when there is a JDK upgrade not available with the s390x (such as `21+35-ea` => `21+35`)?

- Updatecli still open a PR with the new version in the parameters files (hieradata)
- The build passes as this is only a "hieradata -> puppet -> JCasc YAML file" check
- Then ci.jenkins.io gets updated when the updatecli PR is merged and its "tool" are setup with the new version
- And then, all builds calling the `tool 'jdk21'` pipeline instructions fails as the tool installer responds a HTTP/404 error.


There is no problem on the top-level idea to stop (temporarily) tracking the `s390x` and keep it to a lower version, but it means you must update the tool configuration to keep the old version with a TODO comment as part of the PR here.

dduportal (changes requested):
This is really good job, but we can't merge this "as-it" otherwise we'll break the s390x tool

updatecli/scripts/check-jdk.sh
smerle33: yes but this mean changing the target to point on each of them no ? 
`key: $.profile::jenkinscontroller::jcasc.tools_default_versions.jdk21`

dduportal: That would work yes. Or you can also keep the updatecli code as it to track and define the default JDK21 version (e.g. `21+35`) and define an exception in the specific hieradata of ci.jenkins.io which defines the s390x agent, as the override mechanism is for the most specific cases like this one.

dduportal (commented):
For info, please double check the results as the twin of this PR is failing to update the proper JDK21 download URL when using the new (stable) version in https://github.com/jenkins-infra/packer-images/pull/848#pullrequestreview-1681928106


=> might be packer related only, but worth checking

dduportal (changes requested):
- Updatecli looks to do what is it expected to do:

```
⚠ - key "$.profile::jenkinscontroller::jcasc.tools_default_versions.jdk21" should be updated from "21+35-ea-beta" to "21+35", in file "hieradata/common.yaml"
```

But this PR can't be merged yet:

- The JDK21 URLs must be changed in puppet code (otherwise the PR for `21+35` will break the ci.jenkins.io` tool installation
- The hieradata for ci.jenkins.io does not show specific EA version for s390x

hieradata/common.yaml
```diff
@@ -244,7 +244,7 @@ profile::jenkinscontroller::jcasc:
     jdk8: 8u392-b08
     jdk11: 11.0.21+9
     jdk17: 17.0.8+7
-    jdk21: 21+35-ea-beta
+    jdk21: 21+35
```
dduportal, line 247:
```suggestion
    jdk21: 21+35-ea-beta
```

dduportal:
Added content to fix the "upgrade" problem: 

- Adoptium template updated to cover EA and GA cases:
  - Support of JDK8, 11, 17, 21 and 21-preview
  - Support of EA and GA
  - Support of "1 digit only" versions such as `21+35`
  - Better management of line returns (allowing to remove a now-unneeded `chop()` for `build::agent` profile)
  - A LOT of comments added to explain
  - Vagrant hieradata setup to cover a lot of cases

- Fixed a bug for `build::agent` profile where `arm64` / `aarch64` architectures were not supported for JDK installation
- Updated the `check-jdk.sh` shell script used by updatecli, cherry-picking from jenkinsci/docker where the split "21 GA/ 21 EA" already exists

- As only the ci.jenkins.io's `s390x` agent need a JDK21 EA, the EA version is specified only for it, otherwise JDK21 GA is used everywhere else
  - Added an updatecli manifest to track only the EA version for `s390x` agent

The following tests were run locally, in addition of to the puppet unit tests:

- `vagrant up --provision jenkins::agent` => check it installs the proper JDK versions (tried on a macOS machine with Silicon so the `aarch64` temurin binaries were installed, and tried forcing the `amd64` which downloaded the x86 binaries as expected)
- `vagrant up --provision jenkins::controller` the 2 new cases (EA 3 digits and EA 1 digit):
  - Provisionning went well
  - The following command returned `200` for each of the found links: `for url in $(cat /var/lib/jenkins/casd.d/tools.yaml | grep temurin | cut -d'"' -f2);do echo "== echo $url"; curl -fsSL -I -o /dev/null -w "%{http_code}" "$url"; echo; done`

- Ran the new updatecli manifest (for 21 EA) locally with the target's `scmdid` commented out: it returned `⚠ - key "$.profile::jenkinscontroller::jcasc.tools.jdk.jdk21.installers.s390x.version" should be updated from "21+35-ea-beta" to "21.0.1+12-ea-beta", in file "hieradata/clients/controller.ci.jenkins.io.yaml"`

updatecli/scripts/check-jdk.sh
dduportal: For info, see https://github.com/jenkins-infra/jenkins-infra/pull/3113#issuecomment-1786727093

dduportal:
Ready for review (cc @hervelemeur):

- Please don't merge it, as I would want to deploy it using `noop` and carefull testing on *.ci.jenkins.io
- The updatecli check failed due to the new `s390x` key not found: will be successfuly once merged (see my comment above: was tested)
- The JDK21 version is still pinned to 21+35(.*). Updatecli is expected, once merged, to bump to 21.0.12(.*).

updatecli/scripts/check-jdk.sh
```diff
@@ -25,11 +25,22 @@ function get_jdk_download_url() {
       ## JDK17 URLs have an underscore ('_') instead of a plus ('+') in their archive names
       echo "https://github.com/adoptium/temurin17-binaries/releases/download/jdk-${jdk_version}/OpenJDK17U-jdk_${platform}_hotspot_${jdk_version//+/_}";
       return 0;;
-    21*)
-      # JDK version (21+35-ea-beta)
+    19*)
```
smerle33: jdk 19 ?

updatecli/scripts/check-jdk.sh
```diff
@@ -45,8 +56,12 @@ case "${1}" in
     platforms=("x64_linux" "x64_windows" "aarch64_linux" "s390x_linux");;
   17.*+*)
     platforms=("x64_linux" "x64_windows" "aarch64_linux" "s390x_linux");;
-  21*+*)
+  19.*+*)
```
smerle33: jdk19 ?

smerle33 (commented):
except for the JDK19, it looks perfect.

updatecli/scripts/check-jdk.sh
dduportal: Good catch! Removed in https://github.com/jenkins-infra/jenkins-infra/pull/3113/commits/29cd4d756c40b42af68929c9c558bbe45ecf6193

.gitignore
```diff
@@ -11,3 +11,4 @@ tmp/
 vendor/
 *.log
 .fixtures.yml
+*.zip
```
dduportal, line 14: Added because local tests are creating a "rclone.zip" file at the root of the repository. Will be fixed in a later PR

updatecli/scripts/check-jdk.sh
dduportal: Good catch! Removed in https://github.com/jenkins-infra/jenkins-infra/pull/3113/commits/29cd4d756c40b42af68929c9c558bbe45ecf6193

lemeurherve (approved):
Nice 👏

dduportal:
Merging to deploy and verify in its current state

dduportal:
- After deployment, run the following  pipeline with success:

```
parallel linuxs390x: {
    node('s390x') {
        infra.runMaven(['-v'], 21)
    }
},
linuxamd64vm: {
    node('linux-amd64 && vm') {
        infra.runMaven(['-v'], 21)
    }
},
linuxamd64container: {
    node('kubernetes') {
        infra.runMaven(['-v'], 21)
    }
},
windowsamd64vm: {
    node('windows && vm') {
        infra.runMaven(['-v'], 21)
    }
},
windowsamd64container: {
    node('aci') {
        infra.runMaven(['-v'], 21)
    }
}
```

- 2 PRs were opened (https://github.com/jenkins-infra/jenkins-infra/pull/3165/files and https://github.com/jenkins-infra/jenkins-infra/pull/3166/files) as expected: proof this PR works as expected

PR #3115, opened 2023-10-13
cleanup(updates.jenkins.io): remove File Share URL with SAS token

lemeurherve:
This PR removes this env file as it's not needed anymore.
This is now set manually as global credentials in trusted.ci.jenkins.io, to avoid showing it in logs when setting it from an env var.


dduportal (approved):
Oh good point, thanks!

Don't forget to remove the files once applied: Puppet will "only" stop tracking them once this PR is deployed. 


I let you merge when ready

PR #3131, opened 2023-10-24
Remove `sshd` plugin from cert.ci

daniel-beck:


lemeurherve (approved):

dduportal (approved):
Merging and deploying. 

FWIW This change only tells Puppet to stop checking for installing this plugin if absent (which we want).
But it still requires a manual uninstallation on the controller once deployed (and checking the controller is restarted and the plugin absent from the filesystem)

lemeurherve:
Plugin successfully removed from cert.ci

daniel-beck:
>  But it still requires a manual uninstallation on the controller once deployed (and checking the controller is restarted and the plugin absent from the filesystem)

I know, thanks. I had already marked it for uninstallation.

PR #3125, opened 2023-10-21
Split the jenkinscontroller-agents.yaml file into three files

shubhamch71:
The three files created after splitting jenkinscontroller-agents.yaml  are as follows :-  1) jenkinscontroller-agents-windows-vm-disk-size.yaml  2) jenkinscontroller-agents-packer-agent-all-in-one.yaml  3)jenkinscontroller-agents-jnlp-container-jdk17.yaml . 

This has been discussed in issue #3071

PR #3132, opened 2023-10-24
fix(ci.jenkins.io) specify correct 'java' PATH for the basic jnlp container

dduportal:
While testing https://github.com/jenkinsci/docker-inbound-agent/releases/tag/3181.v78543a_987053-1, I realized that the default `jnlp` container agent template was misconfigured: the java installation is not on the same path as the one we use on our packer-images.

Ideally, we should specify the packer image container image "all in one" on this one, but on short term, I would like to have this fixed, then proceed with #3125 and finally see what we can do to change to the "all in one" image


dduportal:
Self merging to provide a publicly visible solution for 3181 bugfix

PR #3144, opened 2023-10-27
cleanup: remove all usages of JDK19

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3289

Following up https://github.com/jenkins-infra/docker-inbound-agents/pull/125 and https://github.com/jenkins-infra/packer-images/pull/835, this PR removes all definitions of JDK19 (tools installer, agent templates, updatecli resources).


hieradata/vagrant/common.yaml
```diff
@@ -249,7 +239,7 @@ profile::jenkinscontroller::jcasc:
           virtualNetworkName: "public-vnet"
           virtualNetworkResourceGroupName: "public"
           subnetName: "public-vnet-ci_jenkins_io_agents"
-          javaHome: /opt/jdk-19       # Test override of the default JDK for builds
+          javaHome: /opt/jdk-11       # Test override of the default JDK for builds
```
lemeurherve, line 242: Do these tests still needed? Or are they stil "test"?

lemeurherve (approved):
🚀

hieradata/vagrant/common.yaml
dduportal: yes, the comment is pretty self-explanatory. The goal is not to test the JDK11 or 19, but the goal is to validate the override mechanism does what it is expected to do.

lemeurherve: I understood what troubled me, I missed we're in tests here, not the prod.

PR #3189, opened 2023-11-22
feat(ci.jenkins.io) add azurevm cloud configuration for the new subscription

dduportal:

This PR adds support for a secondary azureVM cloud on ci.jenkins.io.
Implementation assumes that VM templates must be identical between the different clouds.

Tested with local unit tests (see also CI checks) and also with the local vagrant

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
@@ -1,14 +1,15 @@
 <%- if @jcasc['cloud_agents'] && @jcasc['cloud_agents']['disabled'].to_s != "true" -%>
 jenkins:
   clouds:
-  <%- if @jcasc['cloud_agents']['azure-vm-agents'] && !@jcasc['cloud_agents']['azure-vm-agents'].empty? -%>
+  <%- if @jcasc['cloud_agents']['azure-vm-agents'] && @jcasc['cloud_agents']['azure-vm-agents']['azureVMClouds'] -%>
+  <%- @jcasc['cloud_agents']['azure-vm-agents']['azureVMClouds'].each do |azureVMCloudName, azureVMCloudSetup| -%>
   - azureVM:
-      azureCredentialsId: "<%= @jcasc['cloud_agents']['azure-vm-agents']['azureCredentialsId'] %>"
-      name: "azure-vms"
+      azureCredentialsId: "<%= azureVMCloudSetup['azureCredentialsId'] %>"
+      name: "<%= azureVMCloudName %>"
       deploymentTimeout: 1200
-      existingResourceGroupName: "<%= @jcasc['cloud_agents']['azure-vm-agents']['resource_group'] %>"
+      existingResourceGroupName: "<%= azureVMCloudSetup['resource_group'] %>"
       resourceGroupReferenceType: "existing"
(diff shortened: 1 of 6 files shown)


dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
```diff
@@ -1,14 +1,15 @@
 <%- if @jcasc['cloud_agents'] && @jcasc['cloud_agents']['disabled'].to_s != "true" -%>
 jenkins:
   clouds:
-  <%- if @jcasc['cloud_agents']['azure-vm-agents'] && !@jcasc['cloud_agents']['azure-vm-agents'].empty? -%>
+  <%- if @jcasc['cloud_agents']['azure-vm-agents'] && @jcasc['cloud_agents']['azure-vm-agents']['azureVMClouds'] -%>
```
lemeurherve: The `empty` condition wasn't needed after all?

hieradata/clients/controller.cert.ci.jenkins.io.yaml
```diff
@@ -92,8 +92,10 @@ profile::jenkinscontroller::jcasc:
         url: "https://groovy.jfrog.io/artifactory/dist-release-local/groovy-zips/apache-groovy-binary-3.0.6.zip"
   cloud_agents:
     azure-vm-agents:
-      azureCredentialsId: "azure-sp-agents" # Managed manually
-      resource_group: "cert-ci-jenkins-io-ephemeral-agents"
+      azureVMClouds:
```
lemeurherve: Out of curiosity, what the reasoning behind the choice of camelCase here?

lemeurherve (dismissed):
Two small questions, LGTM!

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
dduportal: It's not needed now that the next instruction is an `each` loop which handles the emptiness case.

hieradata/clients/controller.cert.ci.jenkins.io.yaml
dduportal: As I'm not really caring for casing conventions, It comes from me writing things without even noticing.

WDYT about?

```suggestion
      azure-vm-clouds:
```

lemeurherve: Suggestion:
```suggestion
      azure-cloud-vms:
```

dduportal: The name of the cloud is "Azure VM" and its JCasc key is `azureVM` because it is the https://plugins.jenkins.io/azure-vm-agents/ plugin.

As such, the `azure-cloud-vms` is invalid semantically.


WDYT about `clouds` since this is a sub-key of `azure-vm-agents` ?

dduportal: Updated the "casing" for the scope of this PR: WDYT?

dduportal: New updates following our discussions:

- Case changed only on the scope of this PR (the file deserves more of it but it is non functionnal)
- CamelCase kept for attributres directly mapped from hieradata to JCasc in the template
- snake_case for hieradata pure attribiutes
- Kept a few kebab-case when YAML key is mapped to a functionnal field (such as the agent cloud name `azure-vms`)

lemeurherve (approved):

PR #3194, opened 2023-11-27
feat(ci.jenkins.io) disable spot for the whole azure-vm agent cloud

dduportal:

This PR introduces a new hieradata attribute `disableSpot` to allow disabling "Spot" on all VM templates of a given Azure-VM agent cloud.

This new attribute is set to `true` for the new ci.jenkins.io's Azure VM cloud on the secondary subscription, as we don't have enough quota for spot VMs.


smerle33 (approved):
🚀

PR #3212, opened 2023-12-13
cleanup(ci.jenkins.io) remove the ACI cloud (primary azure subscription) to only rely on the sponsored subscription

dduportal:

This PR removes the former subscription to only uses the sponsored one for ACI.


dduportal:
Self merging as we can only "test" this on production

PR #3214, opened 2023-12-14
feat(trusted.ci.jenkins.io) add a secondary Azure VM cloud to spawn agents in the new subscription

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3818#issuecomment-1855950963


dduportal:
Self merging as the only viable test is the production for this

PR #3219, opened 2023-12-15
feat(cert.ci.jenkins.io) add secondary Azure VM subscription

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3818#issuecomment-1857435110


dduportal:
Self merging as we can only test in production

PR #3249, opened 2024-01-10
Remove BO from trusted.ci

NotMyFault:
Following the ongoing cleanup

### DIFF
--- a/hieradata/clients/controller.trusted.ci.jenkins.io.yaml
+++ b/hieradata/clients/controller.trusted.ci.jenkins.io.yaml
@@ -137,7 +137,6 @@ profile::jenkinscontroller::jcasc:
 profile::jenkinscontroller::plugins:
   - ansicolor
   - azure-vm-agents
-  - blueocean
   - build-timeout
   - buildtriggerbadge
   - cloudbees-folder


timja (commented):
I don't think we need to remove from here. weekly sure, but blueocean still has the easiest to use visualisation for viewing logs imo.

NotMyFault:
> I don't think we need to remove from here. weekly sure, but blueocean still has the easiest to use visualisation for viewing logs imo.

Isn't the pipeline stage view plugin enough to view logs on a per-node basis?

timja:
> > I don't think we need to remove from here. weekly sure, but blueocean still has the easiest to use visualisation for viewing logs imo.
> 
> Isn't the pipeline stage view plugin enough to view logs on a per-node basis?

it's awful no thanks

NotMyFault:
> > > I don't think we need to remove from here. weekly sure, but blueocean still has the easiest to use visualisation for viewing logs imo.
> > 
> > 
> > Isn't the pipeline stage view plugin enough to view logs on a per-node basis?
> 
> it's awful no thanks

I mean pipeline graph view, oops

timja:
> I mean pipeline graph view, oops

Somewhat, I still prefer Blueocean on a day to day basis, I find it easier to use going straight into the logs

dduportal:
> > I mean pipeline graph view, oops
> 
> Somewhat, I still prefer Blueocean on a day to day basis, I find it easier to use going straight into the logs

Same here, for the exact same reason. Somehow, if there was a link between the classic "steps" view and its "per-step output" with the pipeline-graph viewer in classic view, that would be nice.

PR #3287, opened 2024-02-08
Remove beer from hieradata

NotMyFault:
Given we overwrite this list on every controller and don't have beer installed anywhere currently, I propose to remove it from here too.


lemeurherve (approved):
Thanks! 🥂

NotMyFault:
I've triggered the job manually on infra.ci, I think my team membership doesn't suffice 😅

dduportal (approved):

PR #3295, opened 2024-02-16
feat(pkg) add `azcopy` CLI to allow improved Azure Storage copy operations

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3414

- chore: a new profile named `azcopy` which role is to ensure `azcopy` is installed in `/usr/local/bin/` when the profile is called. Its content is a code factorization of the current `azcopy` installation tasks from the agent.trusted profile.
- feat: install `azcopy` on the VM `pkg.origin.jenkins.io`.

A few notes:

- The `azcopy` version has been factorized to the top-level `common.yaml` hieradata. The goal is to update all machines at once when the version is bumped.
- As such (change of the `version` attribute location), there will be a subsequent PR to introduce an `updatecli` manifest to track this version (there are new versions so the upgrade should happen on both agent.trusted and pkg machines).
- The [script to generate an SAS short-lived token](https://github.com/jenkins-infra/pipeline-library/pull/839) is not part of this PR: should be a subsequent one


lemeurherve (approved):

dduportal:
Puppet disabled on both agent.trusted.ci and pkg.origin.jenkins.io

dduportal:
Failed to apply with the following error during the unarchiving of `azcopy`:

```
sh: 1: /usr/bin/tar: not found
```

See #3299

PR #3302, opened 2024-02-19
fix(profiles `azcopy` / `buildagent`) fix version checks and tools installation

dduportal:
This PR corrects 2 PRs related to `azcopy`:

- Fixup of #3173 which is not strict enough when checking the `azcopy` version, as the latest available version can be printed alongside the current version in the output of `azcopy --version`
  - Tested locally

- Fixup of #3295 which removed the `$architecture` variable from the profile `buildagent` when factorizing `azcopy`


lemeurherve (approved):
Strange that the grep on the version only didn't work 🤷

dduportal:
> Strange that the grep on the version only didn't work 🤷

I've added the output example in the [commit message](https://github.com/jenkins-infra/jenkins-infra/pull/3302/commits/b90710a6f8807d65e82400dff038efb5acd436c5):

```bash
$ azcopy --version
azcopy version 10.21.0
INFO: azcopy 10.21.0: A newer version 10.23.0 is available to download
```

=> grepping on the `10.23.0` does not fail as this version is in the output string

PR #3322, opened 2024-03-07
feat(pkg,agent): add `azure-cli`

lemeurherve:
This PR adds `azure-cli` if its version is set in hieradata, needed to generate short-lived SAS token for file share with a service principal.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-1952035761
- https://github.com/jenkins-infra/helpdesk/issues/3414#issuecomment-1973817455


dduportal (approved):
Nice work !

PR #3323, opened 2024-03-07
feat(pkg,agent): include `get-fileshare-signed-url.sh` script

lemeurherve:
This PR includes the script from the shared pipeline library to generate short-lived SAS token for a File Share with a service principal so it's available and kept up to date for everything needing it:
- crawler
- update-center2
- mirror-scripts

Follow-up of:
- #3322 

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-1952035761
- https://github.com/jenkins-infra/helpdesk/issues/3414#issuecomment-1973817455

### DIFF
--- a/dist/profile/manifests/azcopy.pp
+++ b/dist/profile/manifests/azcopy.pp
@@ -46,5 +46,11 @@ class profile::azcopy (
       ensure  => "${az_cli_version}-1~${facts['os']['distro']['codename']}",
       require => Class['apt::update'],
     }
+
+    file { '/usr/local/bin/get-fileshare-signed-url.sh':
+      ensure => file,
+      mode   => '0755',
+      source => 'https://raw.githubusercontent.com/jenkins-infra/pipeline-library/master/resources/get-fileshare-signed-url.sh',
+    }
   }
 }


lemeurherve:
Example of usage:
- https://github.com/jenkins-infra/crawler/pull/144

dduportal (approved):
Lgtm.

just to be sure : do we agree that:
- any change in the pipeline library (such as branch renaming) would break this?
- any change in the script content would be deployed blindly as soon the puppet agent re-runs?

lemeurherve:
> Lgtm.
> 
> just to be sure : do we agree that:
> 
> * any change in the pipeline library (such as branch renaming) would break this?
> 
> * any change in the script content would be deployed blindly as soon the puppet agent re-runs?

Yes indeed.

This is a first iteration, with the constrains you mentioned.

In a second phase I'll put this script in this repository, with an updatecli manifest to keep it up to date while also allowing us to control when it will be updated.

lemeurherve:
> > * any change in the pipeline library (such as branch renaming) would break this?
> > * any change in the script content would be deployed blindly as soon the puppet agent re-runs?
>
> I'll put this script in this repository, with an updatecli manifest to keep it up to date while also allowing us to control when it will be updated.

Addressed in https://github.com/jenkins-infra/jenkins-infra/pull/3344

PR #3324, opened 2024-03-08
chore(updatecli): fix `azcopy` target

lemeurherve:
This PR fixes the target of `azcopy` updatecli manifest.

Fixup of:
- #3322


lemeurherve:
Tested locally, passing here, self-merging to restore updatecli on main.

PR #3342, opened 2024-03-21
feat(ci.jenkins.io): Adding a label `linux-arm64` for arm64 VM to match infra.ci

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3823#issuecomment-1961400721
allow to use the same label for both ci and infra.ci unified pipeline for ARM64


lemeurherve (approved):

dduportal (approved):

dduportal:
@smerle33 we let you merge when you are ready (and when the puppet agent on ci.jenkins.io is re-enabled)

PR #3343, opened 2024-03-21
fix(ci.jenkins.io): Correct the label for jdk tools to match the arm labels

smerle33:
the tool for jdk on ARM64 agents is using the wrong label for now `linux` is amd64 only (to avoid mixing platform at runtime)


MarkEWaite (approved):
Good catch.  The "linux" label is assumed by many locations to be specific to AMD64.

dduportal (approved):

dduportal:
@smerle33 we let you deploy when you are ready

Issue #3353, opened 2024-03-28
[pkg.origin.jenkins.io] Cleanup user `www-data`


dduportal:
Since https://github.com/jenkins-infra/helpdesk/issues/3414, the user `www-data` should not be used to run SSH/Rsync commands anymore, but only by the Apache2 webserver on the VM `pkg.origin.jenkins.io`.

As such, some cleanup is required to remove the ssh keys and other "exotic" resources we could find not related directly to the Apache2 setup.

dduportal:
Additionnaly, puppet fails applying on `pkg.origin.jenkins.io` since https://github.com/jenkins-infra/jenkins-infra/pull/3348 with the following errors:

```
(/Stage[main]/Profile::Updatesite/Ssh_authorized_key[updatesite-key]/target) target changed '/srv/releases/.ssh/authorized_keys' to '/var/www/.ssh/authorized_keys' (corrective)
Mar 29 17:48:34 ip-172-31-33-96 puppet-agent[5045]: Puppet::Util::FileType::FileTypeFlat could not write /srv/releases/.ssh/authorized_keys: Permission denied @ rb_sysopen - /srv/releases/.ssh/authorized_keys
Mar 29 17:48:34 ip-172-31-33-96 puppet-agent[5045]: (/Stage[main]/Profile::Updatesite/Ssh_authorized_key[updatesite-key]) Could not evaluate: Puppet::Util::FileType::FileTypeFlat could not write /srv/releases/.ssh/authorized_keys: Permission denied @ rb_sysopen - /srv/releases/.ssh/authorized_keys
Mar 29 17:48:34 ip-172-31-33-96 puppet-agent[5045]: (/Stage[main]/Profile::Updatesite/File[/var/www/updates.jenkins.io]/mode) mode changed '0775' to '0755' (corrective)
```

dduportal:
#3365 was deployed with success. The errors are gone and both `sync.sh` and update-center2 are working as expected.

PR #3365, opened 2024-03-29
fix(`pkg.origin.jenkins.io`) ensure user www-data cannot login and does not have any SSH access

dduportal:
Ref. https://github.com/jenkins-infra/jenkins-infra/issues/3353

This PR cleans up the user `www-data` on the VM `pkg.origin.jenkins.io with the following changes:

- Disable its shell login
- Removes any SSH authentication data (moving to the `mirrorbrain` user)
- Cleans up unused key (profile::jenkins::mirroring_privkey)

Deployment process:

- Disable Puppet agent on the VM
- Run a `noop` operation to ensure that the expected changes are shown
- Apply and ensure that the errors from #3353 are fixed
- Cleans up the `/var/www/.ssh` directory
- Re-run the puppet agent to ensure no more changes
- Watch the `sync.sh` and trusted.ci's update)center2 jobs


dduportal:
Agent disabled on the VM, self merging to deploy

dduportal:
Deployed with success!

PR #3372, opened 2024-04-04
cleanup: remove superseded blobxfer variable and checks

lemeurherve:
This PR removes the `AZURE_STORAGE_ACCOUNT` variable previously used by blobxfer in pkg sync scripts (its value is also in `STORAGE_NAME`, used by the get-fileshare-signed-url.sh script) and associated checks that were added to pass shellcheck before implementing azcopy as replacement of blobxfer in https://github.com/jenkins-infra/mirror-scripts/pull/14.

Verification procedure after merging this PR: ensure update_center job still passes on trusted.ci.jenkins.io

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3414#issuecomment-2034806598

### DIFF
--- a/dist/profile/manifests/pkgrepo.pp
+++ b/dist/profile/manifests/pkgrepo.pp
@@ -141,7 +141,6 @@ Host ftp-osl.osuosl.org
     group   => $mirror_group,
     mode    => '0600',
     content => "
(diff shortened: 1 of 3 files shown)


dduportal (approved):
You may merge and deploy when you are ready to check the usual script executions on `pkg` VM

PR #3483, opened 2024-06-19
chore(updatecli) use native YAML and HCL resources feature to parse remote files or checking keys in conditions

dduportal:
While reviewing https://github.com/jenkins-infra/contributor-spotlight/pull/146, I realized that we could use the (recent?) native features of `updatecli` YAML and HCL to simplify the setups.

Why this change? https://blog.codinghorror.com/regular-expressions-now-you-have-two-problems/ => we want to avoid overusing regex when not needed. As `updatecli` evolved to provide us the "remote YAML file" or "remote HCL" ability, we can get rid of the tedious pattern matching + transformer submatch.

Additionally, we have some historical conditions over the existence of YAML keys which can now benefit from the `keyonly` attribute, unlocking the native YAML parsing (instead of tedious pattern matching, again).

### DIFF
--- a/updatecli/weekly.d/docker-ce.yaml
+++ b/updatecli/weekly.d/docker-ce.yaml
@@ -24,11 +24,12 @@ sources:
 
 conditions:
   checkIfHieradataCommonHasDockerVersionKey:
-    kind: file
+    kind: yaml
     disablesourceinput: true
     spec:
       file: hieradata/common.yaml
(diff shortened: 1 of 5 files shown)


lemeurherve (approved):
🚀 

Doc page containing "keyonly": https://www.updatecli.io/docs/plugins/resource/yaml/

PR #3515, opened 2024-07-08
feat(cert.ci.jio,trusted.ci.jio) rotate Azure Client passwords used for L.E. validation

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4153


dduportal (commented):
Self merging for deployment

PR #3528, opened 2024-07-18
feat(jenkinscontroller) use native git tool on cert.ci.jio but keep default jgit tool on ci.jio and trusted.ci.jio

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4186#issuecomment-2236868804

This PR introduces a new customization for controller Jenkins `git` tool.

By default, we keep the Jgit tool, but you can disable it with a boolean passed to hieradata.

cert.ci.jenkins.io switches to native `git` CLI as per the linked issue, while we keep the default behavior for ci.jenkins.io and trusted.ci.jenkins.io.


smerle33 (approved):
🚀

PR #3551, opened 2024-07-25
chore(agent-images/all controllers): unencrypt subscription id for azure vms gallery

smerle33:
as this information is already public, no need to encrypt, it will make life easier with vagrant.


dduportal (approved):
For info:

- Comes from https://github.com/jenkins-infra/azure/blob/8a94cf7d8bc19bd1e48566c804e512b3def80ed1/packer-resources.tf#L72-L73
- https://github.com/jenkins-infra/azure/blob/main/providers.tf#L9

PR #3564, opened 2024-07-31
feat(profile::jenkinscontroller): Added template for jdk-17 and jdk-21 SSH agents

jayfranco999:
As per https://github.com/jenkins-infra/helpdesk/issues/4124#issue-2335345048:

Modified JCasC ERB template for jdk ssh agents

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
@@ -120,8 +120,14 @@ jenkins:
             systemctl enable jenkins-agent
             systemctl start jenkins-agent || systemctl status jenkins-agent
           ) 2>&1 | tee /var/log/agent-init-jenkins.log
-          <%-end -%>
 
+          <%- else -%>
+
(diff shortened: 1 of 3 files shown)


dduportal (changes requested):
This change must be grouped with #3565 and #3566 as it is not deliverable alone:

- If we merge the change here, we'll change the default JDK of trusted.ci and cert.ci Azure VM Agents from JDK11 to JDK17 which will break all the builds.
- If we merge only #3565 (or #3566) then all the templates will use jdk11 until we merge the change here.


=> I propose that we'll work on cherry-picking the commits #3565 and #3566 in the PR here, we'll close them and we'll work with only a single PR.

dduportal (approved):
Last changes (adding javaHome for cert and trusted) is 👍 

Let's deploy (pairing session the 2 of us)

dduportal:
Deployed with success to both cert.ci and trusted.ci (ci.jio is NOT impacted by this change).

We checked that the default template (with a pipeline `node('linux') { ...}`) is still using JDK11 after this change (it's only made explicit now).

PR #3568, opened 2024-08-01
feat(jenkinscontroller) use Docker registry mirror for Azure VM VM agents

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4192#issuecomment-2263339493

This PR adds setup, on all controllers, of the Docker Engine (in Azure VM agents) to use the new ACR registry as a mirror cache to decrease impact on DockerHub (it's a common good!) and limit risks of rate limiting.

Note: it also fixes Azure VM arm64 invalid labels caught during the pairing session with @timja 

Tested manually on ci.jenkins.io with success (first on Linux, then on Windows): if you run `docker info` on these machines, you should see the registry mirror setup in the output


timja (approved):

dduportal:
Puppet agent disabled on all controllers: let's deploy manually, step by step

PR #3578, opened 2024-08-07
feat(controllers) ensure the plugin disable-job-button is installed

dduportal:
cc @daniel-beck

PR #3579, opened 2024-08-07
Revert adding disable-job to ci.jenkins.io

timja:
see https://github.com/jenkins-infra/jenkins-infra/pull/3578#issuecomment-2273528093

This goes again the removal of it and ci.jenkins.io being an exemplar instance as well.

The disable-job-button plugin reduces a couple of clicks for when people are frequently disabling jobs. I'm not aware of jobs ever being disabled on ci.jenkins.io and if they are the extra click or two shouldn't be a pain for the rarity of it


dduportal (approved):
Ack. , make sense. I'm not aware of the reason to remove this feature from core, it is not immediate for common users like me :) 

But make sense!

PR #3589, opened 2024-08-08
hotfix(pkg.jio/updates.jio) pin az CLI to 2.60.0

dduportal:
… as the last released version for Ubuntu 18.04 Bionic


dduportal:
Self merging: already applied

PR #3691, opened 2024-10-16
Update maven-3.9 from 3.9.4 to 3.9.9

daniel-beck:


MarkEWaite (approved):

dduportal (approved):
Thanks Daniel! Should we keep these versions up to date automatically (using automated PRs, patch updates only of course, for each minor line) ?

daniel-beck:
> Thanks Daniel! Should we keep these versions up to date automatically (using automated PRs, patch updates only of course, for each minor line) ?

I don't see why not, generally.

PR #3730, opened 2024-11-18
feat(ci.jenkins.io) set up initial `aws.ci.jenkins.io` VM client data

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4315

Requires https://github.com/jenkins-infra/terraform-aws-sponsorship/pull/37 to be merged and deployed succerfully.

This PR follows up #3725 but does NOT change its content.

However, it does reimport the history from the current Azure ci.jenkins.io controller to make sure future `git history` / `git blame` will have the whole history.
It also has a bunch of commits which goal is to document "why" each change is made from the Azure setup.


dduportal:
LGTM as per https://github.com/jenkins-infra/helpdesk/issues/4315#issuecomment-2483626190

PR #3823, opened 2025-01-16
feat(jenkinscontroller) add support of EC2 Windows VMs using WinRM connection

dduportal:
This PR introduces support of WinRM (instead of SSH) communicator with the EC2 plugin (e.g. specifying "Windows Launcher" instead of "Unix Launcher").

It also fixes the `PATH` environment variable for both Linux and Windows agents.

Related to https://github.com/jenkins-infra/helpdesk/issues/4316#issuecomment-2593374355

(edit) Tested with local vagrant, and then applying excerpt of the configuration to aws.ci.jenkins.io. No regression possible on the current Azure/Kube for ci.jenkins.io and other controllers (already tested in production). Any error would only require a fix for aws.ci.jenkins.io.


smerle33 (dismissed):
LGTM 🚀

smerle33 (approved):
🚀

PR #3876, opened 2025-02-12
fix(openvpn) renew server-side certificate

dduportal:
hotfix


smerle33:
ref https://github.com/jenkins-infra/helpdesk/issues/4537

PR #3916, opened 2025-02-27
feat(ci.jio) add back the permanent agent s390x forgotten on the new ci.jenkins.io

smerle33:
during the migration from azure to aws, the permanent agent s390x has been forgotten ... here it is back in work.


MarkEWaite (approved):
Thanks!

PR #3958, opened 2025-03-19
feat(ci.jenkins.io) add spot instance agent

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4598

creating a spot instance (as ssh agent for now)


smerle33:
tested locally with vagrant

PR #3959, opened 2025-03-19
feat(ci.jenkins.io): add spot instances versions for ec2 agents

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4598

duplicate vm agent on ec2 as spot instances with specific labels: only those used by buildplugin(), jenkinsci/docker-*, jenkins/core, ATH will exist as non-spot and spot instances.
All the other will be spot only to save cash

### DIFF
--- a/hieradata/clients/aws.ci.jenkins.io.yaml
+++ b/hieradata/clients/aws.ci.jenkins.io.yaml
@@ -393,7 +393,7 @@ profile::jenkinscontroller::jcasc:
         agent_definitions:
           ####
           # x86_64 Linux VMs
-          - description: "Ubuntu 22.04 LTS x86_64 with JDK8 set as default java CLI"
+          - description: "Spot Ubuntu 22.04 LTS x86_64 with JDK8 set as default java CLI"
             maxInstances: 50
             instanceType: M6aXlarge # 4 vCPUs / 16 Gb
             os: "ubuntu"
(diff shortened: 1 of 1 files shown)


hieradata/clients/aws.ci.jenkins.io.yaml
```diff
@@ -403,33 +403,45 @@ profile::jenkinscontroller::jcasc:
...
+              - linux
+              - java
+              - docker
+              - linux-amd64
             spot: false
             acpServerId: aws-internal
-          - description: "Ubuntu 22.04 LTS x86_64 with JDK17 set as default java CLI"
+          ####
+          # Linux VMs jdk17 (default spot version for buildPlugin() / docker-*)
+          - description: "Spot Ubuntu 22.04 LTS x86_64 with JDK17 set as default java CLI"
```
dduportal: We should invert the logic and define "spot" as the default while "ondemand" should be opt-in

hieradata/clients/aws.ci.jenkins.io.yaml
```diff
@@ -403,33 +403,45 @@ profile::jenkinscontroller::jcasc:
             javaHome: /opt/jdk-8
             labels:
               - ubuntu-22-amd64-maven8
-            spot: false
+              - spot
```
dduportal: We should not specify the label as it, but add a suffix `-spot` (or `-nonspot` or `-ondemand` if we follow my suggestion below).

Otherwise someone requesting a `spot` won't know if it is a Windows or Linux, or Intel, or JDK17 machine.

dduportal (approved):

PR #3968, opened 2025-03-21
feat(ci.jenkins.io) add support for S3 Artifact Manager and enable it

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4596


dduportal:
Merging as agent is disabled on ci.jio and change already applied

PR #3969, opened 2025-03-21
feat(ci.jenkins.io): use nvme disk if available for workspace

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4597

tested on a specific agent smerle:

```
+ df -h
15:39:25  Filesystem       Size  Used Avail Use% Mounted on
15:39:25  /dev/root        146G   11G  136G   7% /
15:39:25  tmpfs            7.8G     0  7.8G   0% /dev/shm
15:39:25  tmpfs            3.1G  8.9M  3.1G   1% /run
15:39:25  tmpfs            5.0M     0  5.0M   0% /run/lock
15:39:25  efivarfs         128K  4.1K  119K   4% /sys/firmware/efi/efivars
15:39:25  /dev/nvme1n1p15  105M  6.1M   99M   6% /boot/efi
15:39:25  /dev/md127       140G  1.1G  139G   1% /home/jenkins/agent
15:39:25  tmpfs            1.6G  4.0K  1.6G   1% /run/user/1001
```

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
@@ -479,6 +479,8 @@ jenkins:
           #!/bin/sh
           set -eux
 
+          systemctl stop sshd.service
+
           ## Setup Datadog service
           (
             systemctl stop datadog-agent.service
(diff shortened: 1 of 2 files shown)


dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
```diff
@@ -494,6 +496,83 @@ jenkins:
             systemctl start datadog-agent.service
           ) 2>&1 | tee /var/log/agent-init-datadog.log
 
+          ## Setup NVMe if exist
+          MNT_DIR="<%= agent['agentDir'] ? agent['agentDir'] : @jcasc['agents_setup'][$os]['agentDir'] %>"
```
timja, line 501: is it better to have this in a script in a repository and download it on startup? then you can shell check it and don't have to escape every line basically?

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
```diff
@@ -479,6 +479,8 @@ jenkins:
           #!/bin/sh
           set -eux
 
+          systemctl stop sshd.service
```
timja, line 483: why?

dduportal:
```suggestion
          # Ensure Jenkins controller does not start the agent in process in its workspace until we are ready
          systemctl stop sshd.service
```

> why?

Cloud Init might not be finished when Jenkins establish the SSH connection and start writing in the agent workspace. If it happens before mounting the NVMe, it will fail the agent.

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
dduportal: Yes, that would avoid escaping, but it creates a dependency to another webserver. It's a good optim. but for later.

dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
smerle33: by locking the sshd at the beginning and restarting it at the end, we ensure that the agent isn't seen as ready as long as the cloud-init is not done ... (it's already assumed by the loop within the initscript but this loop isn't working on windows) as we set the new mount point as the workspace, we need the agent to wait until the mount is done, not to write in the wrong spot

dduportal (approved):
Tested in pair with success!

PR #3996, opened 2025-03-27
feat(ci.jenkins.io): aws cloud init script to format and mount NVMe disk on windows

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4597

This PR ensure that the sshd service is stopped at startup (it may be too late, so make sure to have a 120s boot delay for ssh on windows agents) and then, if there is NVMe disk, it will Initialize them and mount them on the agentDir.


dduportal (dismissed):

smerle33:
tested multiple time in https://ci.jenkins.io/job/Infra/job/acceptance-tests/job/infra-checks

PR #4001, opened 2025-03-28
feat(ci.jenkins.io): windows agent new mount point on Z for NVMe

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4612
we discovered that docker is not able to use folder mounted volume, so we are now using Z: 
and set this as the AgentDir (Remote Fs)


dduportal (approved):
LGTM! As discussed together we'll be able to eventually improve the code (factorizing, extracting the later from the `agentDir` string, etc.) but it is good enough to provide a fix in production 👍

dduportal:
Merging as the updatecli check failure is unrelated (packer image)

PR #4037, opened 2025-04-08
cleanup(jenkinscontroller) remove Azure Container Agents setups

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4554

Let's get rid of code we won't use anymore

### DIFF
--- a/hieradata/rspec/profile_jenkinscontroller.yaml
+++ b/hieradata/rspec/profile_jenkinscontroller.yaml
@@ -150,28 +150,9 @@ profile::jenkinscontroller::jcasc:
           usePrivateIP: true
           spot: false
           agentDir: 'C:/Foo'
-    azure-container-agents:
-      clouds:
-        aci-windows:
-          credentialsId: azure-credentials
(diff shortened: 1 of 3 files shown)


hieradata/rspec/profile_jenkinscontroller.yaml
```diff
@@ -150,28 +150,9 @@ profile::jenkinscontroller::jcasc:
...
-          cpus: 4
-          memory: 8
-          agentJavaBin: java # From image maven:3.8.6-eclipse-temurin-11
-          labels:
-            - maven-11-windows
 # These are plugins we need in our ci environment
 profile::jenkinscontroller::plugins:
-  - azure-container-agents
   - azure-vm-agents
-  - blueocean
```
smerle33, line 174: are you sure ? it's still in usage on ci ... those this doesn't need to be discuss first with users ?

dduportal: Double check the filename in which this has been removed :)

PR #4146, opened 2025-05-23
fix(profile::jenkins::agent) ensure Temurin JDKs are updated when their version changes

dduportal:
While working on https://github.com/jenkins-infra/helpdesk/issues/4121#issuecomment-2890355054, we caught that the JDK21 in the trusted permanent agent was still using an old JDK21 EA version.

It appears that the Puppet resource "Archive" only checked for the existence of the `java` bin on each installed JDK. So once installed, it was never updated.

This PR changes it to an exec command which performs checks on the JDK version.

### DIFF
--- a/dist/profile/manifests/buildagent.pp
+++ b/dist/profile/manifests/buildagent.pp
@@ -97,12 +97,11 @@ class profile::buildagent (
         cpu_arch => $facts['os']['architecture'],
       }
       $java_dir = "/opt/jdk-${$jdk['major_version']}"
-
+      $java_bin = "${java_dir}/bin/java"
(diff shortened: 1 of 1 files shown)


smerle33 (approved):
🚀

dduportal:
This PR failed to deploy: it appears that the Adoptium URL generated in the specific case of the `buildagent` profile are specifying `amd64` as the CPU architecture in the Adoptium URLs. But Adoptium uses the `x64` architecture code (for `amd64` class.

The local test was done on a macOS silicon machine which as an`aarch64` reported CPU: Adoptium uses the same code in that case so it worked flawlessly. But the production VM is `amd64`.


How could this code have worked in the past? I don't know. But incoming fix!

PR #4164, opened 2025-06-04
break! (ci.jenkins.io): use jdk21 instead of jdk17 as default for javahome/path

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4127#issue-2335969330


dduportal (dismissed):
Good job 👏

PR #4165, opened 2025-06-04
feat(ci.jio): migrate jnlp agent from inbound-agent to allinone

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4127#issue-2335969330

migrate jnlp to allinone in jdk21

### DIFF
--- a/hieradata/clients/controller.ci.jenkins.io.yaml
+++ b/hieradata/clients/controller.ci.jenkins.io.yaml
@@ -243,7 +243,7 @@ profile::jenkinscontroller::jcasc:
               - jdk21
               - jnlp
               - default
-            useAsMuchAsPossible: true # Default template to sue when `node() {}` step is called without labels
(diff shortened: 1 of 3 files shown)


hieradata/clients/controller.ci.jenkins.io.yaml
```diff
@@ -241,6 +241,9 @@ profile::jenkinscontroller::jcasc:
             labels:
               - maven-21
               - jdk21
+              - jnlp
+              - default
+            useAsMuchAsPossible: true # Default template to sue when `node() {}` step is called without labels
```
dduportal:
```suggestion
            useAsMuchAsPossible: true # Default template to use when `node() {}` step is called without labels
```

dduportal (dismissed):

dduportal (approved):
🚀

PR #4172, opened 2025-06-10
chore(updatecli): stopping tracking jnlp container as migrated to allinone

smerle33:
since https://github.com/jenkins-infra/jenkins-infra/pull/4165 we don't use the agent jnlp container, so we can stop tracking it's updates


dduportal (approved):
❤️

Issue #3250, opened 2024-01-10
Use non-deprecated agent URL arguments


basil:
While searching the `jenkins-infra` organization for references to the deprecated `-jnlpUrl` option I found a few in this repository. I _think_ this is what needs to be done to migrate, but I have no way of testing this change so I have not created a PR:

```powershell
diff --git a/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb b/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
index 77f2de9a..1d4403e6 100644
--- a/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
@@ -41,7 +41,7 @@ jenkins:
           # Prepare config
           Write-Output 'Executing agent process...'
           $configExec = '<%= agent['agentJavaBin'] ? agent['agentJavaBin'] : @jcasc['agents_setup'][agent['os'].to_s]['agentJavaBin'] %>'
-          $configArgs = '-jnlpUrl "{0}/computer/{1}/jenkins-agent.jnlp" -workDir {2}' -f $jenkinsserverurl, $agentName, $jenkinsWorkDir
+          $configArgs = '-url {0} -name {1} -workDir {2}' -f $jenkinsserverurl, $agentName, $jenkinsWorkDir
           if ($agentSecret) {
               $configArgs += " -secret `"$agentSecret`""
           }
@@ -85,7 +85,6 @@ jenkins:
             export AGENT_WORKDIR='<%= agent['agentDir'] ? agent['agentDir'] : @jcasc['agents_setup'][agent['os'].to_s]['agentDir'] %>'
             export AGENT_JAR="^${AGENT_WORKDIR}/agent.jar"
             export AGENT_SECRETFILE="^${AGENT_WORKDIR}/agent-secret"
-            export AGENT_URL="^${JENKINS_URL}computer/^${AGENT_NAME}/jenkins-agent.jnlp"
             export JENKINS_JAVA_OPTS='<%= agent['agentJavaOpts'] ? agent['agentJavaOpts'] : @jcasc['agents_setup'][agent['os'].to_s]['agentJavaOpts'] %>'
             export ARTIFACT_CACHING_PROXY_PROVIDER='azure'
[...]

PR #4188, opened 2025-06-17
fix(jenkinscontroller): Use non-deprecated inbound agent URL arguments

zhan9san:
Fix 
* #3250
* https://github.com/jenkins-infra/helpdesk/issues/3912

@dduportal @basil Could you please help review it?

I noticed that these issues have been around for about 1.5 years.
This seems like a great opportunity for me to contribute to the Jenkins infrastructure.

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb
@@ -61,7 +61,7 @@ jenkins:
           # Prepare config
           Write-Output 'Executing agent process...'
           $configExec = '<%= agent['agentJavaBin'] ? agent['agentJavaBin'] : @jcasc['agents_setup'][$os]['agentJavaBin'] %>'
-          $configArgs = '-jnlpUrl "{0}/computer/{1}/jenkins-agent.jnlp" -workDir {2}' -f $jenkinsserverurl, $agentName, $jenkinsWorkDir
+          $configArgs = '-url "{0}" -name {1} -workDir {2}' -f $jenkinsserverurl, $agentName, $jenkinsWorkDir
           if ($agentSecret) {
               $configArgs += " -secret `"$agentSecret`""
           }
@@ -128,7 +128,6 @@ jenkins:
             export AGENT_WORKDIR='<%= agent['agentDir'] ? agent['agentDir'] : @jcasc['agents_setup'][$os]['agentDir'] %>'
             export AGENT_JAR="^${AGENT_WORKDIR}/agent.jar"
             export AGENT_SECRETFILE="^${AGENT_WORKDIR}/agent-secret"
-            export AGENT_URL="^${JENKINS_URL}computer/^${AGENT_NAME}/jenkins-agent.jnlp"
             export JENKINS_JAVA_OPTS='<%= agent['agentJavaOpts'] ? agent['agentJavaOpts'] : @jcasc['agents_setup'][$os]['agentJavaOpts'] %>'
             <%- if @jcasc['artifact_caching_proxy'] && @jcasc['artifact_caching_proxy']['enabled'].to_s == "true" && cloudsetup['acpServerId'] -%>
             export ARTIFACT_CACHING_PROXY_SERVERID='<%= @jcasc['artifact_caching_proxy']['servers'][cloudsetup['acpServerId']]['url'] %>'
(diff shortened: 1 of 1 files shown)


dduportal:
Thanks @zhan9san!

May I ask how did you test this change (or shall we test it)?

zhan9san:
@dduportal 
Thanks for your attention.

I don't have permission to test it in jenkins-infra's real environment.

But i test the new parameter -url on my local jenkins, and it works.

Here is what i tested.

1. Start a jenkins server
2. Create a jnlp agent on the jenkins server
3. On another server, connect the jenkjns controller with -url
4. Check whether the agent can run successfully

basil:
Looks right at first glance, pending testing

dduportal:
Thanks! 

In order to limit the scope of impact, while ensuring we do partial testing and quick delivery of this contribution,
we'll now apply this change only on one of the ci.jenkins.io Linux VM agents (and on one of the Windows VM agents) and we'll test a pipeline replay allocating this, with a long running process (so we can capture the process tree on both and validate everything)



 I'll report here the results!

dduportal (approved):
This change looks really good! Let's proceed!

Many thanks

zhan9san:
@dduportal @basil 

Thanks for your help.

I have a follow-up question:
Would it be possible to use the SSH launch method for Azure nodes as well?

I noticed that the Amazon EC2 agents use SSH, while Azure agents are currently using JNLP.
https://github.com/jenkins-infra/jenkins-infra/blob/cc22f93170e3674ce8505583ee856c7715b0dea7/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb#L582

Furthermore, I’m wondering if we could standardize on using the SSH launch method for both Linux and Windows agents.
Technically, it works well on both platforms based on my testing.

Let me know your thoughts on this approach.

> For the launch method, select SSH or JNLP.
> For Windows agents, if the launch method is SSH then check Pre-Install SSH in Windows Agent or image needs to be custom-prepared with an SSH server pre-installed.
> We recommend to use SSH rather than connecting it to the controller, it's simpler to setup and get much clearer logs.

Reference: https://plugins.jenkins.io/azure-vm-agents/#plugin-content-use-a-custom-vm-image

dduportal:
> @dduportal @basil
> 
> Thanks for your help.
> 
> I have a follow-up question: Would it be possible to use the SSH launch method for Azure nodes as well?
> 
> I noticed that the Amazon EC2 agents use SSH, while Azure agents are currently using JNLP.
> 
> https://github.com/jenkins-infra/jenkins-infra/blob/cc22f93170e3674ce8505583ee856c7715b0dea7/dist/profile/templates/jenkinscontroller/casc/clouds.yaml.erb#L582
> 
> Furthermore, I’m wondering if we could standardize on using the SSH launch method for both Linux and Windows agents. Technically, it works well on both platforms based on my testing.
> 
> Let me know your thoughts on this approach.
> 
> > For the launch method, select SSH or JNLP.
> > For Windows agents, if the launch method is SSH then check Pre-Install SSH in Windows Agent or image needs to be custom-prepared with an SSH server pre-installed.
> > We recommend to use SSH rather than connecting it to the controller, it's simpler to setup and get much clearer logs.
> 
> Reference: https://plugins.jenkins.io/azure-vm-agents/#plugin-content-use-a-custom-vm-image

Azure VM agents can use both methods if you look at the Puppet template code carefully. It depends on the network topology and controllers: we have cases where the controller cannot reach agents unless they got a public IP (which we do not want).

In EC2, using SSH is the default, but as soon as you switch to Spot instances, it requires using inbound agent method as per https://plugins.jenkins.io/ec2/#plugin-content-configure-jenkins-for-spot-support.

=> So there are no reason to stop using any of these.
What was the problem you wanted to solve exactly?


NOTE: if you want to help on the Jenkins infrastructure, I suggest you join one of our meeting every Tuesday - see in https://www.jenkins.io/events/#event-calendar and/or join our Elements Matrix channel (https://www.jenkins.io/chat/#jenkins-inframatrix-org) to ask questions/discuss. We also have an issue tracker at jenkins-infra/helpdesk which we use to describe issues/problem (discuss solutions and prioritize)
Sending PRs out of the wild is a risk that it won't be prioritized or even looked at, and we can help you / give you guidance. And discussing on closed PR/issues is also not the best for visibility

zhan9san:
> It depends on the network topology and controllers: we have cases where the controller cannot reach agents unless they got a public IP (which we do not want).

That makes sense—thanks for the clear explanation.

Regarding the meetings, the timing is a bit challenging for me since I’m in UTC+8. I usually catch up by watching the recordings on YouTube and reviewing the shared Google Docs. I really appreciate the effort you put into making those resources available.

> if you want to help on the Jenkins infrastructure

I'll try my best.

dduportal:
> Regarding the meetings, the timing is a bit challenging for me since I’m in UTC+8.


Oh good grief. I'll bring the subject on the next meeting to see if we can have a changed timing (once every 2 weeks for instance). We already have a team member in India so it might be also beneficial for them as well!

PR #4204, opened 2025-06-23
feat(mirror-script) export rsync filter list for sync.sh to allow mirror sponsors to only keep 1 year of artifacts

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4681

A bit of (chore/nitpick) shell cleanup (using lower case and meaningful name for non environmental variables, fixing comments, etc.)


dduportal:
Required a fixup: https://github.com/jenkins-infra/jenkins-infra/commit/6697eebe7c8661e85329ea457dca831d2c552ec3

Because the rsync filter must have the relative path of files

PR #4263, opened 2025-07-17
fix(jenkinscontroller/ec2-clouds) specify tempdir to agent process and do not override init script tempdir

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4730

This PR introduces 2 fixes related to the `ec2` agent clouds in the `jenkins::controller` profile:

- Stop specifying the `tmpDir` JCasC attributes which overrides the default init script temp. dir. It avoids concurrency issues between the "cloud init" steps (user data field, run as root/admin during Vm boot phase) and the init script itself. For instance: specifying `Z:/Temp` here may happen before the Z:  drive is mounted.
- Specify the resolved temp dir. for the agent process using Java flags. We already did use this resolved value for the environment variables (and it should be used by pipeline shell steps on all OSes) but the agent may also benefit from a custom tmp when specified


smerle33 (approved):
checked in vagrant: `-XX:+PrintCommandLineFlags -Djava.io.tmpdir=/tmp`

PR #4323, opened 2025-08-07
fix(ci.jenkins.io.yaml) correct tool label matching to cover the non default `arm64` introduced with the AWS migration

dduportal:
When me moved ci.jenkins.io from Azure to AWS, we had to introduce spot/non spot instances, along with JDK21 additionally.
It led to unplanned cases met by @daniel-beck where there were missing labels.

As explained in  https://github.com/jenkins-infra/helpdesk/issues/4762#issuecomment-3163158212, I'm not sure if the tool specified labels are matched against the "pipeline/user requested labels" or against the "effective labels of the agent taken for the job/stage" => as such i would like multiple review of this proposed change to be sure it works as expected

cc @MarkEWaite @daniel-beck @timja

### DIFF
--- a/hieradata/clients/aws.ci.jenkins.io.yaml
+++ b/hieradata/clients/aws.ci.jenkins.io.yaml
@@ -60,13 +60,13 @@ profile::jenkinscontroller::jcasc:
         installers:
           linux-arm64:
             type: "zip"
-            label: "linux-arm64 && arm64"
+            label: "(linux-arm64 && arm64) || (linux && arm64)"
             cpu_arch: "aarch64"
       jdk11:
         installers:
           linux-arm64:
             type: "zip"
-            label: "linux-arm64 && arm64"
+            label: "(linux-arm64 && arm64) || (linux && arm64)"
             cpu_arch: "aarch64"
           s390x:
             type: "zip"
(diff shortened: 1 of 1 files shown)


hieradata/clients/aws.ci.jenkins.io.yaml
```diff
@@ -60,13 +60,13 @@ profile::jenkinscontroller::jcasc:
         installers:
           linux-arm64:
             type: "zip"
-            label: "linux-arm64 && arm64"
+            label: "(linux-arm64 && arm64) || (linux && arm64)"
```
timja: the right hand-side makes the most sense from a label selector point of view ( `linux && arm64`)

`linux-arm64 && arm64` seems weird, linux-arm64 already requires arm64, why is arm64 required as well.

timja (dismissed):

hieradata/clients/aws.ci.jenkins.io.yaml
dduportal: Good point. I chose to be conservative and only focused on the things added, but what you underline totally make sense. Let me update the PR

dduportal: https://github.com/jenkins-infra/jenkins-infra/pull/4323/commits/8e335eab43627d1799ad011d3b962599526aad8c

timja (approved):

PR #4330, opened 2025-08-09
fix(jenkinscontroller/ec2-clouds) ensure cloud init marker file are properly created with their parent directories from the temp dir value

dduportal:
Following #4329, while working on https://github.com/jenkins-infra/helpdesk/issues/4730, it appears that our configuration for both Linux and Windows wasn't taking care or creating (recursively) the directory where the marker file for cloud init is expected.


dduportal (commented):
Tested with local vagrant, and on a manual try on ci.jenkins.io

PR #4344, opened 2025-08-14
feat(archives,pkg) switch OSUOSL mirror sync from 'pkg' to 'archives'

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4653

- `pkg.origin.jenkins.io` VM: 
  - Remove all OSUOSL related resources (as per https://github.com/jenkins-infra/helpdesk/issues/4653):
    - No more SSH setup (key, config, etc.)
    - Both `sync.sh` and `sync-recent-releases.sh` scripts are 
      - Updated to copy data directly (and synchronously) to `archives.jenkins.io` instead of OSUOSL
      - Verified with shellcheck (once rendered)
    - Ensure following "mirror-script" files are removed: `/srv/releases/populate-archives.sh` (archives.jenkins.io` is directly populated by the scripts now) and `/srv/releases/rsync.filter` (migrated to `archives.jenkins.io` VM)
  - Fix a warning about a confusing escape (for puppet)
  - Set up SSH for `mirrorbrain` user to always use `archives.jenkins.io` domain, along with its known host (to avoid ssh/rsync command to eventually be stuck with an interactive prompt, or failing due to mismatch of known_hosts)

- `archives.jenkins.io` 
  - Change the `mirrorsync` cron script run every 30 min to sync data to OSUOSL instead of "from" OSUOSL before (as per https://github.com/jenkins-infra/helpdesk/issues/4653):
    - Add the `rsync.filter` file which used to be in the `pkg` VM
    - Update the rsync filter pattern to NOT include broken symlinks, neither the `updates/` and `update-center/` directories (first is a symlink to second). The resulting rsync filter only adds a bunch of `latest` symlinks so we should not expect more data.
  - Set up `mirrorsync` user SSH configuration to allow access to OSUOSL (as we used to retrieve data with `rsync` anonymous protocol): private key, config, known host
----

Testing process:

- Ran unit tests locally ✅ 
- Ran vagrant provisioning locally with both roles `pkg` and `archives `with success ✅ 
- From the local vagrant VMs, added manualyl the private keys (as their values are encrypted) and verified we could SSH to remote servers from them, with no error and no interactive prompts ✅:

```
# From 'pkg' vagrant VM
$ archives_jio_host="mirrorsync@archives.jenkins.io"
$ ssh -n "${archives_jio_host}" whoami 
mirrorsync

# From 'archives' vagrant VM
$ osuosl_host="jenkins@ftp-osl.osuosl.org"
$ ssh "${osuosl_host}" whoami
jenkins
```

----

Deployment process:

- Ran puppet agent once and then disable it on:
  - [x] pkg.origin.jenkins.io
  - [x] archives.jenkins.io
- Check expected deployment with a `noop` puppet agent run (`puppet agent --enable && puppet agent --test --noop && puppet agent --disable`). Take note of expected changes.
- Once PR merged: run the `noop` a second time and see if the changes are looking good (e.g. as expected) otherwise new PR is needed with the same process.
- Then deploy effectively on both machines and trigger a script on each: `sync.sh` on `pkg.origin` and `mirrorsync` on `archives`
- Check that trusted.ci update center works as expected


dduportal:
Tests are good, Puppet agents are disabled: let's roll!

dduportal:
Deployment result:

- No error to deploy
- On `pkg`:
  - Running the `sync.sh` script failed on 1st try with a script error (`/srv/releases/jenkins/jenkins` does not exist) - fixed by https://github.com/jenkins-infra/jenkins-infra/pull/4346
  - Running on 2nd try (with the fix) went further but failed due to permission issues with the `rsync` to archives. Changing the ownership of `/srv/releases` from `www-data` to `mirrorsync` user on archives.jenkins.io did the trick - fixed by https://github.com/jenkins-infra/jenkins-infra/pull/4346
  - 3rd run went well!
  - Also, update_center generation on trusted.ci.jenkins.io is back to green and runs well.
  - Need to carefully check a plugin update though (as next step)
- On `archives`:
  - Running the `/usr/bin/mirrorsync` script failed on 1st try with permission error to write in `/var/log/mirrorbits` even though it's owned by `mirrorsync` user. Had to change the user executing the script `/usr/bin/mirrorsync` from `www-data` to `mirrorsync` - fixed by https://github.com/jenkins-infra/jenkins-infra/pull/4346
  - 2nd try did work but showed a warning about missing known hosts entries for OSUSL: we only had 1 and requires 3 (fort each algorithm) - fixed by https://github.com/jenkins-infra/jenkins-infra/pull/4346

PR #4346, opened 2025-08-14
fix(pkg,archives) correct a few elements from #4344

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4653

This PR fixes up #4344 (see comment in https://github.com/jenkins-infra/jenkins-infra/pull/4344#issuecomment-3188261769 for details):

- archives: Ensure the user `mirrorsync` owns the data dir `/srv/release`
- archives: Ensure the user `mirrorsync` is used by the `/usr/bin/mirrorsync` sync script
- archives: Ensure ososul SSH host fingerprints are all there (there was only 1 while we need 3)
- pkg: Ensure the correct 'basedir' is used in the 'sync.sh' script

### DIFF
--- a/dist/profile/manifests/archives.pp
+++ b/dist/profile/manifests/archives.pp
@@ -148,8 +148,22 @@ Host ${$osuosl_mirroring['host']}
     ensure => present,
   }
 
-  # Create apache dirs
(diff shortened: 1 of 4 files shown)


dduportal:
Deployment result:

- Successful
- Both `sync.sh` and `mirrorsync` scripts are working without failures
- The file `TIME` on OSUOSL was updated with the latest timestamp

PR #4351, opened 2025-08-18
cleanup(pkg) remove all azcopy tasks

dduportal:
Following https://github.com/jenkins-infra/helpdesk/issues/4767#issuecomment-3191690128, we don't need to copy any data with `azcopy` anymore (from `pkg` VM to Azure storage).

This PR removes all `azcopy` tasks, along with the installation of `az` and `azcopy` and associated credentials


dduportal:
- Disabled Puppet agent on the pkg VM to be safe. Gotta deploy manually with a first `noop` step to inspect incoming changes.

PR #4353, opened 2025-08-18
feat(ci.jenkins.io) change EC2 cloud default setup regarding `idleTerminationTime` and `avoidUsingOrphanedNodes`

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4768#issuecomment-3196841090

This PR introduces 2 changes on the EC2 cloud agent templates (ci.jenkins.io only):

- Idle Termination time is now set to 1 minutes. The goal is to run the "EC2 GC" as often as possible

Capture d’écran 2025-08-18 à 15 49 36

- Enable the "Avoid Using Orphaned Nodes": we only use Ephemeral machines, let's use this (relatively) new feature as per @jglick advise:

Capture d’écran 2025-08-18 à 15 49 29

PR #4356, opened 2025-08-19
fix(ci.jenkins.io/profile::jenkinscontroller) stabilization: ensure EC2 clouds removes orphans nodes and fallback to ondemand instance when spot is out of capacity

dduportal:
This PR works around stabilizing the behavior of ci.jenkins.io regarding the EC2 cloud VM agents which are having many issues:

- Enable cleanup of "orphans" instances. As per https://github.com/jenkins-infra/jenkins-infra/pull/4353#discussion_r2282530140 (related to https://github.com/jenkins-infra/helpdesk/issues/4768), EC2 plugin can "garbage" collect instances not connected since 3 hours.
  - Cc @MarkEWaite : I guess it will avoid Pagerduty errors about disk almost full on EC2 instances which are not visible on ci.jenkins.io anymore (but still in EC2).
  - Should alleviate Spot capacity not available, which queues many jobs (for instance ATH jobs in https://github.com/jenkins-infra/helpdesk/issues/4771) because these "orphaned" EC2 instances are eating slots

- Also set up the "fallback to ondemand" option for spot instances, so we always get an instance. It should help in solving https://github.com/jenkins-infra/helpdesk/issues/4771 but we'll have to carefully check cloud credits.
  - Note: we should work on the BOM builds to decrease the amount of (underlying) VM being used.

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
@@ -4,6 +4,7 @@ jenkins:
   <%- @jcasc['cloud_agents']['ec2'].each do |ec2_cloud_name, ec2_cloud_setup| -%>
   - amazonEC2:
       name: <%= ec2_cloud_name %>
+      cleanUpOrphanedNodes: true
       credentialsId: "<%= ec2_cloud_setup['credentialsId'] %>"
     <%- $maxEC2Vms = 0; ec2_cloud_setup['agent_definitions'].each do |agent| -%>
       <%- $maxEC2Vms = agent['maxInstances'] + $maxEC2Vms -%>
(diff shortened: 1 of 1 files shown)


dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
```diff
@@ -4,6 +4,7 @@ jenkins:
   <%- @jcasc['cloud_agents']['ec2'].each do |ec2_cloud_name, ec2_cloud_setup| -%>
   - amazonEC2:
       name: <%= ec2_cloud_name %>
+      cleanUpOrphanedNodes: true
```
jglick, line 7: https://github.com/jenkinsci/ec2-plugin/pull/1115 FTR

PR #4376, opened 2025-09-02
feat(trusted.ci) adding csp plugin and jcasc configuration

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4776

installing CSP plugin along with it's config as code

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/csp-plugin.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/csp-plugin.yaml.erb
@@ -7,4 +7,4 @@ security:
     <%- if @jcasc['csp-plugin']['rule'] -%>
     rule: <%= @jcasc['csp-plugin']['rule'] %>
     <%- end -%>
-<% end %>
+<%- end -%>
--- a/dist/profile/templates/jenkinscontroller/casc/csp-plugin.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/csp-plugin.yaml.erb
@@ -1,0 +1,10 @@
+<%- if @jcasc['csp-plugin'] && @jcasc['csp-plugin']['reportOnly'] -%>
+security:
+  contentSecurityPolicyConfiguration:
+    <%- if @jcasc['csp-plugin']['reportOnly'] -%>
+    reportOnly: <%= @jcasc['csp-plugin']['reportOnly'] %>
+    <%- end -%>
+    <%- if @jcasc['csp-plugin']['rule'] -%>
+    rule: <%= @jcasc['csp-plugin']['rule'] %>
+    <%- end -%>
+<% end %>
--- a/dist/profile/templates/jenkinscontroller/casc/csp-plugin.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/csp-plugin.yaml.erb
@@ -1,4 +1,4 @@
-<%- if @jcasc['csp-plugin'] && @jcasc['csp-plugin']['reportOnly'] -%>
+<%- if @jcasc['csp-plugin'] -%>
 security:
   contentSecurityPolicyConfiguration:
     <%- if @jcasc['csp-plugin']['reportOnly'] -%>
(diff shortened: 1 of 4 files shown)


dist/profile/templates/jenkinscontroller/casc/csp-plugin.yaml.erb
```diff
@@ -0,0 +1,10 @@
+<%- if @jcasc['csp-plugin'] && @jcasc['csp-plugin']['reportOnly'] -%>
```
timja: your configuration is doing more than reportOnly so it shouldn't really be gated on that.

If you do want to keep it then the if block on L4-6 should be removed as this if block ensures its always set
```suggestion
<%- if @jcasc['csp-plugin'] -%>
```

smerle33: you are perfectly right ! thank you @timja !

hieradata/vagrant/common.yaml
```diff
@@ -54,6 +54,10 @@ profile::jenkinscontroller::jcasc:
         # but hierdata is behaving weirdly (bug in YAML parser?)
         defaultDisplayUrlProvider:
           providerId: "org.jenkinsci.plugins.displayurlapi.ClassicDisplayURLProvider"
+  csp-plugin:
+    reportOnly: true
+    rule: |-
+      "default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data: avatars.githubusercontent.com; media-src 'self' data: ; script-src 'self' 'report-sample' usage.jenkins.io;"
```
timja, line 60: I don't think `report-sample` is a thing, I think @daniel-beck made a typo in his comment.

Shouldn't it be this?
```suggestion
      "default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data: avatars.githubusercontent.com; media-src 'self' data: ; script-src 'self'; report-to usage.jenkins.io;"
```

To support firefox you may also want this:

```suggestion
      "default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data: avatars.githubusercontent.com; media-src 'self' data: ; script-src 'self'; report-to usage.jenkins.io; report-uri https://usage.jenkins.io"
```

timja (changes requested):
Have you configured the `Reporting-Endpoints` header to respond with `usage.jenkins.io=https://usage.jenkins.io` ?

This seems like Jenkins or CSP plugin should probably support setting this

hieradata/vagrant/common.yaml
timja: cc @daniel-beck

daniel-beck: https://github.com/jenkins-infra/helpdesk/issues/4776#issuecomment-3206262731 has the exact string that should be used.

The `report-uri` directive is dynamically generated, so while you are correct in isolation/in general, it doesn't apply to CSP plugin. (Also, `usage.jenkins.io` does not _accept_ reports, it _hosts a script_ for usage stats we allow with this rule. So it's an additional allowed source.)

> I don't think `report-sample` is a thing

It is, and gives a snippet of the script. See e.g. https://github.com/jenkinsci/stapler/pull/612 for an attempt at adding the source file path to it.

daniel-beck:
> Have you configured the `Reporting-Endpoints` header to respond with `usage.jenkins.io=https://usage.jenkins.io` ?

This seems to stem from a misintepretation of the rule. CSP plugin dynamically adds `report-uri` to enrich the report with information about the affected page and user. See https://github.com/jenkinsci/csp-plugin/blob/2c2349695a1800b3276b826bf84af17d6dd5e517/src/main/java/io/jenkins/plugins/csp/ContentSecurityPolicyFilter.java#L69-L70

hieradata/vagrant/common.yaml
timja: Ah I see it here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy#report-sample doesn't come up on google very well

timja (approved):

timja:
Thanks it was a misinterpretation, I've filed https://github.com/jenkinsci/csp-plugin/issues/57 as the CSP plugin is using the deprecated approach.

PR #4397, opened 2025-09-10
feat(ci.jio): temporarily switch agent used for ATH to non-spot

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4795

switch to non spot agent temporarily

### DIFF
--- a/hieradata/clients/aws.ci.jenkins.io.yaml
+++ b/hieradata/clients/aws.ci.jenkins.io.yaml
@@ -534,7 +534,8 @@ profile::jenkinscontroller::jcasc:
               - highram
               - docker-highmem
               - linux-amd64-big
-            spot: true
+            # we switch temporarily to non spot as per https://github.com/jenkins-infra/helpdesk/issues/4795
+            spot: false
             acpServerId: aws-internal
           - description: "Ubuntu 22.04 LTS x86_64 High Memory instance with JDK21 set as default java CLI"
             maxInstances: 50


hieradata/clients/aws.ci.jenkins.io.yaml
```diff
@@ -534,7 +534,8 @@ profile::jenkinscontroller::jcasc:
               - highram
               - docker-highmem
               - linux-amd64-big
-            spot: true
+            # we switch temporarily to non spot as per https://github.com/jenkins-infra/helpdesk/issues/4795
+            spot: false
```
timja, line 538: is it better to just change ATH to not request spot or do you want to fake it here so that others are not impacted?

timja (approved):

hieradata/clients/aws.ci.jenkins.io.yaml
dduportal: Faking it is easier as we still want to use Spot to decrease costs: It is a temporary measure until we are able to spread EC2 Spot requests over 3 AZs (and/or we change the instance sizer to avoid hitting capacity).

I would prefer to avoid changing pipelines (which disturbs developers by creating useless noise) back and forth.

Also, we plan to "factorize" the node allocation logic into a pipeline library function to provide the nodes allocations facilities to users: between the "use spot or not", "retry on agent failure", "retry with ondemand if using spot" and the label management, that would abstract this complexity from users.

dduportal (approved):

PR #4414, opened 2025-09-19
feat(usage): add the new VM usage.do.jenkins.io

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/1626

- change the way we set the usage of letsencrypt with a global switch
- detect if lvm plugin is needed or not
- add the new VM to the puppet master

testing done
impact limited to the new machine, but as a security measure, I will disable puppet agent on the current usage VM.

### DIFF
--- a/hieradata/vagrant/common.yaml
+++ b/hieradata/vagrant/common.yaml
@@ -9,6 +9,7 @@ lookup_options:
       strategy: deep
       merge_hash_arrays: true
 profile::jenkinscontroller::letsencrypt: false
+letsencrypt::enable: false # TODO use this setting on all profiles (see usage.yaml)
 # Per-host Datadog configuration
 datadog_agent::host: "vagrant.local"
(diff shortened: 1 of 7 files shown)


hieradata/vagrant/common.yaml
```diff
@@ -9,13 +9,22 @@ lookup_options:
...
 datadog_agent::agent_extra_options:
   bind_host: "0.0.0.0" # All hosts interfaces to allow container access
+# Fake self signed certificate
+ssl_legacy_cert: |
+  -----BEGIN CERTIFICATE-----
+  -----END CERTIFICATE-----
+# Fake self signed certificate
+ssl_legacy_key: |
+  -----BEGIN PRIVATE KEY-----
+  -----END PRIVATE KEY-----
```
smerle33:
```suggestion
```

dduportal (approved):

PR #4479, opened 2025-10-23
feat(jenkinscontroller/cert.ci.jio/ci.jio/trusted) Ensure that JDK tools always use the agent local JDKs instead of installing on each build

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4848 and following the initial work from https://github.com/jenkins-infra/helpdesk/issues/4831

This PR changes ALL the Jenkins JDK tools to use the locally installed JDK (`/opt/jdk-XX` on Linux and `C:/Tools/jdk-XX` on Windows) instead of downloading Adoptium JDK installer on each build.

The "exceptions" are covered like this:

- trusted.ci.jenkins.io: All JDKs from 8 up to 21 are already installed. JDK25 added in <insert PR here>.
- cert.ci.jenkins.io: duplicating the tool configuration with `jdk-XX` configuration so both namings are present. JenSec team pinged to see if they can and want updating their (private in `jenkinsci-cert`) pipelines without breaking their usages.
- ci.jenkins.io: `s390x`: JDK25 added as part github.com/jenkins-infra/helpdesk/issues/4831

----

Tests:

- Verified the updated template is working using local Vagrant:
  - Default hieradata for vagrant is the **same** as ci.jenkins.io to validate JDK tool setups are cumulative. Works as expected ✅ 
  - Manually tried to change the vagrant hieradata with trusted.ci updated config: we see the updatecenter/census labels added (e.g. overriding default config works)
  - Manually tried to change the vagrant hieradata with cert.ci updated config: we see the `jdk-XX` tool installer presents

----

Deployment:

- Once all is green, puppet agent will be disabled on each of the 3 controllers
- Then, for each controller, we'll deploy and verify it works as expected in the current order:
  - cert.ci.jenkins.io => deploy (e.g. puppet noop + visual check + apply if OK or correcting PR otherwise) => run the [`Infra-Team-Validate-Agents`](https://cert.ci.jenkins.io/view/Top-Level%20Items/job/Infra-Team-Validate-Agents/) job and verify that all tools used in this "validation" job are working as expected without downloading installers
  - trusted.ci.jenkins.io => deploy (e.g. puppet noop + visual check + apply if OK or correcting PR otherwise) => run the [`Infra-Team-Validate-Agents`](https://trusted.ci.jenkins.io/job/Infra-Team-Validate-Agents/) job and verify that all tools used in this "validation" job are working as expected without downloading installers => also run the updatecenter freestyle job and check it keeps working as expected
  - ci.jenkins.io => deploy (e.g. puppet noop + visual check + apply if OK or correcting PR otherwise) => run acceptance tests

----

Notes:

- cert.ci.jenkins.io will not have a "default fallback" JDK tool installer anymore. Either we have the JDK tool with a local install or we don't
- This PR is one step (of two) in the direction of removing the tedious template https://github.com/jenkins-infra/jenkins-infra/blob/production/dist/profile/templates/jdk-adoptium-url.erb (it was a useful but is a hard one to maintain). Need another PR on the "build-agent" profile to be able to remove it.
  - For the same reason, `updatecli` JDK elements are not removed yet.

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
@@ -37,38 +37,22 @@ tool:
   jdk:
     installations:
     <%- @jcasc['tools']['jdk'].each do |name, setup| -%>
-      <%- unless setup['disabled'] -%>
     - name: "<%= name %>"
       properties:
       - installSource:
           installers:
-        <%- setup['installers'].each do |installer_name, installer_setup| -%>
-          - <%= installer_setup['type'] %>:
-          <%- if installer_setup['label'] -%>
-              label: "<%= installer_setup['label'] %>"
-          <%- end -%>
-          <%- if installer_setup['type'] == 'zip' -%>
-            <%- $jdk_major_version = name.gsub('jdk-','').gsub('jdk_','').gsub('jdk','') # Remove eventual prefix
-            $jdk = {
-              'name' => name,
-              'major_version' => $jdk_major_version,
-              'version' => installer_setup['version'] ? installer_setup['version'] : @jcasc['tools_default_versions']["jdk" + $jdk_major_version],
-              'cpu_arch' => installer_setup['cpu_arch'] ? installer_setup['cpu_arch'] : 'x64',
-              'os' => installer_setup['os'] ? installer_setup['os'] : 'linux',
-            }
-            -%>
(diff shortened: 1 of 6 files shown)


dduportal:
Deployed with success on cert.ci.jenkins.io

dduportal:
Deployed with success on trusted.ci.jenkins.io

dduportal:
Deployed with success on ci.jenkins.io

PR #4488, opened 2025-10-28
feat(jenkins::agent/agent.trusted.ci.jenkins.io) install JDKs from URLs in hieradata and verify cryptographic signature of downloads

dduportal:
After the work in https://github.com/jenkins-infra/helpdesk/issues/4831 (and also PRs #4476 #4479 #4480), we can continue cleaning up the old "JDK setup" code in this repository (Puppet):

- cleanup: The file `dist/profile/templates/jdk-adoptium-url.erb` is now removed as part of this PR
- feat: the JDKs installation method in the permanent agents (currently only `agent.trusted.ci.jenkins.io` but we might need another agent soon - ref. https://github.com/jenkins-infra/helpdesk/issues/4695 and https://github.com/jenkins-infra/helpdesk/issues/4666) is changed:
  - Hieradata provides both the expected version and the installer direct URLs. We can track all of them with the same "all in one" updatecli manifest as we currently have in packer-images, which is way easier to maintain: it relies on the Adoptium API and can be opt-off by a manual update if need be.
  - The installation now verify the cryptographic signature and checksums of downloads. It improves our security posture.

----

Notes:

- Tested locally with the vagrant VM (`vagrant up --provision jenkins::agent`) which installed the 4 JDKs as expected.
  - Verified it is idempotent (running the provisioning a second time does NOT re-install the JDKs)
  - Also tried a JDK upgrade and downgrade (JDK17 has a new version available, and tried a downgrade of JDK8 to previous version) with success: the `/opt/jdk-XX` directories where updated for JDK17 and JDK8
- Updatecli version tracking will be treated in subsequent PR: https://github.com/jenkins-infra/jenkins-infra/pull/4489
- Deployment will be performed manually on agent.trusted.ci.jenkins.io (with the usual "noop" technique)


dduportal:
- Noop looked good: 

```
puppet agent --enable && puppet agent --test --noop && puppet agent --disable "dduportal is doing some manual maintenance"
Info: Using environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Applying configuration version '1761651277'
Notice: /Stage[main]/Profile::Buildagent/Exec[Ensure Adoptium GPG key is present]/returns: current_value 'notrun', should be ['0'] (noop)
Notice: Class[Profile::Buildagent]: Would have triggered 'refresh' from 1 event
Notice: Stage[main]: Would have triggered 'refresh' from 1 event
Notice: Applied catalog in 8.56 seconds
```

=> applied with success!

PR #4489, opened 2025-10-28
chore(updatecli) track JDK versions for Jenkins permanent agents

dduportal:
Requires https://github.com/jenkins-infra/jenkins-infra/pull/4488 to be merged (+ source branch of this PR rebased)

This PR changes the way `updatecli` tracks the JDK versions installed in our permanent Jenkins agents managed by Puppet. It reuses the same technique as in our jenkins-infra/packer-images (e.g. specify the list of JDKs to track in `updatecli/values.yaml` - duplicated from the target, can be improved -) and an all in one manifest.

----

Notes:

- Add a fixup (rspec unit tests) for #4490 
- Remove the (now unused) shell script `updatecli/scripts/check-jdk.sh`
- Tested with JDK17 (and the old version removed in #4490 ) and it proposes an update (new JDK17 version has both x86_64 and arm64 installers available)
- We expect the checks to detect a new JDK21 version but the condition will fail as not all platform installers are available yet


dduportal:
The logs shows that a new JDK21 version has been found but not all installers are available so no expected PR. This is the expected behavior.

PR #4560, opened 2025-12-02
feat(aws.ci.jenkins.io): custom `infratest-windows` agent template with Windows Server Core 2025 and no local NVMe mount

lemeurherve:
In https://github.com/jenkinsci/docker-ssh-agent/pull/581 I successfully built a Windows Server Core 2022 image from a 2025 host by removing docker daemon config setting `data-root` to `Z:\docker` in order to avoid hitting the issue described in https://github.com/moby/moby/issues/48093.

This PR udpates `infratest-windows` ec2 agent template on ci.jenkins.io to use a 2025 image and to disable local NVMe.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4791
- https://github.com/jenkins-infra/jenkins-infra/blob/71eff0ea69c667180f748e16c1bd3f42c099edae/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb#L20-L130

PR #4562, opened 2025-12-03
feat(aws.ci.jenkins.io): add `customDataDiskNotUsedForDocker` ec2 agent template parameter

lemeurherve:
This PR adds a `customDataDiskNotUsedForDocker` ec2 agent template parameter allowing to not create any docker daemon config file using local NVMe as data-root if a custom disk is defined in a ci.jenkins.io ec2 template, to avoid hitting https://github.com/moby/moby/issues/48093 when building Windows images of different versions from a same host version.

If this parameter is undefined, a docker daemon config using custom disk is still created by default in the cloud-init script of ec2 VM, to avoid breaking existing agent templates.

### Testing done
`vagrant up jenkins::controller --provision`

Then in http://localhost:8080/manage/cloud/aws-us-east-2/configure checked the different cloud-init content (in "Advanced" details) for the following ec2 templates:
- "Spot Windows 2019 x86_64 with JDK21 with custom disk"
  - `customDataDisk: 'Z:'` set in its vagrant definition
  - Has a local NVMe as custom disk and a docker daemon config using this custom disk
  - <details>

    image
    </details>
- "Windows 2022 x86_64 with default (current) JDK without custom disk"
  - No `customDataDisk: 'Z:'` set in its vagrant definition
  - Does not have a local NVMe as custom disk nor a docker daemon config using this custom disk
  - <details>

    image
    </details>
- "Windows 2025 x86_64 with default (current) JDK with custom disk but not using it for docker"
  - `customDataDisk: 'Z:'` set in its vagrant definition
  - `customDataDiskNotUsedForDocker: true` set in its vagrant definition
  - Has a local NVMe as custom disk but no docker daemon config using this custom disk
  - <details>

    image
    </details>

Refs:
- https://github.com/jenkins-infra/jenkins-infra/pull/4560#pullrequestreview-3531102268
- https://github.com/jenkins-infra/helpdesk/issues/4791

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
@@ -171,8 +171,6 @@ jenkins:
                 # Restart docker engine
                 Restart-Service docker
                 <%- end -%>
-                <%- else -%>
-                ## No NVMe(s) as custom disk(s)
                 <%- end -%>
 
                 <%- if $remoteAdmin != 'Administrator' -%>
--- a/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
@@ -22,6 +22,7 @@ jenkins:
       <%- $cloudInitDoneParentDir = "#{$tempDir}" -%>
       <%- $cloudInitDoneFile = "#{$cloudInitDoneParentDir}/.cloud-init.done" -%>
       <%- $javaHome = agent['javaHome'] ? agent['javaHome'] : @jcasc['agents_setup'][$os]['javaHome'] -%>
+      <%- $noCustomDataDiskForDocker = agent['noCustomDataDiskForDocker'] == nil ? false : agent['noCustomDataDiskForDocker'] -%>
       - ami: "<%= agent['ami'] ? agent['ami'] : @jcasc['agent_images']['ec2_amis'][$os + "-" + agent['os_version'].to_s + "-" + agent['architecture'].to_s] %>"
         amiType:
       <%- if $os == 'windows' -%>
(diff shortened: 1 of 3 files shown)


dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
```diff
@@ -22,6 +22,7 @@ jenkins:
       <%- $cloudInitDoneParentDir = "#{$tempDir}" -%>
       <%- $cloudInitDoneFile = "#{$cloudInitDoneParentDir}/.cloud-init.done" -%>
       <%- $javaHome = agent['javaHome'] ? agent['javaHome'] : @jcasc['agents_setup'][$os]['javaHome'] -%>
+      <%- $customDataDiskNotUsedForDocker = agent['customDataDiskNotUsedForDocker'] == nil ? false : agent['customDataDiskNotUsedForDocker'] -%>
```
dduportal: In Ruby, you can test for both presence AND `true` value with `agent['customDataDiskNotUsedForDocker']`, like Groovy.

Given the logic order you used here, I suggest to use the same ternary order as the line above for `javaHome`



```suggestion
      <%- $customDataDiskNotUsedForDocker = agent['customDataDiskNotUsedForDocker'] ?  : false -%>
```

dduportal: Even better (see comment below): the variable is unneeded as we only use it once.


```suggestion
```

dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
```diff
@@ -156,6 +157,9 @@ jenkins:
                   }
                 }
 
+                <%- if $customDataDiskNotUsedForDocker -%>
```
dduportal:
```suggestion
                <%- if agent['customDataDiskNotUsedForDocker'] -%>
```

Using ruby + puppet "checks": either the key exists and has the value `true` or it does not

dduportal (approved):
LGTM

dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
lemeurherve: My condition was more complex in the beginning, hence this var.

As it's simpler now, your suggestion make sense, PR updated.

PR #4596, opened 2026-01-05
feat(trusted.ci.jenkins.io): add `arm64docker` agent template

lemeurherve:
This PR adds an arm64 agent template (based on the newer ubuntu one), ~~and adds an `amd64` label to the default `docker` one for parity with the corresponding ci.jenkins.io agent template (not sure about that change, I haven't found the corresponding ci.jenkins.io agent template label 🤔~~ see https://github.com/jenkins-infra/jenkins-infra/pull/4596#discussion_r2662409010).

Note: I've reused the same `arm64docker` label as the AWS ci.jenkins.io agent template one so we can use the same labels in both ci.jenkins.io & trusted.ci.jenkins.io in docker image pipeline:
https://github.com/jenkins-infra/jenkins-infra/blob/f67283f8dc6dac49954ba5407796bce2e1562eb8/hieradata/clients/aws.ci.jenkins.io.yaml#L409-L420

https://github.com/jenkinsci/docker/blob/52d8ee8b73d046decfcbf97cd16582e187c86fa2/Jenkinsfile#L18-L20

Refs:
- https://github.com/jenkins-infra/helpdesk/issues/4934
- https://github.com/jenkinsci/docker/issues/2154#issuecomment-3709445653

PR #4599, opened 2026-01-06
fix(trusted.ci.jenkins.io) use correct instance size for arm64 Azure VM template

dduportal:
Fixup of a754f695 (#4596)

Ref. https://github.com/jenkins-infra/helpdesk/issues/4934#issuecomment-3715666716 and https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dpdsv5-series?tabs=sizebasic


dduportal:
Tested manually with success: let's roll

PR #4603, opened 2026-01-09
breaking(controller.cert.ci): switch `maven-<jdk>-windows`, `docker-windows` and `maven-windows` labels to Windows 2025 agent templates

lemeurherve:
This change makes those default `maven-<jdk>-windows` labels returning a Windows Server Core 2025 agent template on cert.ci.jenkins.io.
- maven-8-windows
- maven-11-windows
- maven-17-windows
- maven-21-windows
- maven-25-windows

It also take care of:
- docker-windows
- windows (still used?)
- maven-windows (deleted)

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4955


dduportal (changes requested):
In its current state, this PR introduces regression on ci.jenkins.io related to the JDK17 being used instead of JDK21.

This would be a bad surprise for end users.


Also: some comments added are stating a wrong assertions, might need clarification

lemeurherve:
> In its current state, this PR introduces regression on ci.jenkins.io related to the JDK17 being used instead of JDK21.
>
> This would be a bad surprise for end users.

Initially put there to align "default template" labels ("contracts") between ci.jenkins.io and trusted.ci.jenkins.io to ensure jobs like Jenkins controller and agents images use equivalent agent templates, to be done elsewhere in another time, reverted.

> Also: some comments added are stating a wrong assertions, might need clarification

Extracted to https://github.com/jenkins-infra/jenkins-infra/pull/4621 opened in draft as support for discussion (not related to https://github.com/jenkins-infra/helpdesk/issues/4955).

lemeurherve:
Extracted ci.jenkins.io changes to #4622 in https://github.com/jenkins-infra/jenkins-infra/pull/4603/commits/090fa265f4633c43e995fe596b9e46493c2d5b06 as ci.jenkins.io default template labels should be changed in a distinct PR.

dduportal:
> Initially put there to align "default template" labels ("contracts") between ci.jenkins.io and trusted.ci.jenkins.io to ensure jobs like Jenkins controller and agents images use equivalent agent templates, to be done elsewhere in another time, reverted.

Can you be more precise so we can delimit the problem you are thinking about? 

I'm asking because most of the used labels are not "default": `maven-XX` or `maven-XX-windows` are most of our usages (>85%) in ci.jenkins.io and are the priority. Add to that the "tool" cases (which select the JDK and Maven from Jenkins tool): the leftover are rare so we should be able to list them exhaustively before trying a solution.

Regarding the "alignement", it looks like you are thinking around the "Docker" cases (e.g. the images): are they using a JDK for builds?

lemeurherve:
I've removed trusted.ci changes and took care of only requested `maven-<jdk>-windows` labels for cert.ci cf finalised https://github.com/jenkins-infra/helpdesk/issues/4955:
> - [ ] On cert.ci.jenkins.io, `maven-<jdk>-windows`
>
> Note: on trusted.ci.jenkins.io jobs only use specific windows-<version> labels, nothing to do.

dduportal (approved):

PR #4625, opened 2026-01-18
fix(aws.ci.jenkins.io/`profile::jenkinscontroller`) fix EC2 Windows agent initialization (`jenkins` user init. /CryptoAPI)

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4939 and https://github.com/jenkinsci/credentials-plugin/pull/999

This PR is a first step which fixes the Windows EC2 agents with:

- Define `jenkins` user as an explicit local user 
- Correct user and user profile initialization to avoid spreading them across C: and Z:
- Prepare `jenkins` user to be a non admin (SSH group and Docker daemon setup) closing the gap with Linux and as a requirement for the JVM Crypto API setup
- Remove unused agent template

Note: this change is not enough and the crypto API setup with a custom `exe` will need to be optimized in the future.

PR #4639, opened 2026-01-26
feat(aws.ci.jenkins.io): set `jenkins` user for EC2 agents as non administrator and with a system seed generator enabled without external resources

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4939#issuecomment-3796250223 and https://github.com/jenkinsci/credentials-plugin/pull/999#issuecomment-3795380493

This PR ensures that the `jenkins` user of the EC2 Windows Agents is:

- feat(jenkinscontroller/ec2): Only created **after** setting up the Windows registry to use the external disk (if any)
- feat(jenkinscontroller/ec2): Not an administrator anymore to ensure the system seed generator used by the JVM CryptoAPI can be used without permission errors (Administrators have a different CryptoAPI container)
  - Ref. https://github.com/jenkinsci/credentials-plugin/pull/999
  - Note that #4625 set up the prerequisites in the EC2Launchv2 code ("cloud init" or "user data provided init") and jenkins-infra/pipeline-library#983 has already been merged to avoid `buildPlugin()` errors
- fix(jenkinscontroller/ec2): The system seed generator is not initialized with a custom `exe` debugging binary: instead it relies on the JVM with the `jshell` command
  - Ref. https://github.com/jenkinsci/credentials-plugin/pull/999#issuecomment-3767662011 courtesy of @jtnord 
- fix(aws.ci.jenkins.io): use `jenkins` user on all EC2 Windows templates (instead of `Administrator` for `infratest-windows`
- fix(jenkinscontroller/ec2): Copy SSH public key from existing `Administrator` (we don't care about this not being the safest, same as the clear password: user has access to Docker so they can do anything. These agents are ephemeral so we don't really care) instead of download it. One less network call during the init. phase.
- chore(jenkinscontroller/ec2) work around comments:
  - remove unused comment which is confusing and wrong
  - Added comment explaining the order of user initialization tasks

---

Tested with:
- Local unit tests (puppet only)
- Local vagrant (`vagrant up --provision jenkins::controller`)  and checked the user data for the `maven-21-windows` template which has the set of parameters we're interested in => code looked valid
- On ci.jenkins.io with https://ci.jenkins.io/job/Plugins/job/credentials-plugin/job/PR-999/29
  - Copied the code from local vagrant, changed the datadog API key from the "dummy key" to the production key and inserted the user data on the `infratest-windows` template
  - Also set up the default agent user of the `infratest-windows` template to `jenkins`

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds-ec2.yaml.erb
@@ -113,7 +113,6 @@ jenkins:
         # Note: Escape '$' in JCasc with '^' - ref. https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/features/secrets.adoc#passing-secrets-through-variables
         userData: |-
       <%- if $os == 'windows' -%>
-          # Requires using YAML for the Windows "Cloud Init" stuff. Multipart upload of a powershell script does not work.
           version: 1.1
           tasks:
           - task: executeScript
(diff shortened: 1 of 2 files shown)


dduportal:
Tested, let's roll and I'll verify we've unblocked https://github.com/jenkinsci/credentials-plugin/pull/999 and https://github.com/jenkins-infra/helpdesk/issues/4939 first, while checking the jenkinsci/docker `master` branch as well (to check 2019 and 2022 init).

Issue #4694, opened 2026-02-14
The openvpn updatecli manifest opens and closes a lot of PRs


lemeurherve:
> <img width="3122" height="4190" alt="Image" src="https://github.com/user-attachments/assets/81d84202-2259-4dbe-92d2-333daf58ffc7" />

It's also adding new commits each time!

On the last one it arrived at 194 commits: https://github.com/jenkins-infra/jenkins-infra/pull/4693/commits

jayfranco999:
The repeated PRs and growing commit history were caused by `common.yaml` being rewritten in a way that reintroduced a stale `profile::openvpn::image_tag`. This made Updatecli think there was always something to “fix,” resulting in version flips and noisy PR churn (https://github.com/jenkins-infra/jenkins-infra/pull/4693/commits) 

This resolved this by:

Removing the separate image_tag target to avoid conflicting writes (https://github.com/jenkins-infra/jenkins-infra/pull/4725)

Updating the route generation script to set profile::openvpn::image_tag in the same yq operation that updates the VPN routes (https://github.com/jenkins-infra/jenkins-infra/pull/4726)

This makes the update atomic and deterministic, and stops the version from flipping back. PR churn is now resolved.

`updatecli/scripts/update-vpn-routes.sh` tested locally with success, the condition in `updatecli/updatecli.d/openvpn.yaml`  fails testing locally since the dockerhub acr registry is ip restricted. The following command was used to test the yq operation that updates the VPN routes:

```
bash ./updatecli/scripts/update-vpn-routes.sh 3.7.16 servers | grep openvpn::image_tag

profile::openvpn::image_tag: 3.7.16
```

Tested with success on infra.ci (https://infra.ci.jenkins.io/job/updatecli/job/jenkins-infra/job/PR-4726/1/stages/log?nodeId=24):

```
[2026-02-19T12:18:37.562Z] ⚠ Bump OpenVPN version and configuration:
[2026-02-19T12:18:37.562Z] 	Source:
[2026-02-19T12:18:37.562Z] 		✔ [getUpdatedHieradataCommon] 
[2026-02-19T12:18:37.562Z] 		✔ [getUpdatedHieradataOpenVPN] 
[2026-02-19T12:18:37.562Z] 		✔ [latestVersion] Get the latest stable docker-openvpn version
[2026-02-19T12:18:37.562Z] 	Condition:
[2026-02-19T12:18:37.562Z] 		✔ [checkIfKeyExist] 
[...]

PR #4696, opened 2026-02-14
chore(updatecli): rename openvpn.yaml to openvpn.yaml.KO to temporary disable it

lemeurherve:
To avoid receiving notifications all the weekend.

Ref:
- #4694


dduportal (approved):
Don’t forget to update vpn routes manually then

PR #4775, opened 2026-03-30
feat: add Datadog tags on ec2 & Azure VMs to collect their controller FQDN and agent type

lemeurherve:
This change adds Datadog tags to VMs spawned by Jenkins controllers so we can filter them out in https://github.com/jenkins-infra/datadog/blob/1f4e8f1da9cc5c0b83b118548c388457d35061de/datadog-monitors.tf#L22-L27 to avoid PagerDuty when an ephemeral agent disk is almost full (and which we don't care about as it's expected for some builds to almost fill their agent disk)

Refs:
- https://docs.datadoghq.com/getting_started/tagging/assigning_tags/?tab=noncontainerizedenvironments
- https://github.com/jenkins-infra/helpdesk/issues/5056

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/clouds-azurevm.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/clouds-azurevm.yaml.erb
@@ -9,6 +9,11 @@ jenkins:
       existingResourceGroupName: "<%= cloudsetup['resourceGroup'] %>"
       resourceGroupReferenceType: "existing"
       maxVirtualMachinesLimit: <%= cloudsetup['maxInstances'] %>
+      nodeProperties:
+        - envVars:
+            env:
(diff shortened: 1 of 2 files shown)


dduportal (changes requested):
The specified environment variable will never be visible by the datadog agent.
It's a JCasC setup which tells controllers what custom environment variable to add in the `agent.jar` process, run by the `jenkins` user while datadog is run by the `datadog` user.

You want to set up datadog tags during the cloud init process, which is already setting up the API key of the agent.

lemeurherve:
Setting tags via datadog.yaml file, merging.

lemeurherve:
All good 🎉 

> image

lemeurherve:
Caused:
- https://github.com/jenkins-infra/helpdesk/issues/5055#issuecomment-4163108982

PR #4784, opened 2026-04-08
feat(datadog): add `jenkins_agent_description` tag

lemeurherve:
Storing `description` as it's available in all ec2 and azure agent template definitions (while `name` is only defined for the azure ones).

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5056


lemeurherve:
> image

PR #4825, opened 2026-05-01
feat(trusted.ci) add the new controller-sponsored VM

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5084#issuecomment-4343922083

Hieradata is a copy of current `controller.trusted.ci.jenkins.io` with the following changes:

- `agent-2` commented out
- hostnames adapted

UUID already set up as well in the `site.pp`

### DIFF
--- a/hieradata/clients/controller-sponsored.trusted.ci.jenkins.io.yaml
+++ b/hieradata/clients/controller-sponsored.trusted.ci.jenkins.io.yaml
@@ -1,0 +1,419 @@
+---
+lookup_options:
+  "^profile::jenkinscontroller::jcasc$":
+    merge:
+      strategy: deep
+      merge_hash_arrays: true
+  "^letsencrypt":
+    merge:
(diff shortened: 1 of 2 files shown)


lemeurherve:
Too bad we're loosing the git history from this new file, then difficult to do otherwise while two controllers temporarily existed at the same time.

dduportal:
> Too bad we're loosing the git history from this new file.

Not sure to agree. It is a new VM , with a new context. So it makes sense to not to carry over the context of the old vm. If something need to be searched through the history, it is crucial that the researcher get the context through the change which led to the new file. Otherwise it will only lead to context-less results from old vm mostly not adapted for the new vm.

do you have a particular use case in mind ?

lemeurherve:
I (initially) wanted to check if agent template labels changed as docker-ssh-agent release builds are stuck without any "linux" agents available.

PR #4835, opened 2026-05-05
feat(trusted.ci): use Maven from agents instead of installing it

lemeurherve:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5090

#### Testing done
```bash
vagrant up --provision jenkins::controller
vagrant ssh jenkins::controller
vagrant@localhost:~$ sudo su -
root@localhost:~$ cat /var/lib/jenkins/casc.d/tools.yaml
```
```yaml
tool:
  maven:
    installations:
    - name: "mvn"
      properties:
      - installSource:
          installers:
          - maven:
              id: "3.9.15"
    - name: "mvnCustomVersion"
      properties:
      - installSource:
          installers:
          - maven:
              id: "3.5.4"
    - name: "mvnCustomHome"
      home: "/usr/share/apache-maven-3.8.8"
    - name: "mvnCustomInstaller"
      properties:
      - installSource:
          installers:
          - command:
              command: "echo /usr/share/apache-maven-3.9.15"
              label: "maven-17 || maven-21 || maven-25"
              toolHome: "/usr/share/apache-maven-3.9.15"
          - batchFile:
              command: "echo C:/tools/apache-maven-3.9.15/bin"
              label: "maven-17-windows || maven-21-windows || maven-25-windows"
              toolHome: "C:/tools/apache-maven-3.9.15/bin"
          - command:
              command: "echo /usr/share/apache-maven-3.9.15"
              label: "updatecenter || census"
              toolHome: "/usr/share/apache-maven-3.9.15"
    - name: "mvnCustomInstallerAndVersion"
      properties:
      - installSource:
          installers:
          - command:
              command: "echo /usr/share/apache-maven-4.0.0"
              label: "maven4"
              toolHome: "/usr/share/apache-maven-4.0.0"
```
<details><summary>UI screenshot</summary>

image

</details>

### DIFF
--- a/dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
@@ -68,15 +68,15 @@ tool:
           installers:
           <%- setup['installers'].each do |installer_name, installer_label| -%>
             <%- if installer_name =~ /windows/ -%>
-          - command:
-              command: "echo /usr/share/apache-maven-<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>"
-              label: "<%= installer_label %>"
-              toolHome: "/usr/share/apache-maven-<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>"
-            <%- else -%>
           - batchFile:
               command: "echo C:/tools/apache-maven-<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>/bin"
               label: "<%= installer_label %>"
               toolHome: "C:/tools/apache-maven-<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>/bin"
+            <%- else -%>
+          - command:
+              command: "echo /usr/share/apache-maven-<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>"
+              label: "<%= installer_label %>"
+              toolHome: "/usr/share/apache-maven-<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>"
             <%- end -%>
           <%- end -%>
         <%- elsif setup['version'] || setup['default_version'] -%>
--- a/dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
+++ b/dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
@@ -70,6 +70,18 @@ tool:
               id: "<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>"
         <%- elsif setup['home'] -%>
       home: "<%= setup['home'] %>"
+        <%- elsif setup['useMavenFromAgentLabels'] -%>
+      properties:
+      - installSource:
+          installers:
+          - command:
+              command: "echo /usr/bin"
+              label: "<%= setup['useMavenFromAgentLabels'] %>"
+              toolHome: "/usr/bin"
+          - batchFile:
+              command: "echo C:\tools\apache-maven-<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>\bin"
+              label: "<%= setup['useMavenFromAgentLabels'] %>s"
+              toolHome: "C:\tools\apache-maven-<%= setup['version'] ? setup['version'] : @jcasc['tools_default_versions']['maven'] %>\bin"
         <%- end -%>
       <%- end -%>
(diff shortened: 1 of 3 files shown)


hieradata/clients/controller-sponsored.trusted.ci.jenkins.io.yaml
```diff
@@ -66,6 +66,9 @@ profile::jenkinscontroller::jcasc:
       jdk25:
         installers:
           linux-amd64: "(linux && amd64) || linux-amd64 || jdk8 || jdk-8 || maven-8 || updatecenter || census"
+    maven:
+      maven:
```
dduportal:
```suggestion
      mvn:
```

The current Maven installer is named `mvn`:

```
tool:
  maven:
    installations:
    - name: "mvn"
      properties:
      - installSource:
          installers:
          - maven:
              id: "3.9.15"
```

Image

dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
```diff
@@ -70,6 +70,18 @@ tool:
...
         <%- elsif setup['home'] -%>
       home: "<%= setup['home'] %>"
+        <%- elsif setup['useMavenFromAgentLabels'] -%>
+      properties:
+      - installSource:
+          installers:
+          - command:
+              command: "echo /usr/bin"
+              label: "<%= setup['useMavenFromAgentLabels'] %>"
+              toolHome: "/usr/bin"
```
dduportal: It's really important to test this before deploying:
- The tool installation is `/usr/share/apache-maven-<maven version> as per https://github.com/jenkins-infra/packer-images/blob/90aef898600f606dd63f26435724476442d8a334/provisioning/ubuntu-provision.sh#L379
- But the binary `mvn` ins symlinked in `/usr/bin`: https://github.com/jenkins-infra/packer-images/blob/90aef898600f606dd63f26435724476442d8a334/provisioning/ubuntu-provision.sh#L380

I'm not sure which is the expected value by the tool installer

lemeurherve: > I'm not sure which is the expected value by the tool installer

I don't either

hieradata/clients/controller-sponsored.trusted.ci.jenkins.io.yaml
```diff
@@ -66,6 +66,9 @@ profile::jenkinscontroller::jcasc:
...
+        useMavenFromAgentLabels: "maven-17 || maven-21 || maven-25 || maven-17-windows || maven-21-windows || maven-25-windows"
```
dduportal: Note: this setup does NOT cover the `agent-2`. It does not seem critical since I'm not aware of any job in trusted.ci which would need the `mvn` tool installer.

However, the current setup is manual but works if needed on the `agent-2`, while this new "as code" setup would not: are you ok to add a comment so at least we know (the day a crisis will force us to use agent-2 for crawler or RPU for instance)?

dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
dduportal: WDYT if we change, in packer-images, the installation path to `/usr/share/apache-maven-3` ?

- It would be the same pattern as the JDK tool above (static path to avoid updates on each patch update) 
- With only the major version to allow us support both Maven 3 and 4 with tools

hieradata/clients/controller-sponsored.trusted.ci.jenkins.io.yaml
lemeurherve: Addressed in https://github.com/jenkins-infra/jenkins-infra/pull/4835/changes/5a13c8ca2f9f18cb8b88962210ff3d4a6df69516

dist/profile/templates/jenkinscontroller/casc/tools.yaml.erb
lemeurherve: In the end I don't think it's necessary as the path is not static anymore and as we can specify a different version than the default ones on these "installers" as shown with the additional test case introduced in https://github.com/jenkins-infra/jenkins-infra/pull/4835/commits/d9677fafd5c1ca3a35e9e82ab45d63016ba0c85c

lemeurherve: > > I'm not sure which is the expected value by the tool installer
> 
> I don't either

Addressed by using the full path like for cert.ci.jenkins.io in https://github.com/jenkins-infra/jenkins-infra/pull/4835/changes/1e6447e2bf4dba299c02614dac2ffcfebacea40c

dduportal:
Closing

PR #4842, opened 2026-05-07
feat(buildagent): install `mvn` if defined in `$tools_versions`

lemeurherve:
This change ensures the default Maven version defined in `profile::jenkinscontroller::jcasc::tools_default_versions` is installed on `jenkins::agent` profile (ex: agent-2.trusted.ci.jenkins.io).

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5090#issuecomment-4398089621

### Testing done

```bash
$ vagrant up --provision jenkins::agent
$ vagrant ssh jenkins::agent
>> no serverspec defined for privateci
Last login: Thu May  7 18:40:06 2026 from 172.17.0.1
vagrant@db28cfe37370:~$ mvn --version
Apache Maven 3.9.15 (98b2cdbfdb5f1ac8781f537ea9acccaed7922349)
Maven home: /usr/share/apache-maven-3
Java version: 21.0.11, vendor: Eclipse Adoptium, runtime: /opt/jdk-21
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "6.12.76-linuxkit", arch: "aarch64", family: "unix"
vagrant@db28cfe37370:~$ echo $PATH
/usr/share/apache-maven-3/bin:/opt/jdk-21/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/opt/puppetlabs/bin
```

### DIFF
--- a/dist/profile/manifests/buildagent.pp
+++ b/dist/profile/manifests/buildagent.pp
@@ -153,6 +153,110 @@ class profile::buildagent (
         unless  => "/usr/bin/test -f ${java_bin} && ${java_bin} -version 2>&1 | /bin/grep --quiet '${jdk_version_to_check}'",
       }
     }
+
+    $maven_version = lookup('profile::jenkinscontroller::jcasc.tools_default_versions.maven', { default_value => undef })
+
+    if $maven_version {
+
+      $maven_major   = regsubst($maven_version, '^([0-9]+)\..*$', '\1')
+      $maven_archive = "apache-maven-${maven_version}-bin.tar.gz"
+
+      $base_url = "https://archive.apache.org/dist/maven/maven-${maven_major}/${maven_version}/binaries"
+
+      $maven_url       = "${base_url}/${maven_archive}"
+      $maven_checksum  = "${maven_url}.sha512"
+      $maven_signature = "${maven_url}.asc"
+
+      $maven_dir = "/usr/share/apache-maven-${maven_major}"
+      $maven_bin = "${maven_dir}/bin/mvn"
+
+      $temp_archive  = "/tmp/${maven_archive}"
+      $temp_checksum = "${temp_archive}.sha512"
+      $temp_sig      = "${temp_archive}.asc"
+      $maven_keys_file = '/tmp/maven-keys.gpg'
+
+      exec { 'Download Apache Maven key':
+        command => "/usr/bin/curl --silent --show-error --location https://downloads.apache.org/maven/KEYS --output ${maven_keys_file}",
+        creates => $maven_keys_file,
+        require => Package['curl'],
+        unless  => "/usr/bin/gpg --list-keys | /bin/grep 'Apache Maven'",
+      }
+
+      exec { 'Import Apache Maven key':
+        command => "/usr/bin/gpg --import ${maven_keys_file}",
+        require => [
+          Package['gpg'],
+          Exec['Download Apache Maven key'],
+        ],
+        unless  => "/usr/bin/gpg --list-keys | /bin/grep 'Apache Maven'",
+      }
+
+      exec { "Download Maven ${maven_version}":
+        command => "/usr/bin/curl --silent --show-error --location ${maven_url} --output ${temp_archive}",
+        require => Package['curl'],
+        unless  => "/usr/bin/test -f ${maven_bin} && ${maven_bin} --version | /bin/grep --quiet '${maven_version}'",
+      }
+
+      exec { "Download Maven ${maven_version} checksum":
+        command => "/usr/bin/curl --silent --show-error --location ${maven_checksum} --output ${temp_checksum}",
+        require => Package['curl'],
+        unless  => "/usr/bin/test -f ${maven_bin} && ${maven_bin} --version | /bin/grep --quiet '${maven_version}'",
+      }
+
+      exec { "Download Maven ${maven_version} signature":
+        command => "/usr/bin/curl --silent --show-error --location ${maven_signature} --output ${temp_sig}",
+        require => Package['curl'],
+        unless  => "/usr/bin/test -f ${maven_bin} && ${maven_bin} --version | /bin/grep --quiet '${maven_version}'",
+      }
+
+      exec { "Verify Maven ${maven_version} signature":
+        command => "/usr/bin/gpg --verify ${temp_sig}",
+        require => [
+          Exec['Import Apache Maven key'],
+          Exec["Download Maven ${maven_version}"],
+          Exec["Download Maven ${maven_version} signature"],
+        ],
+        unless  => "/usr/bin/test -f ${maven_bin} && ${maven_bin} --version | /bin/grep --quiet '${maven_version}'",
+      }
+
+      exec { "Verify Maven ${maven_version} checksum":
+        command => "/bin/bash -c \"echo \\$(cat ${temp_checksum})  ${temp_archive} | sha512sum --check\"",
+        cwd     => '/tmp',
+        require => [
+          Exec["Download Maven ${maven_version} checksum"],
+          Exec["Download Maven ${maven_version}"],
+        ],
+        unless  => "/usr/bin/test -f ${maven_bin} && ${maven_bin} --version | /bin/grep --quiet '${maven_version}'",
+      }
+
+      file { $maven_dir:
+        ensure => directory,
+        owner  => 'root',
+      }
+
+      exec { "Extract Maven ${maven_version}":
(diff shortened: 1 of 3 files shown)


dist/profile/manifests/buildagent.pp
```diff
@@ -153,6 +153,110 @@
         unless  => "/usr/bin/test -f ${java_bin} && ${java_bin} -version 2>&1 | /bin/grep --quiet '${jdk_version_to_check}'",
       }
     }
+
+    $maven_version = lookup('profile::jenkinscontroller::jcasc.tools_default_versions.maven', { default_value => undef })
```
dduportal: This instruction creates a strong link between 2 distinct profiles (`buildagent` and `jenkinscontroller`) which is not recommended.

We should specify the Maven version through the hieradata key `profile::buildagent::tools_versions.maven` instead, which should populate the input Hash variable `$tools_versions` here (in the profile's class input values): https://github.com/lemeurherve/jenkins-infra/blob/9bc5fa1a07d3483cd280a7707c60268c58d6136b/dist/profile/manifests/buildagent.pp#L8.

I realize that the current jdk installations in this profile are already doing such a (bad) link and I'm most probably the culprit.

dist/profile/manifests/buildagent.pp
```diff
@@ -153,6 +153,110 @@
...
+        ensure  => file,
+        mode    => '0755',
+        content => "export PATH=${maven_dir}/bin:\$PATH\n",
+      }
+
+      file { '/etc/profile.d/java.sh':
+        ensure  => file,
+        mode    => '0755',
+        content => "export JAVA_HOME=/opt/jdk-21\nexport PATH=\$JAVA_HOME/bin:\$PATH\n",
+      }
```
dduportal:
```suggestion
```

Unneeded and not wanted: these profile script would only be loaded on certain SSH+shell+login type conditions and would not work within a Jenkins agent.

As such, let's keep binary where you put them and we'll setup the `PATH` in our Jenkins pipelines/agents setups/pipeline libraruies)

dduportal (changes requested):
- The Puppet code implementation looks good overall: it does the job and is idempotent.
- The input of the Maven version should be moved to the `tools_version` variable of the `buildagent` plugin (despite the current behavior of JDK installations)
- Need to remove the profile scripts

dist/profile/manifests/buildagent.pp
lemeurherve: Make sense, addressed in https://github.com/jenkins-infra/jenkins-infra/pull/4842/commits/861788c7d0852db8cc1aa52189de5630e2f12504

hieradata/clients/agent-2.trusted.ci.jenkins.io.yaml
```diff
@@ -7,6 +7,8 @@ profile::buildagent::ssh_keys:
     type: ed25519
     key: AAAAC3NzaC1lZDI1NTE5AAAAIItJB1E9CwGG+/bVRToJAH0rdSKNnYmmnhskKwxPVD9u
 profile::buildagent::trusted_agent: true
+profile::buildagent::tools_versions:
+  maven: 3.9.15
```
lemeurherve: TODO follow-up: add as target of https://github.com/jenkins-infra/jenkins-infra/blob/production/updatecli/updatecli.d/jenkinscontroller-tools-maven.yaml

dist/profile/manifests/buildagent.pp
```diff
@@ -153,6 +153,98 @@
         unless  => "/usr/bin/test -f ${java_bin} && ${java_bin} -version 2>&1 | /bin/grep --quiet '${jdk_version_to_check}'",
       }
     }
+
+    $maven_version = pick($tools_versions['maven'], undef)
+
+    if $maven_version {
```
dduportal:
```suggestion
    if $tools_versions['maven'] {
```

No need to use fancy Ruby function :p

lemeurherve: I had to add https://github.com/jenkins-infra/jenkins-infra/pull/4842/commits/ae193104a03269400efadf4fc0eb581b61283893 to define `$maven_version`.
That `pick` wasn't that fancy in the end, tempted to revert ^^

dduportal: `pick` is not available on the puppetmaster (like many other ruby functions). It usually is hard to determine until production. Puppet code as a tendency to look "primal" usually

dduportal: > That pick wasn't that fancy 

If a function provides an "elvis operator" or "coalesce operator" outside SQL, it is fancy stuff :D

dduportal (changes requested):
Thanks for the fixes!

I was tired when I reviewed yesterday: the current code is not idempotent and Maven is reinstalled on each provisioning. I missed it because the hierdata for the vagrant VM does not specify the new key `profile::buildagent::tools_versions:`. I'm checking why it's not idempotent

dduportal:
> Thanks for the fixes!
> 
> I was tired when I reviewed yesterday: the current code is not idempotent and Maven is reinstalled on each provisioning. I missed it because the hierdata for the vagrant VM does not specify the new key `profile::buildagent::tools_versions:`. I'm checking why it's not idempotent

```
$ /usr/bin/test -f /usr/share/apache-maven-3/bin/mvn && /usr/share/apache-maven-3/bin/mvn --version | /bin/grep --quiet '3.9.15'
The JAVA_HOME environment variable is not defined correctly,
this environment variable is needed to run this program.
```

=> caused by my suggestion https://github.com/jenkins-infra/jenkins-infra/pull/4842/commits/998b835f56cf013627c63a1f0c40ef3c8acd58f4.

Absence of the profile is still needed (the Puppet code won't load them outside Vagrant alas). I'm checking what we use to do in the past

dist/profile/manifests/buildagent.pp
```diff
@@ -153,6 +153,98 @@
...
+
+      exec { "Extract Maven ${maven_version}":
+        command => "/usr/bin/tar --extract --gunzip --file=${temp_archive} --directory=${maven_dir} --strip-components=1 && /usr/bin/rm -f ${temp_archive} ${temp_checksum} ${temp_sig}",
+        require => [
+          Exec["Verify Maven ${maven_version} signature"],
+          Exec["Verify Maven ${maven_version} checksum"],
+          File[$maven_dir],
+          Package['tar'],
+        ],
+        unless  => "/usr/bin/test -f ${maven_bin} && ${maven_bin} --version | /bin/grep --quiet '${maven_version}'",
```
dduportal:
```suggestion
        unless  => "/usr/bin/test -f ${maven_bin} && JAVA_HOME=/opt/jdk-25 ${maven_bin} --version | /bin/grep --quiet '${maven_version}'",
```

to solve https://github.com/jenkins-infra/jenkins-infra/pull/4842#issuecomment-4404228966

(suggestion to add to all the other blocks which need it)

lemeurherve: Thanks, addressed in https://github.com/jenkins-infra/jenkins-infra/pull/4842/commits/8fb6234d0cb1d1134cc53acf1e76837cfd07a8b7

Out of curiosity, how did you notice that the code wasn't idempotent? By provisionning the agent twice?

dduportal:
> Out of curiosity, how did you notice that the code wasn't idempotent? By provisionning the agent twice?

(from https://github.com/jenkins-infra/jenkins-infra/pull/4842#discussion_r3207406045).

Yes, exactly: the second run of `vagrant up --provision jenkins::agent` should not show Maven being installed again by the puppet logs.
Also, I usually run a 3rd time after changing the specified version (like downgrade to 3.9.14 for instance) to check if it downgrades the installation as expected.

Note: these elements used to be automatically tested by Puppet unit tests (idempotency/non regression) and integration tests (downgrade/nominal feature) but with the deprecation of many things by Perforce, we've stoped running and maintaining the test frameworks.

When we'll move to Ansible, the Molecule test framework will provide us these facilities with easier tooling (OOB): https://docs.ansible.com/projects/molecule/philosophy/#testing-workflow-fundamentals

dduportal:
Added https://github.com/jenkins-infra/jenkins-infra/pull/4842/commits/5eab30591ec4d234531f3236a1528e76ea474eda to:

- Make sure all VM with the buildagent profile applied would install Maven (we want it as there are no reason to have a permanent agent without Maven)
- Ensure the vagrant `jenkins::agent` manually tested VM also has Maven installed

dduportal (approved):
LGTM:

- First run of `vagrant up --provision jenkins::agent` installs Maven 3.9.15 as expected
- Second run of `vagrant up --provision jenkins::agent` does not do anything else as expected
- Third run with the hierdata tool version set to `3.9.14` downgrades Maven as expected

PR #4895, opened 2026-06-12
chore(updatecli) track the nodejs versions in consumer repos

jayfranco999:
Ref: https://github.com/jenkins-infra/helpdesk/issues/5175

After we dropped ASDF (#5093), the consumer repos lost their way of knowing which NodeJS version the agents actually run. This adds an updatecli manifest that closes that loop: it reads the Node version from the production packer-images and keeps each repo's `.tool-versions` in sync.

It works the same way as the Maven tools manifest, a two-hop source:
1. read the deployed packer-image version from `hieradata/common.yaml`
2. read `nodejs_linux_version` from that image's `tools-versions.yml`

then writes the `nodejs` line in each consumer's `.tool-versions`.

Covers all 8 consumer repos: plugin-site, contributor-spotlight, stats.jenkins.io, docs.jenkins.io, stories, jenkins-io-components, gatsby-plugin-jenkins-layout, incrementals-publisher.

`updatecli diff` does what we'd expect, no change for the repos already onthe current version, and a bump (24.15.0 → 24.16.0) plus a PR for the ones behind:

```diff
--- .tool-versions
+++ .tool-versions
@@ -1 +1 @@
-nodejs 24.15.0
+nodejs 24.16.0
```

Templated the nodejs-consumers manifest per the agreed approach (explicit values-list, loop wrapped outside the YAML doc - same pattern as the azure *.tf.tpl manifests).

Local updatecli diff against all 8 consumer repos: source resolves correctly (packer-image 2.108.1 → nodejs 24.16.0), 5 already up to date, 3 would be bumped 24.15.0 → 24.16.0 with PRs (stories, jenkins-io-components, incrementals-publisher). 0 failures.


dduportal (approved):
- Checks is green and mentions the expected behavior (update of outdated tool versions and no changes for up to date too versions)
- No warnings in output related to the added manifests
- Templating and data structure looks good to me (I like the list, which preserves order, of object with the key `repository`, which allows tracking other GH org. projects such as `jenkinsci` in the future)

dduportal:
Let's merge and verify that `updatecli` opens the expected PR for real (to rule out any missing GH app permission which cannot be detected in the diff)

PR #4926, opened 2026-06-25
feat(ci.jenkins.io): retrieve Maven cache from S3 in Windows EC2 VMs

lemeurherve:
Follow-up of:
- https://github.com/jenkins-infra/jenkins-infra/pull/4925

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5194

PR #4927, opened 2026-06-25
Revert "feat(ci.jenkins.io): retrieve Maven cache from S3 in Windows EC2 VMs"

lemeurherve:
Reverts jenkins-infra/jenkins-infra#4926 as no new Windows agents since its deployment.


dduportal:
The issues were due to Spot instances: https://github.com/jenkins-infra/helpdesk/issues/5199#issuecomment-4799116693

But now we're sure :) You can revert the revert

PR #4930, opened 2026-06-25
feat(buildagent): install jq on Linux build agents with GPG verification

jayfranco999:
Ref: 
- https://github.com/jenkins-infra/helpdesk/issues/5117

Adds `jq` as a managed tool on the Linux Jenkins build agents (`profile::buildagent`), installed from the official `jqlang/jq` GitHub release as a single binary, driven by `profile::buildagent::tools_versions`.

The downloaded binary is verified with GPG against the jq release signing key before installation, following the same pattern as the Maven install in this manifest: the key is imported and the binary's detached signature is checked with gpg --verify.

`update-center2` runs on the Puppet-managed permanent agent (`agent-2.trusted.ci.jenkins.io`), not on a packer-images agent image, so the packer-images jq upgrade does not reach it. Today `jq` is not managed on the trusted.ci agents at all, which is why `update-center2/site/publish.sh` self-downloads its own `jq` from the deprecated `stedolan/jq` repo.

Managing `jq` here is the prerequisite for Acceptance Criteria 2 of jenkins-infra/helpdesk/issues/5117 (single jq source): once the agent reliably ships `jq`, the self-download block in `publish.sh` can be removed so the job uses the agent-provided binary.

Version 1.8.2 matches the packer-images bump (jenkins-infra/packer-images#1064), keeping a single consistent jq version across agents.

Companion to jenkins-infra/packer-images#2867, which adds the same verification
to the packer-built agent images.

## Testing

Validated locally with the repo's Vagrant harness against the `updatecenter` role (which includes `profile::buildagent` via `role::jenkins::agent`):

- `vagrant up --provision updatecenter` → Puppet applies cleanly, `Exec[Download jq 1.8.2]` succeeds with no errors.
- Removed the binary (`rm -f /usr/local/bin/jq`) and re-ran `vagrant provision updatecenter` to confirm reinstall from a clean state.
- Verified inside the VM:
  - `jq --version` → `jq-1.8.2`
  - `which jq` → `/usr/local/bin/jq`
  - The key is imported, the binary and its signature are downloaded, gpg --verify
  passes, and jq is installed to /usr/local/bin/jq (mode 755)
  - Install is gated on the verify step, so a failed verification leaves jq uninstalled

puppet parser validate passes on the manifest.

Follow-up (not in this PR):

- Remove the `stedolan/jq` self-download from `update-center2/site/publish.sh` once this lands. Its three jq filters were verified compatible with jq 1.8.


dist/profile/manifests/buildagent.pp
```diff
@@ -245,6 +245,19 @@
...
       }
     }
+
+    if $tools_versions['jq'] {
+
+      $jq_version = $tools_versions['jq']
+      $jq_bin     = '/usr/local/bin/jq'
+      $jq_url     = "https://github.com/jqlang/jq/releases/download/jq-${jq_version}/jq-linux-${architecture}"
+
+      exec { "Download jq ${jq_version}":
```
dduportal, line 273: We need to verify the download using the GPG key (described in https://jqlang.org/download/)

lemeurherve: Should the same be done in https://github.com/jenkins-infra/packer-images/blob/0ef4e58f15c4ae63a269953733bb8d0b3739428c/provisioning/ubuntu-provision.sh#L233-L239?

dduportal (approved):
Good work. Tested with success 2 additional cases:

- Idempotency: second run of puppet provision on the vagrant local machine does not re-rerun any command
- Downgrade/upgrade: changed the version to 1.8.1: a re-run downgrade successfully. Reverting the change and re-runing again upgrade the binary.

PR #4934, opened 2026-06-30
chore(profile::jenkinscontroller) Cleanup plugins and JCasC management

dduportal:
This PR fixes a few elements in the `jenkinscontroller` profile which will is not expecting to impact our 3 controllers (ci.jio, cert.ci and trusted.ci) BUT could accidentally.

The fixed elements are related to the provisioning of Jenkins plugin installations and JCasC configuration files.
The goal is to fix the following error or edge cases, anticipating for the support of EC2 Fleet plugin and also migration to Ansible:

1. (local dev/tests) Vagrant local VM does not provision and need 2-3 reprovisioning until it reaches a stable state. 
2. (production/local dev/tests) when a JCasC configuration file require a given plugin, the file is often written before the plugin is installed, which may lead to endless loop of restarts (case with ec2 fleet but also have been the case with artifact manager, azure VM and EC2 in the past) requiring a manual operation.

5 changes are expected to solve these 2 problems, they are all well-defined within distinct commits:

- (chore) Avoid specifying a `KUBECONFIG` empty variable when not using credential-less Kubernetes agents. Avoid bad null-iness checks on downstream setups.
- (chore) install plugins with distinct containers than the Jenkins controller. Allows to break endless loops when failing. The additional overhead is not worth mentioning.
  - Note: not removing (`docker run --rm ...`) containers as we have a GC. Useful to keep them for logs when it fails to install a given plugin.
- (chore) improve plugins auto-detection from JCasC configuration (a new plugin is covered, fixed the nested map merge in hieradata and use the resolved jcasc configuration which is more efficient)
- (chore) add a dummy SSH key credential to break endless loop of restarts
- (nit) Removed a now unused resource

⚠️ Do not merge once reviewed: I want to carefully deploy this manually on each controller to avoid weird behavior ⚠️ 

----

Notes: 
- A last commit is added for unit tests (I wrote them at the beginning but forgot to commit it).
- These changes are required ofr EC2 fleet support (see incoming PR definition) to support setting up AWS credential on vagrant

PR #4936, opened 2026-06-30
fix(jenkinscontroller) ensure plugins autodetection works as expected and avoid empty element in `docker run`

dduportal:
Following up #4934 , I realized that instead of using the `map` lambda (ref. https://www.puppet.com/docs/puppet/6/function.html#map) which results in arrays with `nil` elements, we could use the `filter` lambda (ref. https://www.puppet.com/docs/puppet/6/function.html#filter).

It fixes 2 elements:

- (major) autodetection of plugins, which removed some detected plugins when sorting/removing duplicates
- (nit) avoids empty elements in the `env` and `volumes` arrays passed to docker run. Also unit tests have been updated for clarity

### DIFF
--- a/dist/profile/manifests/jenkinscontroller.pp
+++ b/dist/profile/manifests/jenkinscontroller.pp
@@ -384,13 +384,13 @@ class profile::jenkinscontroller (
       'LANG=C.UTF-8', # For context, cfr https://github.com/jenkinsci/docker/pull/1194
       "PATH=${final_container_path}",
       $kubeconfig_envvar_value,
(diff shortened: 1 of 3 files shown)


lemeurherve (approved):

dduportal:
Agent disabled on the controllers, Let's merge

dduportal:
> Agent disabled on the controllers, Let's merge

Applied with success. Triggered a plugin install +restart on cert.ci (which detected the `toolenv` and `workflow-aggregator`)

PR #4945, opened 2026-07-01
fix(ci.jenkins.io) EC2 Fleet Windows agents uses JDK21 by default

dduportal:
Follows up (and requires) https://github.com/jenkins-infra/terraform-aws-sponsorship/pull/550

> The expected JDK (outside the `maven-XX-windows` agents) is 21. Not 25 (ref. https://github.com/jenkins-infra/acceptance-tests/blob/240c14b40248e82e2de34e53a7482e3cdf35d939/checks.ps1#L17).


dduportal:
Tested locally with success!

PR #4994, opened 2026-07-31
chore(updatecli/playwright-consumers) set PR title to a valid semantic message

dduportal:
Goal: avoid problems such as https://github.com/jenkins-infra/jenkins-io-components/pull/234#issuecomment-5132069277


dduportal:
CI checks shows expected new value: 
```
Action		: chore: Bump Playwright version to 1.62.0
```

PR #5000, opened 2026-08-04
chore(updatecli) track NodeJS version in `incremental-publishers` Dockerfile

dduportal:
Ref. https://github.com/jenkins-infra/incrementals-publisher/pull/135#pullrequestreview-4845907907

Requires https://github.com/jenkins-infra/incrementals-publisher/pull/143 first to ensure the build `ARG` instruction is present for tracking.

PR #5001, opened 2026-08-04
chore(updatecli) adapt PR title for `nodejs-consumers` to handle non-chore cases

dduportal:
Fixup of #5000 : if we change the build `ARG` in a `Dockerfile` then it's not really a `chore` anymore


dduportal:
Note: required https://github.com/jenkins-infra/jenkins-infra/commit/b9c9c5c1dcb71b943d4ed3d22bc7a34bd4edad1b as "hotfix"

PR #5024, opened 2026-08-25
feat(census) add `golang` and `postgres`

dduportal:
This PR introduces the following changes for https://github.com/jenkins-infra/helpdesk/issues/4666#issuecomment-5407426309

- Remove the old (and unused) `census:agent` roles and profiles (and associated resources)
- Factorize the golang installation from `profile::letsencrypt` to a new `profile::golang`
- Set up a new `profile::census` with golang (see above) and Postgres (using Docker). It's a simple first implementation
  - Postgres 18 is used. Container image is pulled on each restart to always get the latest 18.x version.
  - The DB is not exposed outside the VM's `localhost:5432`
  - A default DB named `census-data` is set up with a `census` user. It's password is encrypted in production (obviously) and a dummy password is set up for the purpose of Vagrant testing
  - DB data dir is a Docker bind mount set up by puppet with the proper permissions (using numeric UID from the container image). It was too clumsy to move the Docker data dir (`/var/lib/docker`)  to the VM's data disk.

----

Blocked by https://github.com/jenkins-infra/digitalocean/pull/305 (and the associated FS resizing).


dduportal:
Following https://github.com/jenkins-infra/digitalocean/pull/305#issuecomment-5427449150, let's roll!

dduportal:
Initial failure with the following error on the `census.do.jenkins.io` caused by a misnaming of the hierdata client file:

```
Aug 26 15:24:03 census puppet-agent[695941]: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, Class[Profile::Census]: expects a value for parameter 'postgres_password' (file: /etc/puppetlabs/code/environments/production/dist/role/manifests/census.pp, line: 5, column: 3) on node census.do.jenkins.io
```

Fixed by https://github.com/jenkins-infra/jenkins-infra/commit/57440014fbbbfaaefd6964318ef6661c68da921e

PR #5057, opened 2026-09-09
cleanup(updatecli): remove obsolete `windowsVersions` from values

lemeurherve:
This change removes that (now) unused section from updatecli values.

Amends:
- https://github.com/jenkins-infra/jenkins-infra/pull/4979


lemeurherve:
Cleanup concerning only updatecli which is passing, merging.

PR #5072, opened 2026-09-23
feat(jenkins::controller,ci.jenkins.io) enforce default global trusted library to only discover branches and do not allow version override

dduportal:
- Tested locally with Vagrant
- Should only impact ci.jenkins.io


dist/profile/templates/jenkinscontroller/casc/global-libraries.yaml.erb
```diff
@@ -4,13 +4,13 @@ unclassified:
     libraries:
   <%-  @jcasc['global_libraries'].each do |name, setup| -%>
       - defaultVersion: "<%= setup['defaultVersion'] ? setup['defaultVersion'] : 'main' %>"
+        allowVersionOverride: false
```
dduportal, line 7:
```suggestion
        allowVersionOverride: true
```

otherwise contributors of pipeline-library can't contribute (thanks @daniel-beck ).

Fixed by #5073

